PR #162, opened 2014-05-20
Put render time before item name, for ergonomics

estoner:
This makes it easier to scan the list of render times and see what's slow--if the time is to the right of the name, the variable length of the name makes it hard to compare the times.  It'd be even better to color code the pills (slower == redder), but I couldn't figure out a straightforward way to do that.

### DIFF
--- a/app/templates/render_item.handlebars
+++ b/app/templates/render_item.handlebars
@@ -4,8 +4,8 @@
 
       <span {{bind-attr title="name"}}>
         <span class="cell__arrow"></span>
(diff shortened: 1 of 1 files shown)


vvscode:
@teddyzeenny 

Nested tree

In this case duraction column should be first and aligned into one line ( but views as nested tree ) -- do I understand right?

teddyzeenny:
@vvscode correct.

locks:
Superseded by https://github.com/emberjs/ember-inspector/pull/599.

PR #288, opened 2015-01-28
Options dialog

xymbol:
Migrate to new options dialog, as [recommended](https://developer.chrome.com/extensions/optionsV2#migration).


teddyzeenny:
Thanks!

Issue #304, opened 2015-02-06
1.7.1 No longer works with Ember 1.5.1


ryansonshine:
Chrome
Ember-Inspector: 1.7.1
Ember: 1.5.1
Handlebars: 1.0.0
jQuery: 1.10.2

App loads fine, inspector breaks and console returns:

````
Uncaught TypeError: undefined is not a function		VM1027:5940
(anonymous function)							VM1027:5940
triggerOnce 									VM1027:5968
Ember.runLoadHooks 							ember.js:20705
(anonymous function)		 					ember.js:41468
(anonymous function) 							ember.js:41470
(anonymous function)							ember.js:44267
````

It's not client specific; multiple devs having same issue with different configurations (but the same app).

teddyzeenny:
It's not the Ember version. I think in your app the global `define` is defined but `requireModule` is not.  The issue is [here](https://github.com/emberjs/ember-inspector/blob/9532d1e06c6df385512253e4c451eb78a4fb0414/ember_debug/vendor/loader.js#L2). The `&&` should probably be `||`.

@tmquinn I believe you encountered the same issue.

ryansonshine:
Thanks,

Is there already an issue open for this?

tmquinn:
Yes, this appears to be the same as my issue.

ryansonshine:
Can confirm; changing && to || works for my application.

teddyzeenny:
Would you like to submit a PR?

ryansonshine:
https://github.com/ryansonshine/ember-inspector/pull/1

Does that look right? That was my first github pull request

teddyzeenny:
@ryansonshine nice, congrats! :) However you opened the PR to your own fork. You need to open a PR from your fork to this repo. It's usually the default behavior when you open a PR (no extra steps are usually needed).

ryansonshine:
Thanks!

Gotcha, how's that?

PR #305, opened 2015-02-06
Fix for apps where define or requireModule are not both defined

ryansonshine:
Fixes https://github.com/emberjs/ember-inspector/issues/304


teddyzeenny:
:+1: thanks!

PR #316, opened 2015-02-13
Fix errors when EXTEND_PROTOTYPES=false

### LINKED ISSUE #312: Inspector doesn't work with EXTEND_PROTOTYPES=false
I have an app that disables Ember's prototype extensions, which seems to break the inspector. When I try to open the tab, I get `Uncaught TypeError: undefined is not a function` because of the following code (in "ember-debug/libs/source-map"):

```js
      sourceMapCache: function() {
        return {};
      }.property(),
```

omghax:
Fixes #312.

This PR fixes an issue where `ember_debug/libs/source-map.js` assumes Ember's prototype extensions are enabled. If they aren't, then the inspector will crash upon loading.

### DIFF
--- a/ember_debug/libs/source-map.js
+++ b/ember_debug/libs/source-map.js
@@ -5,6 +5,7 @@
  */
 var Ember = window.Ember;
 var EmberObject = Ember.Object;
+var computed = Ember.computed;
 export default EmberObject.extend({
 
   map: function(stack) {
@@ -21,9 +22,9 @@ export default EmberObject.extend({
     });
   },
 
-  sourceMapCache: function() {
+  sourceMapCache: computed(function() {
     return {};
-  }.property(),
+  }),
 
   getSourceMap: function(url) {
     var sourceMaps = this.get('sourceMapCache');


teddyzeenny:
Thanks!

huafu:
Thanks for this! But I still have issues with my app. When this will be deployed to the Chrome store? Or is it and I am missing something?

ZebraFlesh:
Looks like this was incorrectly marked as going into the 1.7.2 release, judging by history.  1.7.2 was released on Feb. 7, but this PR wasn't merged until Feb. 13. With the release of Ember 1.11/ember-cli 0.2.2, this is a Big Deal for add on devs (prototype extensions are now disabled by default).

huafu:
Yup, I usually have to make a PR on each addon I need to use because of the incompatibility with `EXTEND_PROTOTYPES=false`

teddyzeenny:
I plan to release a new patch version this weekend.
> Looks like this was incorrectly marked as going into the 1.7.2 release

@ZebraFlesh where is it marked as 1.7.2?

ZebraFlesh:
> @teddyzeenny where is it marked 1.7.2?

Perhaps I'm reading the history wrong, but it looks to me like this fix was merged to master with a parent commit of fcbbb2e, which is adding the 1.7.2 changelog.

Issue #261, opened 2014-11-13
Ember not detected in chrome:// URLs


nathan-osman:
I am building a Firefox add-on that uses Ember. Unfortunately, the inspector doesn't detect Ember in URLs that begin with `chrome://`. It would be terrific if I could debug my add-on with the Ember inspector.

jakecraige:
I can't find where this was answered before, but chrome won't allow extensions to access the `chrome://` protocol, so it's not possible. 

Using something like [ember-cli-remote-inspector](https://github.com/joostdevries/ember-cli-remote-inspector) you might be able to get it working

dharma-guardian:
Same problem here. Accounts for ```chrome-extension://``` as well.
Thx jakecraige, but I cannot use ```ember serve```, because I need ```chrome.runtime``` as I'm building a chrome extension.

nathan-osman:
This bug is for Firefox, however.

Noah-Huppert:
Just curious where do `chrome://` urls apear in Firefox?

nathan-osman:
Firefox uses the `chrome://` URI scheme to serve content from add-ons.

iamolivinius:
Is there any progress on this issue? I'd love to use ember-inspector to debug my chrome extension :<3:

nathan-osman:
This bug isn't about Chrome extensions, it's about Firefox add-ons. Please file a separate bug for Chrome.

locks:
Closing issue due to staleness. Please reopen if it still applies, and please provide a reproduction!

Issue #115, opened 2014-01-29
Is it possible use ember-extension to dev chrome extension with emberjs


flyingsky:
Hi,

Actually I tried to dev chrome extension with emberjs, but I cannot use the ember-extension to debug my chrome extension. Do you know how to do it?

BTW: do you use ember-extension to help you to dev ember-extension?

rwwagner90:
Does anyone here have ideas on how to make iframes/extensions work?

rwwagner90:
We will support extensions via the standalone app #881.

PR #324, opened 2015-03-20
Adding ability for communication between extensions

pete-the-pete:
#261 and #115 inspired me to create an addon that enables the ember inspector to work with a chrome extension.  I haven't published it yet since it depends on changes to ember-inspector, but wanted to see if it would be useful for anyone.

the addon is here: [pete_the_pete/ember-cli-inspector-bridge](https://github.com/pete-the-pete/ember-cli-inspector-bridge)

### DIFF
--- a/skeleton_chrome/background-script.js
+++ b/skeleton_chrome/background-script.js
@@ -3,7 +3,8 @@
   "use strict";
 
   var activeTabs = {},
-      ports = {};
+      ports = {},
+      externalExtensions = [];
 
   function generateVersionsTooltip(versions) {
     return versions.map(function(lib) {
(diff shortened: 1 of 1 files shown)


jlsalvador:
I was developing a password manager that uses Ember, and this plugin can save me time debugging. What do you (ember-inspect devs) need to make this thing happen?

rwwagner90:
@teddyzeenny do we still want this functionality?

rwwagner90:
@thorsteinsson any insights into this? I'm not exactly sure what this gives us. I think you probably know more about Chrome extensions than I do 😃

rwwagner90:
We will investigate supporting extensions as part of the standalone app.

PR #329, opened 2015-03-22
Upgrade to Ember 1.10

teddyzeenny:


stefanpenner:
nice! Good to see our add-on story is helping remove a-lot of code here.

PR #330, opened 2015-03-23
Upgraded to ember-cli 0.2.0, tests are qunit 2.0 compatible

teddyzeenny:


stefanpenner:
awesome, were you able to use ember-cli-watson for the qunit upgrades?

teddyzeenny:
I found out about it after the update :( 
It was mostly find/replace so the upgrade was really quick, but it would've been nice to try ember-watson.

stefanpenner:
LGTM dude.

PR #337, opened 2015-04-01
Use "ember-new-computed" to polyfill new CP syntax 

salzhrani:
and minor indentation fix

### DIFF
--- a/app/models/promise.js
+++ b/app/models/promise.js
@@ -1,24 +1,25 @@
 import Ember from "ember";
 import escapeRegExp from "ember-inspector/utils/escape-reg-exp";
+import computed from 'ember-new-computed';
 var typeOf = Ember.typeOf;
-var computed = Ember.computed;
-var or = computed.or;
-var equal = computed.equal;
-var not = computed.not;
+var or = Ember.computed.or;
+var equal = Ember.computed.equal;
+var not = Ember.computed.not;
 
 var dateComputed = function() {
-  return Ember.computed(
-    function(key, date) {
-      if (date !== undefined) {
-        if (typeOf(date) === 'date') {
-          return date;
-        } else if (typeof date === 'number' || typeof date === 'string') {
-          return new Date(date);
-        }
+  return computed({
+    get: function() {
+      return null;
+    },
+    set: function(key, date) {
+      if (typeOf(date) === 'date') {
+        return date;
+      } else if (typeof date === 'number' || typeof date === 'string') {
+        return new Date(date);
       }
       return null;
     }
-  ).property();
+  });
 };
 
 export default Ember.Object.extend({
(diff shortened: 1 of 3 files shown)


app/models/promise.js
```diff
@@ -1,24 +1,22 @@
...
-var or = computed.or;
-var equal = computed.equal;
-var not = computed.not;
+var or = Ember.computed.or;
+var equal = Ember.computed.equal;
+var not = Ember.computed.not;
 
 var dateComputed = function() {
-  return Ember.computed(
+  return computed({ set:
```
teddyzeenny: Can the `set` be on the same line as its function?  Same as https://github.com/rwjblue/ember-new-computed#usage

app/models/promise.js
```diff
@@ -1,24 +1,22 @@
...
-        }
+      if (typeOf(date) === 'date') {
+        return date;
+      } else if (typeof date === 'number' || typeof date === 'string') {
+        return new Date(date);
       }
       return null;
-    }
-  ).property();
+    }, get: function() { return null; }
```
teddyzeenny:
```js
},
get: function() {
  return null;
}
```
Same as https://github.com/rwjblue/ember-new-computed#usage

ember_debug/promise-debug.js
```diff
@@ -1,8 +1,8 @@
 import PortMixin from 'ember-debug/mixins/port-mixin';
 import PromiseAssembler from 'ember-debug/libs/promise-assembler';
+import computed from 'ember-new-computed';
```
teddyzeenny: Unfortunately we can't use addons in `ember_debug` since it's not part of the Ember app (this is the part that's injected into the app being debugged). The `ember-new-computed` module does not exist here. The tests are accidentally passing due to module leaks between the Ember app and ember_debug.

teddyzeenny:
@salzhrani thanks for this! I left some comments.

salzhrani:
Updated.

app/models/promise.js
```diff
@@ -1,24 +1,25 @@
...
+    set: function(key, date) {
+      if (typeOf(date) === 'date') {
+        return date;
+      } else if (typeof date === 'number' || typeof date === 'string') {
+        return new Date(date);
       }
       return null;
     }
-  ).property();
+  }).property();
```
rwjblue: No need to call `.property` on the return value of `ember-new-computed`.  It is already a computed property descriptor...

teddyzeenny:
:+1:

rwjblue:
Awesome!

salzhrani:
I am confused about `instrumentWithStack` ... it is not in ember's code base. where is it injected from?

teddyzeenny:
@salzhrani it's not part of the Ember app, but it *is* part of this repo. The inspector is made up of two separate parts, an Ember app (which is the one you see in the devtools panel when using the inspector) and an injected script (everything under `ember_debug` directory in this repo) which is injected from this repo into the app you are debugging.

PR #340, opened 2015-04-05
Tests now use async/await instead of async test helpers

teddyzeenny:
ES7 async/await allows us to get rid of async test helpers and especially `andThen`. It gives us more control over what to wait for, which helps in testing loading routes or timers. The syntax looks much better, and it removes a whole set of bugs and confusions caused by andThens.

Actual test example:

```javascript

test("Searching records", async function(assert) {
  let rows;
  await visit('/data/model-types');
  await click(findByLabel('model-type').find('a').eq(1));
  rows = findByLabel('record-row');
  assert.equal(rows.length, 2);
  await fillIn('[data-label=records-search] input', 'Hello');
  rows = findByLabel('record-row');
  assert.equal(rows.length, 1);
  assert.equal(findByLabel('record-column', rows[0]).first().text().trim(), '2');
});
```

The steps I took to get this working (for anyone who wants to try this):

- Upgraded to Babel 5.0
- Added `optional: ['es7.asyncFunctions']` (thanks @stefanpenner) to the `EmberApp` `babel` options
- Prepended the Babel polyfill to the vendor directory (https://github.com/babel/ember-cli-babel/pull/6 and https://github.com/babel/ember-cli-babel/issues/24 will make things better)
- Replaced jscs with eslint ([which has really great Babel support](https://github.com/babel/babel-eslint)). 
- Set babel-eslint as the eslint parser.
- Disabled jshint in tests using `/* jshint ignore: start */`. Jshint is not needed since eslint covers jshint + jscs.
- Named all async anonymous functions to get around a Babel bug: https://github.com/babel/babel/issues/1135

There's an Issue tracking the steps needed to make this the default for all Ember Cli users: https://github.com/ember-cli/ember-cli/issues/3529

cc @stefanpenner


jonathanKingston:
@teddyzeenny as you know that is now merged, is there anything else I can help with?

Disabling JSHint relates to: https://github.com/switchfly/ember-cli-mocha/issues/44 which I think might require changes to ember-cli to pass the environment config.

@teddyzeenny have you noticed any issues whilst using babel-eslint some of the rules last release caused issues for me, just not had time to look through them all.

teddyzeenny:
@jonathanKingston yeah JSHint should be optional. Maybe it should be extracted from the testing addon into its own addon.

I haven't noticed any issues with babel-eslint yet. Please keep me posted if you have more info.

stefanpenner:
Instead of doing all of stage one you can just add `optional: ['es7.asyncFunctions']`

teddyzeenny:
@stefanpenner :+1:

Issue #222, opened 2014-09-22
Add Error pane to object inspector


teddyzeenny:
Catch errors caused by eagerly calling CPs in the object inspector, and display them in an `Errors` pane above `Own Properties'

Issue #167, opened 2014-05-29
Inspecting an object should not fire .get on all its CPs


machty:
Clicking an object to inspect it in Ember inspector shouldn't run .get() on all of its CPs, since in some cases where you're debugging lazy CP bugs and other issues, Ember Inspector will put you into Heisenbug land if you're trying to figure out why a computed property hasn't been calculated/invalidated.

I propose we offer similar behavior to how Chrome inspector doesn't eagerly computed getter properties, but forces you to click the property value to calculate it:

![clicking properties evaluates getter](http://recordit.co/uHv9lmkJSB.gif)

But for CPs we should display a cached value for CPs that have already been computed, and allow you at any time to click the calculator icon to explicitly call .get() on it.

/cc @hjdivad

Bestra:
@rwwagner90 I think the issue is more that showing a property in the inspector will cause that computed property to recalculate/be evaluated/whatever you want to call it.  I don't think the syntax change had an effect on the root of this issue.  I bet we've all run into the problems that Alex brings up (I've totally had the exact problems he mentions in his final comment) but I think it's worth closing this for the time being.

rwwagner90:
Closing as wontfix for now. After reading through this more thoroughly, I tend to agree that we should not be trying to support all the things here. I think for normal day to day inspection, it is nice to see the property values already, without extra clicks.

PR #342, opened 2015-04-12
Handle and display errors caused by CP calculation in the object inspector

teddyzeenny:
Fixes https://github.com/emberjs/ember-inspector/issues/222

Opening an object in the inspector means we eagerly calculate all of that object's CPs. Sometimes CPs that don't expect to be calculated yet may throw an error (for example CPs that depend on another property that hasn't been set yet). 

Previously, the object inspector just wouldn't open, and the thrown errors were displayed in the app's console.

After this PR these errors are caught by the inspector and displayed above the object's properties.

screen shot 2015-04-12 at 6 16 15 pm

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -20,36 +20,37 @@ export default Ember.Controller.extend({
 
   inspectorExpanded: false,
 
-  pushMixinDetails: function(name, property, objectId, details) {
+  pushMixinDetails(name, property, objectId, details, errors) {
     details = {
-      name: name,
-      property: property,
-      objectId: objectId,
-      mixins: details
+      name,
+      property,
+      objectId,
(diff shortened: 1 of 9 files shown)


stefanpenner:
i know it was done for ergonomic reasons, but the fact that the inspector forces CP's to compute is dubious as merely observing the system should not change it.

teddyzeenny:
Yeah, the issue discussing this is still open https://github.com/emberjs/ember-inspector/issues/167. I don't feel strongly about either approach - both have pros and cons. It may be a good idea to reopen this discussion.

Whatever we decide to do, this PR remains valid as it also handles errors triggered by manual CP calculation. Errors may be more common due to eager computation, but we'll need to handle them either way.

PR #343, opened 2015-04-12
Catch errors triggered by the inspector.

teddyzeenny:
Ember Inspector caused errors are now caught in production and logged as warnings instead of being thrown.

screen shot 2015-04-12 at 8 45 01 pm


stefanpenner:
awesome

PR #352, opened 2015-05-08
Use ember-computed-new in ember-debug to avoid warnings

cibernox:
I've been getting deprecation warnings in my app coming from `ember_debug/models/promise.js`.

Now it uses the `ember-computed-new` addon, that is already being used in other areas.

### DIFF
--- a/ember_debug/addons/ember-new-computed/index.js
+++ b/ember_debug/addons/ember-new-computed/index.js
@@ -1,0 +1,36 @@
+import canUseNewSyntax from './utils/can-use-new-syntax';
+
+var Ember = window.Ember;
+var computed = Ember.computed;
+
+export default function() {
+  var polyfillArguments = [];
+  var config = arguments[arguments.length - 1];
+
+  if (typeof config === 'function' || canUseNewSyntax) {
+    return computed.apply(this, arguments);
+  }
+
+  for (var i = 0, l = arguments.length - 1; i < l; i++) {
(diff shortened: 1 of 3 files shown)


stefanpenner:
:+1:

teddyzeenny:
Unfortunately we can't use addons in ember-debug since it is not part of the ember-cli app. See https://github.com/emberjs/ember-inspector/pull/337#issuecomment-88542966

If we want to use the new computed syntax, we'll have to include the complete polyfill.

cibernox:
@teddyzeenny I'm ok with including the polyfill in the repo if you are. Having deprecation warnings coming not from your code or the code in the addons of your app but from the instrumentation used by the inspector is very misleading and I'd like to fix it.

teddyzeenny:
@cibernox completely agree, and thanks for tackling this! The polyfill would need to be somewhere in [this directory](https://github.com/emberjs/ember-inspector/tree/dd495c054bebb8f021a3a697d174e31ec503ddc1/ember_debug) (maybe create a `polyfills` or `addons` folder?). Then imported into [main](https://github.com/emberjs/ember-inspector/blob/dd495c054bebb8f021a3a697d174e31ec503ddc1/ember_debug/main.js).

cibernox:
@teddyzeenny I've included the `ember-new-computed` in a newly created `/addons` folder. I only had to ES5sify a couple things.

teddyzeenny:
Thank you!

PR #357, opened 2015-05-12
[Glimmer] The view tree now supports glimmer

### LINKED ISSUE #351: [Glimmer] Cannot read property 'forEach' in ember-debug/view-debug
I'm seeing the following error when Ember Inspector is installed.  If the inspector is not installed the exception is not thrown.

```
Ember.Object.extend.appendChildren	@	VM296:2893
Ember.Object.extend.viewTree	@	VM296:2814
(anonymous function)	@	VM296:2775
fn	@	ember.debug.js:377
Queue.invoke	@	ember.debug.js:878
Queue.flush	@	ember.debug.js:943
DeferredActionQueues.flush	@	ember.debug.js:748
Backburner.end	@	ember.debug.js:173
Backburner.run	@	ember.debug.js:228
executeTimers	@	ember.debug.js:610
(anonymous function)	@	ember.debug.js:599

```

### LINKED ISSUE #356: View/Template link is blank
Any idea why some of my components don't show up in the inspector? They are in the list but the View/Template column is blank which makes it harder to inspect...

screen shot 2015-05-11 at 4 04 45 pm

Just trying to figure out if this is caused by something on my end—I've seen this happen a couple of times and it's consistent when it does happen. Thoughts?

teddyzeenny:
Fixes #351 
Fixes https://github.com/emberjs/ember.js/issues/11027 
Fixes #356

There was a shift in concept from a view tree to a render node tree.
Previously, the template tree was made of a hierarchy of view instances.
After the Glimmer rewrite, view instances are optional and secondary, and the main template representation is a render node tree.

This required a re-write of all of view-debug.js. Old code was left untouched, so the current version supports both pre-glimmer and post-glimmer. We detect glimmer by checking for the absence of `_childViews` on the top level view.

All code that touches private api (anything that has to do with render nodes) needs to be moved to Ember core and tested.

### DIFF
--- a/app/components/view-item.js
+++ b/app/components/view-item.js
@@ -51,4 +51,3 @@ export default Component.extend({
   }
 
 });
-
--- a/app/controllers/view-tree.js
+++ b/app/controllers/view-tree.js
@@ -20,20 +20,21 @@ export default Controller.extend({
 
   actions: {
     previewLayer: function(node) {
-      this.get('port').send('view:previewLayer', { objectId: node.value.objectId });
+      // We are passing both objectId and renderNodeId to support both pre-glimmer and post-glimmer
+      this.get('port').send('view:previewLayer', { objectId: node.value.objectId, renderNodeId: node.value.renderNodeId });
     },
 
-    hidePreview: function(node) {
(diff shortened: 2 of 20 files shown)


rwjblue:
:+1:

PR #358, opened 2015-05-12
Speed up the deprecations tab

teddyzeenny:
Previously, if deprecations were triggered on page load (before the app boots), the page could freeze for up to 1 minute.

- Unlike the browser, we don't need to get the source map synchronously. Make everything async.
- Only compute source maps and parse deprecations when requested instead of when triggered.

cc @wagenet @ebryn


wagenet:
Awesome :)

PR #370, opened 2015-05-25
- WIP, Refactorring.

Mawaheb:
This PR, is made to see If I am on the right direction, Doing the right thing, and to hear any advice, suggestions, comments.

This is a WIP, Refactoring the code to use arrow functions and the new functions definition syntax on properties.

@teddyzeenny  Please let me know what do you think.

### DIFF
--- a/app/adapters/basic.js
+++ b/app/adapters/basic.js
@@ -12,7 +12,7 @@
  * ```
  */
 import Ember from "ember";
-const K = Ember.K;
+const { computed, K } = Ember;
 export default Ember.Object.extend({
   name: 'basic',
   /**
@@ -20,20 +20,20 @@ export default Ember.Object.extend({
 
     @param type {Object} the message to the send
   **/
-  sendMessage: function() {},
+  sendMessage() {},
 
   /**
     Register functions to be called
     when a message from EmberDebug is received
   **/
-  onMessageReceived: function(callback) {
+  onMessageReceived(callback) {
     this.get('_messageCallbacks').pushObject(callback);
   },
 
-  _messageCallbacks: function() { return []; }.property(),
+  _messageCallbacks: computed(function() { return []; }),
 
-  _messageReceived: function(message) {
-    this.get('_messageCallbacks').forEach(function(callback) {
+  _messageReceived(message) {
+    this.get('_messageCallbacks').forEach(callback => {
       callback.call(null, message);
     });
   },
@@ -42,6 +42,6 @@ export default Ember.Object.extend({
   willReload: K,
 
   canOpenResource: false,
-  openResource: function(/* file, line */) {}
+  openResource(/* file, line */) {}
 
 });
--- a/app/adapters/bookmarklet.js
+++ b/app/adapters/bookmarklet.js
@@ -1,27 +1,28 @@
 import BasicAdapter from "./basic";
+import Ember from 'ember';
+const { on, computed } = Ember;
 
 export default BasicAdapter.extend({
   name: 'bookmarklet',
 
-  inspectedWindow: function() {
+  inspectedWindow: computed(function() {
     return window.opener || window.parent;
-  }.property(),
+  }),
 
-  inspectedWindowURL: function() {
+  inspectedWindowURL: computed(function() {
     return loadPageVar('inspectedWindowURL');
-  }.property(),
+  }),
 
   sendMessage: function(options) {
     options = options || {};
     this.get('inspectedWindow').postMessage(options, this.get('inspectedWindowURL'));
   },
 
-  _connect: function() {
-    let self = this;
+  _connect: on('init', function() {
 
-    window.addEventListener('message', function(e) {
+    window.addEventListener('message', e => {
       let message = e.data;
-      if (e.origin !== self.get('inspectedWindowURL')) {
+      if (e.origin !== this.get('inspectedWindowURL')) {
         return;
       }
       // close inspector if inspected window is unloading
@@ -29,11 +30,12 @@ export default BasicAdapter.extend({
         window.close();
       }
       if (message.from === 'inspectedWindow') {
-        self._messageReceived(message);
+        this._messageReceived(message);
       }
     });
-  }.on('init')
+  })
 });
+
 
 function loadPageVar (sVar) {
   return decodeURI(window.location.search.replace(new RegExp("^(?:.*[&\\?]" + encodeURI(sVar).replace(/[\.\+\*]/g, "\\$&") + "(?:\\=([^&]*))?)?.*$", "i"), "$1"));
--- a/app/adapters/chrome.js
+++ b/app/adapters/chrome.js
@@ -1,5 +1,7 @@
 /* globals chrome */
 import BasicAdapter from "./basic";
+import Ember from 'ember';
+const { on, computed } = Ember;
 
 let emberDebug = null;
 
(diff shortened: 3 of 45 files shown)


app/adapters/bookmarklet.js
```diff
@@ -16,12 +16,10 @@ export default BasicAdapter.extend({
     this.get('inspectedWindow').postMessage(options, this.get('inspectedWindowURL'));
   },
 
-  _connect: function() {
-    let self = this;
-
+  _connect: () => {
```
locks: wouldn't this be `_connect() {`?

app/adapters/bookmarklet.js
```diff
@@ -16,12 +16,10 @@ export default BasicAdapter.extend({
...
     window.addEventListener('message', function(e) {
```
locks: you forgot to make this a fat arrow fn, the `this` on L22 will be wrong

locks: you forgot to make this a fat arrow fn, the `this` on L22 will be wrong

app/components/drag-handle.js
```diff
@@ -10,30 +10,29 @@ export default Ember.Component.extend({
   isLeft: Ember.computed.equal('side', 'left'),
   minWidth: 60,
 
-  startDragging: function() {
-    let self = this,
-        $container = this.$().parent(),
+  startDragging() {
+    let $container = this.$().parent(),
```
locks: These seem like it could be `const`? I don't see the reference changing during the scope.

app/adapters/bookmarklet.js
Mawaheb: ops, I actually wanted to restore it to it's original state, Any functions with a trailing `.`  (eg `.on()`, `.property()` etc) gives me an error (unexpected `.`) when trying to refactor them!
Any workaround/suggestion for this?

teddyzeenny: Yes, we should stop using `.on` and `.property`.
```js
const { on, computed } = Ember;

_connect: on('init', function() {
}),

myProp: computed('key1', 'key2', function() {
  return this.get('key1') + this.get('key2);
})

```

app/adapters/chrome.js
```diff
@@ -20,7 +20,7 @@ export default BasicAdapter.extend({
     let chromePort = this.get('_chromePort');
     chromePort.postMessage({ appId: chrome.devtools.inspectedWindow.tabId });
 
-    chromePort.onMessage.addListener(function(message) {
+    chromePort.onMessage.addListener( message => {
```
teddyzeenny: no space before `message`

app/adapters/firefox.js
```diff
@@ -38,16 +38,15 @@ export default BasicAdapter.extend({
     }
   },
 
-  _sendIframes: function (urls) {
-    let self = this;
+  _sendIframes(urls) {
     urls.forEach(function(url) {
```
teddyzeenny: `url => `

app/adapters/websocket.js
```diff
@@ -3,34 +3,33 @@ import BasicAdapter from "./basic";
...
-  sendMessage: function(options) {
+  sendMessage(options) {
     options = options || {};
     this.get('socket').emit('emberInspectorMessage', options);
   },
 
-  socket: computed(function() {
+  socket: computed(() => {
     return window.EMBER_INSPECTOR_CONFIG.remoteDebugSocket;
   }).property(),
```
teddyzeenny: no need for `.property` here

app/adapters/basic.js
```diff
@@ -20,20 +20,20 @@ export default Ember.Object.extend({
...
+  onMessageReceived(callback) {
     this.get('_messageCallbacks').pushObject(callback);
   },
 
   _messageCallbacks: function() { return []; }.property(),
 
-  _messageReceived: function(message) {
-    this.get('_messageCallbacks').forEach(function(callback) {
+  _messageReceived(message) {
+    this.get('_messageCallbacks').forEach( callback => {
```
teddyzeenny: No space before `callback`

Mawaheb:
Ok, I think now this PR is ready..
I refactored `on`, `observes` , `property` and `bind`, And as @locks  suggested, I will make another PR dedicated to refactoring the properties to use the new function definition syntax `{ foo() {} }`
@teddyzeenny  Please let me know what do you think, and if you have any notes. Thanks in advance.

app/adapters/bookmarklet.js
```diff
@@ -1,27 +1,28 @@
 import BasicAdapter from "./basic";
+import Ember from 'ember';
+const { on } = Ember;
 
 export default BasicAdapter.extend({
   name: 'bookmarklet',
 
   inspectedWindow: function() {
```
teddyzeenny: `computed(`

app/adapters/bookmarklet.js
```diff
@@ -1,27 +1,28 @@
...
     return window.opener || window.parent;
-  }.property(),
+  },
 
   inspectedWindowURL: function() {
```
teddyzeenny: `computed(`

app/adapters/bookmarklet.js
```diff
@@ -1,27 +1,28 @@
...
+  },
 
   sendMessage: function(options) {
     options = options || {};
     this.get('inspectedWindow').postMessage(options, this.get('inspectedWindowURL'));
   },
 
-  _connect: function() {
-    let self = this;
+  _connect: on('init', () => {
```
teddyzeenny: this needs to be `on('init', function() {` otherwise `this` wouldn't refer to the object instance. Use `function` instead of `=>` for `Ember.on`, `Ember.observer`, `Ember.computed` so you that `this` would be the object itself.

app/adapters/chrome.js
```diff
@@ -13,37 +15,37 @@ export default BasicAdapter.extend({
 
   _chromePort: function() {
```
teddyzeenny: `computed(`

app/adapters/chrome.js
teddyzeenny: extra space ^

app/adapters/websocket.js
teddyzeenny: ^

app/controllers/deprecations.js
```diff
@@ -8,8 +8,8 @@ export default Controller.extend({
   needs: ['application'],
   search: null,
   searchVal: debounceComputed('search', 300),
-  filtered: filter('model', function(item) {
-    return searchMatch(get(item, 'message'), this.get('search'));
+  filtered: filter('model', item => {
```
teddyzeenny: `function`

app/controllers/deprecations.js
```diff
@@ -8,8 +8,8 @@ export default Controller.extend({
...
+    return searchMatch(get(item, 'message'), get('search'));
```
teddyzeenny: `this.get`

app/controllers/iframes.js
```diff
@@ -1,9 +1,9 @@
 import Ember from "ember";
-const { ArrayController, computed, run } = Ember;
+const { ArrayController, computed, run, observer } = Ember;
 const { alias, map } = computed;
 
 export default ArrayController.extend({
-  model: map('port.detectedApplications', function(item) {
+  model: map('port.detectedApplications', item => {
```
teddyzeenny: `function`

app/controllers/promise-item.js
```diff
@@ -11,7 +11,7 @@ const COLOR_MAP = {
 export default Ember.ObjectProxy.extend({
   promiseTreeController: function() {
```
teddyzeenny: `computed(function(){`

app/models/promise.js
```diff
@@ -176,16 +174,16 @@ export default Ember.Object.extend({
...
     }
     return true;
-  }.property('parent.isExpanded', 'parent', 'parent.isVisible'),
+  }),
 
   _allChildren: function() {
     let children = Ember.$.extend([], this.get('children'));
-    children.forEach(function(item) {
+    children.forEach(item => {
       children = Ember.$.merge(children, item._allChildren());
```
teddyzeenny: We should move `$` to `const { $ } = Ember`

app/routes/promises/index.js
```diff
@@ -3,16 +3,15 @@ const { RSVP: { Promise } } = Ember;
 
 export default Ember.Route.extend({
   beforeModel: function() {
-    let route = this;
-    return new Promise(function(resolve) {
-      route.get('port').one('promise:supported', this, function(message) {
+    return new Promise(resolve => {
+      this.get('port').one('promise:supported', this, function(message) {
```
teddyzeenny: We can use `message =>` here right?

teddyzeenny:
@Mawaheb Nice work, this is almost ready! Left some comments. Leaving the new method definition syntax to another PR is a good idea.

Mawaheb:
@teddyzeenny thank you for the feedback, I modified the file accordingly :), Further notes, suggestions are always welcome.

Mawaheb:
Please let me know If I did [this one](https://github.com/emberjs/ember-inspector/pull/370/files#diff-6683cd6732311fd24625eb899b170186L18) right.

And I couldn't  figure out how to re-factor [this one](https://github.com/Mawaheb/ember-inspector/blob/useArrowFunctions/app/controllers/render-tree.js#L27) because it keeps erroring when changed.

teddyzeenny:
Looks good! @Mawaheb that's how I would handle both cases.

PR #371, opened 2015-05-26
Upgrading to lastest ember-cli-eslint

jonathanKingston:
Changes config over to used shared eslint-config-ember with some slight fixes for that; all the rest have been made as exemptions in the new .eslintrc files


teddyzeenny:
:+1:

Issue #392, opened 2015-06-13
Inspector does not boot on Ember 1.13.0


martypenner:
The inspector shows "Ember application not detected" when loading the inspector on a standard Ember 1.13.0 app. Versions 1.12.* are still fine. Here some jsbins demonstrating the problem:

1.12.1:
http://output.jsbin.com/tatono

1.13.0:
http://output.jsbin.com/libuvi

#### Screenshots

1.12.1:
![screen shot 2015-06-13 at 13 06 36](https://cloud.githubusercontent.com/assets/1819798/8145231/abb5ea16-11cd-11e5-8138-8cab30de709a.png)

1.13.0:
![screen-shot-2015-06-13-at-13 07 57](https://cloud.githubusercontent.com/assets/1819798/8145235/b5ecb00a-11cd-11e5-8298-ef2f3bb689d4.jpg)

atsjj:
+1 having same issue

teddyzeenny:
I'm fixing all the things today - this is certainly a priority.

stefanpenner:
@teddyzeenny <3

atsjj:
@teddyzeenny big time thanks!

Issue #315, opened 2015-02-13
Ember.View.addMutationListener is not a function


MartinMalinda:
I am using ember beta with HTMLBars and inspector 1. 7. 2. This error prevents loading of both the inspector and the ember app. Without the inspector, the app runs fine.

This happens in unstable version of chrome 2.0.2298.0 dev (64-bit)
In firefox its the same error but the app continues to run.
In stable chrome, the app runs but I get different errors: 

```
Uncaught TypeError: undefined is not a function
VM201:2771 Ember.Object.extend.viewListener
```
```
Uncaught TypeError: Cannot read property 'updateDurations' of undefined
VM201:2345 __exports__.default.Ember.Object.extend._updateViewTree
```

These errors started to occur after I upgraded to ember-cli 0.1.5 and Ember Beta 1.11. I have also some minor issues with liquid-fire now so its possible that this is not at all related to ember inspector.

This is the error in unstable chrome and firefox
![devtools screenshot]
(http://i.imgur.com/hMKZoQA.png)

teddyzeenny:
Looks like `View.addMutationListener` has been removed by a recent refactor. https://github.com/emberjs/ember.js/pull/10412 just exposed the issue.

backspace:
Just happened to me too. Glad there’s a solution on the way. Is there some quick way I could fix this temporarily? People agreed to develop on the beta of Ember 1.11 but I don’t know that canary would go over well hehe

teddyzeenny:
You can upgrade to Ember canary to remove the error.

backspace:
Okay, thanks!

kriswill:
This also happens in Chrome Canary with:
```sh
Ember                       : 1.10.0
Ember Data                  : 1.0.0-beta.15
jQuery                      : 2.1.3
Ember Simple Auth           : 0.7.3
Ember Simple Auth OAuth 2.0 : 0.7.3
```

teddyzeenny:
@kriswill can you reproduce this in a jsbin?

kriswill:
I'm using ember-cli, but I will attempt to.  I'm also going to try installing as a developer extension in chrome by building the repo locally.  This might be something you've already fixed?

teddyzeenny:
@kriswill the issue was in Ember core, not in the inspector repo. But afaik 1.10 isn't affected by this.

kriswill:
[...]

PR #393, opened 2015-06-13
Fixes for Ember >= 1.13

teddyzeenny:
- Renamed `Ember.Debug` to `Ember.EmberInspectorDebugger`
- Removed any calls to `Ember.View.addMutationListener`
- Use `-view-registry:main` instead of `Ember.View.views`

Fixes https://github.com/emberjs/ember-inspector/issues/392
Fixes https://github.com/emberjs/ember-inspector/issues/315

### DIFF
--- a/ember_debug/view-debug.js
+++ b/ember_debug/view-debug.js
@@ -145,7 +145,6 @@ var ViewDebug = Ember.Object.extend(PortMixin, {
     $(layerDiv).remove();
     $(previewDiv).remove();
     this.get('_lastNodes').clear();
-    Ember.View.removeMutationListener(this.viewTreeChanged);
     this.releaseCurrentObjects();
     this.stopInspecting();
   },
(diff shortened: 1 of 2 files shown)


atsjj:
Fantastic!

I was able to fire up the bookmarklet and everything loaded as expected.

ember_debug/view-debug.js
```diff
@@ -614,10 +611,14 @@ var ViewDebug = Ember.Object.extend(PortMixin, {
...
    */
   _isGlimmer: function() {
-    var id = Ember.keys(Ember.View.views)[0];
-    return id && !Ember.View.views[id].get('_childViews');
+    var id = Ember.keys(this.get('viewRegistry'))[0];
+    return id && !this.get('viewRegistry')[id].get('_childViews');
   },
 
+  viewRegistry: computed('application', function() {
+    return Ember.View.views || this.get('application.__container__').lookup('-view-registry:main');
```
rwjblue: I think the order here should be reversed. In 1.13 Ember.View.views should be present (for legacy support).

Can you confirm (I'm on mobile ATM)? Also, if Ember.View.views (but we should still be using -view-registry if present) is not present on 1.13 I think that is also a bug...

teddyzeenny: I reversed the order. You are right, I found the bug on 1.13 that removed `Ember.View.views`. PRing a fix to Ember core in a bit.

Issue #365, opened 2015-05-19
Component rendered with {{component}} inside {{#each}} cannot be inspected


ramybenaroya:
Components which rendered with the [{{component}} helper](http://emberjs.com/blog/2015/03/27/ember-1-11-0-released.html#toc_component-helper) are not shown in the components inspection view

teddyzeenny:
Thanks for reporting this! Can you add a screenshot of the inspector's Info tab?

ramybenaroya:
Apparently, this happens inside {{#each}} when iterating over an array with dynamic component names.
Here's a fiddle: [code](https://jsfiddle.net/4hq2un16/10/), [output](https://jsfiddle.net/4hq2un16/10/embedded/result/).
 Ember Inspector Info:
![issue365](https://cloud.githubusercontent.com/assets/4154801/7703427/97b1abc2-fe40-11e4-9e4b-746bb6512056.png)

teddyzeenny:
It looks like any component inside an `{{each}}` isn't included in the view tree. Not just the `{{component}}` helper.

Glavin001:
I am also experiencing this issue.

ef4:
This breaks the view tree for anybody using liquid-fire, because all those helpers use `{{#each}}` internally.

teddyzeenny:
Should be fixed by https://github.com/emberjs/ember-inspector/pull/428

PR #428, opened 2015-07-17
Components Inside an Each Added to View Tree

atsjj:
This should address #365, where components inside an {{#each}} are not
inspected.

Feedback on this is greatly appreciated!

### DIFF
--- a/ember_debug/view-debug.js
+++ b/ember_debug/view-debug.js
@@ -640,7 +640,7 @@ export default EmberObject.extend(PortMixin, {
    * @return {Array} children
    */
   _childrenForNode(renderNode) {
-    if (!!renderNode.morphMap) {
+    if (renderNode.morphMap) {
       return keys(renderNode.morphMap).map(key => renderNode.morphMap[key]);
     } else {
       return renderNode.childNodes;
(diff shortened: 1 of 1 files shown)


atsjj:
So, I am getting some odd behavior while inspecting and using my app at the same time. I decided to throw a [WIP] tag in the PR, because this is no where near ready to merge.

ef4:
This sees to work fine, and the code looks good to me. What kind of problems were you seeing?

ember_debug/view-debug.js
```diff
@@ -634,6 +634,20 @@ export default EmberObject.extend(PortMixin, {
   },
 
   /**
+   * Gather the children assigned to the render node.
+   *
+   * @param  {Object} renderNode
+   * @return {Array} children
+   */
+  _childrenForNode(renderNode) {
+    if (!!renderNode.morphMap) {
```
rwjblue: This is causing the linting failure (and subsequently the CI failure). Changing to:

```javascript
if (renderNode.morphMap) { }
```

Should fix.

atsjj:
Sorry for the delay; busy week!

@ef4 the problem I'm having is when inspecting an object, it seems the bindings get all screwed up. I haven't had much time to get back into this, but try inspecting a route or controller and then trigger a transition to another route. That's where I was running into this problem.

ember_debug/view-debug.js
atsjj: Good catch, thanks!

atsjj:
@ef4 I can't reproduce the error any longer and everything is working as I'd expect.

Glavin001:
This is great! Thank you! When will this be published to Chrome extension?

PR #431, opened 2015-07-27
Use an arrow function to preserve this value

kketch:
Using an arrow function here prevent this from being undefined and thus allowing the code to actually clear the interval


rwjblue:
Thanks! Looks good to me.

teddyzeenny:
Thanks!

PR #443, opened 2015-08-07
Show inspector commit id in info tab

okket:
For the internal version that shows up on the info tab use the version string from ember-cli-app-version. The `const version` has to stay because a regex in Gruntfile.js is looking for it.


okket:
Ok, I get it now, the `const version` string gets overwritten by the Gruntfile.js rule. I guess that is superfluous now. I'll remove it for this PR.

teddyzeenny:
Nice, I like that. Though I feel like maybe we should only do that in non-production builds? I'm not sure we should show the commit id in the published version (like the Ember.js does).

rwjblue:
@teddyzeenny - `ember-cli-app-version` is using `git-repo-version` ([here](https://github.com/embersherpa/ember-cli-app-version/blob/v0.3.5/index.js#L9-L13)) which checks if we are on a tag ([here](https://github.com/cibernox/git-repo-version/blob/master/index.js#L11-L14)) and returns it if we are.  

Ember itself uses `git-repo-version` (the underlying library) [here](https://github.com/emberjs/emberjs-build/blob/master/lib/utils/replace-version.js#L9).

@okket - Can you rebase?

teddyzeenny:
@rwjblue nice!

rwjblue:
@okket - Ping?

locks:
Closing due to inactivity. @okket if you're still around and want this, I'd love to merge a fresh PR!

Issue #455, opened 2015-08-21
Filter Models [enhancement]


pete-the-pete:
i'd like to be able to filter models:
- hide empty models
- only show models used by the current route

I'm assuming hiding empty is an easier task.  I'm more than happy to do the work, but any tips on how to approach it are appreciated.

teddyzeenny:
Can you say more about what you mean by empty models? Model types that don't have any records in the store?

pete-the-pete:
Exactly.  Just a simple filter so that only models with records in the store are displayed.  I started looking into it a little, I think I could get something to play with early next week.

vvscode:
Can you describe how can I filter models which used at current route? I want to help, but I can't find idea how to implement that filter. Thx

teddyzeenny:
Can you clarify what you mean by "used at current route"? Has a record that is rendered in the template?  Has a record set as `currentModel` in an active route? Has a record inside an object/array `currentModel`? 
Are you suggesting a filter for the "active" model classes or for the records themselves?

vvscode:
It was a quote from start message ```only show models used by the current route ```. So the question is for @pete-the-pete  I suppose

pete-the-pete:
I was suggesting the 'active' model classes, not necessarily the records themselves.  I was thinking something similar to the View Tree filter options.  So, yes to "Has a record set as currentModel in an active route? Has a record inside an object/array currentModel", but no to "Has a record that is rendered in the template?"

stefanpenner:
seems like this would be a better explored via the route explorer then no?

pete-the-pete:
My request is coming from personal experience; when I wanted to explore the data/models, the Data explorer of ember inspector is pretty noisy (we have too many models in our project).  I was thinking of how I could add some additional filtering within that tab to make exploring the models easier.

Adding something to the Route explorer is an interesting idea, I like that it would tie everything together. , Maybe another column that lists all the models, or a link to the filtered list of models in the Data explorer.

vvscode:
@pete-the-pete  can you explain how can I filter models which are used by current scope of routes?

teddyzeenny:
@pete-the-pete would adding a model type search field help solve the problem of many models?

pete-the-pete:
[...]

PR #458, opened 2015-08-28
Add 'Hide empty models' filter

vvscode:
Improvement for https://github.com/emberjs/ember-inspector/issues/455

### DIFF
--- a/app/controllers/model-types.js
+++ b/app/controllers/model-types.js
@@ -1,9 +1,22 @@
 import Ember from "ember";
-const { Controller, computed: { sort } } = Ember;
+const { Controller, computed: { sort, filter }, get } = Ember;
 
 export default Controller.extend({
   navWidth: 180,
   sortProperties: ['name'],
+  options: {
+    hideEmptyModelTypes: false
+  },
 
-  sorted: sort('model', 'sortProperties')
+  sorted: sort('filtered', 'sortProperties'),
+
+  filtered: filter('model', function(typeItem) {
+    let hideEmptyModels = get(this, 'options.hideEmptyModelTypes');
+
+    if (hideEmptyModels) {
+      return !!get(typeItem, 'count');
+    } else {
+      return true;
+    }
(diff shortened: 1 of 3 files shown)


app/controllers/model-types.js
```diff
@@ -1,9 +1,22 @@
...
+
+  filtered: filter('model', function(typeItem) {
+    let hideEmptyModels = this.get('options.hideEmptyModels');
+
+    if (hideEmptyModels) {
+      return !!Ember.get(typeItem, 'count');
+    } else {
+      return true;
+    }
+  }).property('content', 'options.hideEmptyModels')
```
teddyzeenny: s/content/property

app/controllers/model-types.js
```diff
@@ -1,9 +1,22 @@
...
+  },
 
-  sorted: sort('model', 'sortProperties')
+  sorted: sort('filtered', 'sortProperties'),
+
+  filtered: filter('model', function(typeItem) {
+    let hideEmptyModels = this.get('options.hideEmptyModels');
+
+    if (hideEmptyModels) {
+      return !!Ember.get(typeItem, 'count');
```
teddyzeenny: Let's define `get` at the top of the file (as if we're importing it).
```js
// line 3
const get = Ember.get;
```

teddyzeenny:
Thanks! Looks good.

Since we're calling the classes "Model Types" here can you rename `model` to `model type` where it's mentioned?

app/controllers/model-types.js
vvscode: Thx for your notice

app/controllers/model-types.js
vvscode: Sure. And I used it instead  this.get too

vvscode:
@teddyzeenny  Thanks for correction

teddyzeenny:
:+1: Can you squash the commits?

vvscode:
@teddyzeenny  sure

Issue #463, opened 2015-09-02
Getting error in console when using Ember Inspector: Uncaught TypeError: Cannot set property 'emberExtension' of undefined


Gaurav0:
Stacktrace:

Uncaught TypeError: Cannot set property 'emberExtension' of undefined
(anonymous function) @ content-script.js:29
(anonymous function) @ content-script.js:64

The line 29 is as follows:
document.documentElement.dataset.emberExtension = 1;

While pausing at the error using Chrome debugger,
document.documentElement is 
&lt;svg xmlns="http://www.w3.org/2000/svg" version="1.1" id="Layer_1" x="0px" y="0px" width="139.978px" height="10.32px" viewBox="0 0 139.978 10.32" enable-background="accumulate" xml:space="preserve"&gt;...&lt;/svg&gt;

and document.documentElement.dataset is undefined

teddyzeenny:
Can you help me reproduce the case where `documentElement.dataset` is `undefined`?

Gaurav0:
I am getting this error in Chrome 45 with Ember Inspector 1.9.1 with &lt;svg xmlns&gt;  elements on the page.

devgru:
Seems like SVG elements have no dataset since Chrome 45 (tried on both Mac OS and Windows).

PR #473, opened 2015-09-23
Resolved error: "Cannot set property 'emberExtension' of undefined"

neilthawani:
Generous if/else checks for the existence of document.documentElement.dataset for Chrome browser

### DIFF
--- a/skeleton_chrome/content-script.js
+++ b/skeleton_chrome/content-script.js
@@ -26,9 +26,17 @@
   }
 
   // let ember-debug know that content script has executed
+  if (document) {
+    if (document.documentElement) {
+      if (!document.documentElement.dataset) {
+        document.documentElement.dataset = {};
+      }
+    } else {
+      document.documentElement = {};
+      document.documentElement.dataset = {};
+    }
+  }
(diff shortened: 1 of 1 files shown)


skeleton_chrome/content-script.js
```diff
@@ -26,10 +26,18 @@
...
+    if (document.documentElement) {
+      if (!document.documentElement.dataset) {
+        document.documentElement.dataset = {};
+      }
+    } else {
+      document.documentElement = {};
+      document.documentElement.dataset = {};
+    }
+  }
   document.documentElement.dataset.emberExtension = 1;
```
teddyzeenny: I think if there's not `documentElement` or dataset then it's certainly not an Ember app. We can skip setting `emberExtension = 1` altogether. I think it's better than mutating the document.

neilthawani: Is there a way to reference this issue officially? https://github.com/emberjs/ember-inspector/issues/463

teddyzeenny: @neilthawani what do you mean?

PR #475, opened 2015-09-30
Prevent errors when document.documentElement.dataset is not present.

### LINKED ISSUE #463: Getting error in console when using Ember Inspector: Uncaught TypeError: Cannot set property 'emberExtension' of undefined

rwjblue:
If `document.documentElement` does not have a `dataset` property we will never mark the content script as loaded (and it cannot be an Ember app) this happens in some scenarios with Chrome 45+ and SVG elements.

Fixes #463
Closes #473

### DIFF
--- a/skeleton_chrome/content-script.js
+++ b/skeleton_chrome/content-script.js
@@ -25,8 +25,12 @@
     port.start();
   }
 
-  // let ember-debug know that content script has executed
-  document.documentElement.dataset.emberExtension = 1;
+  // document.documentElement.dataset is not present for SVG elements
+  // this guard prevents that condition from triggering an error
+  if (document.documentElement && document.documentElement.dataset) {
+    // let ember-debug know that content script has executed
+    document.documentElement.dataset.emberExtension = 1;
+  }
 
 
 


rwjblue:
@teddyzeenny - Should be ready for review...

ember_debug/adapters/basic.js
```diff
@@ -117,7 +117,15 @@ export default EmberObject.extend({
...
         this.interval = setInterval(() => {
-          if (document.documentElement.dataset.emberExtension) {
+          if (
+            // if `document.documentElement` does not have a `dataset` property we
+            // will never mark the content script as loaded (and it cannot be an Ember app)
+            // this happens in some scenarios with Chrome 45+ and SVG elements.
+            (document.documentElement && !document.documentElement.dataset) ||
+
+              // when the content script is loaded, this flag is set to `1`
+              document.documentElement.dataset.emberExtensions
```
teddyzeenny: This code wouldn't execute if it's not an Ember app.

skeleton_chrome/content-script.js
```diff
@@ -25,8 +25,13 @@
     port.start();
   }
 
-  // let ember-debug know that content script has executed
-  document.documentElement.dataset.emberExtension = 1;
+  // document.documentElement.dataset is not present for SVG elements
+  // this guard (and a corresponding one in ember-debug/adapters/basic)
+  // prevents that condition from triggering an error
+  if (document.documentElement && document.documentElement.dataset) {
```
teddyzeenny: :+1:

ember_debug/adapters/basic.js
rwjblue: Ok, so you are saying that we do not need to guard here?

teddyzeenny: Confirm

rwjblue: Updated

skeleton_chrome/content-script.js
```diff
@@ -25,8 +25,13 @@
     port.start();
   }
 
-  // let ember-debug know that content script has executed
-  document.documentElement.dataset.emberExtension = 1;
+  // document.documentElement.dataset is not present for SVG elements
+  // this guard (and a corresponding one in ember-debug/adapters/basic)
```
teddyzeenny: can you update the comment?

rwjblue: Updated

teddyzeenny:
Thank you!!

PR #476, opened 2015-09-30
Prevent deprecation for initializer arguments.

rwjblue:
Ember 2.1+ issue a deprecation when using 2 arguments with initializers.

This tweak ensures that our initializer works without deprecation for  all Ember versions.


teddyzeenny:
:+1: Thank you

Issue #378, opened 2015-06-03
Add comment blocks to classes/methods/properties


teddyzeenny:
Part of https://github.com/emberjs/ember-inspector/issues/360

Use YUIDoc format to comment on all classes/methods/properties.

Classes, methods, and properties should have a detailed description above them about what they do and how they are used. If a method/property is clearly self descriptive, no description is needed.

Classes need the `@namespace` and `@class` keywords.
Methods need the `@method method-name`, `@param {Type} name Description`, `@return {Type} Description` keywords.
Properties need  the `@property property-name`, `@type {Type}`, `@default default-value`.

Computed properties are `@property` (not `@method`).

Example:

``` js
// app/adapters/basic.js

const { Object: EmberObject } = Ember;

/**
 * This is the parent adapter to be extended by all other adapters. 
 * It provides an abstraction to keep the code browser/environment agnostic. 
 * Adapters are instantiated and injected into Ember objects automatically. 
 * Properties and methods in this parent class need to be overridden.
 * 
 * @namespace Adapters
 * @class Basic
 */
export default EmberObject.extend({

  /**
   * Inspect a specific DOM element. This usually means using
   * the current environment's tools to inspector the element in the DOM tree.
   *
   * For example, in Chrome, `inspect(elem)` will open the Elements tab in dev tools
   * and highlight the element.
   *
   * @method inspectElement
   * @param {DOM Element} elem
   */
  inspectElement function(/* elem */) {},

  /**
   * Indicates whether the current environment
   * supports opening up a specific file for debugging.
   * An example is opening up a resource in the Chrome
   * Devtools "Sources" panel.
   *
   * @property canOpenResource
   * @type {Boolean}
   * @default false
   */
  canOpenResource: false
});
```

pete-the-pete:
@teddyzeenny I'd like to help with this, it would be a good way to get to know the code.

locks:
@pete-the-pete are you still interested?

pete-the-pete:
Yes, still interested, but obviously got busy and forgot about this.  I
haven't worked on in since that PR.  I can start contributing again next
week, but it someone else is ready to take it, I don't want to be in the
way.

locks:
@pete-the-pete no worries, this is in the "nice to have" camp so you're not blocking work. See you next week!

ynotdraw:
I'd love to lend a hand here!

rwwagner90:
[...]

PR #477, opened 2015-10-05
Adding documentation for extension-level files

pete-the-pete:
Work for #378 

This is a first pass at documentation for the chrome extension files for EmberInspector:
  - content-script.js
  - background-script.js
  - in-page-script.js
  - devtools.js
  - options.js

I wanted to understand the setup and communication flow before getting into the
internals of either EmberDebug or EmberInspector.  The basic pattern is outlined
here, https://developer.chrome.com/extensions/devtools#more, but I wanted to see
exactly how it is all happening.

**There is one logic change**: I removed the setTimeout that called sendIframe in content-script.js.  This seemed unnecessary since EmberInspector calls _injectDebugger() during setup.  Everything
still works locally, but I'm not sure if that is necessary for backward compatibility.

Some Questions:
  - Is camel case naming for the products alright? or should they be dasherized?
     - ex: EmberDebug vs ember-debug?
  - deprecated/old chrome apis? ex. extension.connect vs runtime.connect?
    - There are some differences between the chrome extension docs and the mehtods used throughout the ember-inspector extension.  Do they need to be updated or are they there for older version of chrome?

### DIFF
--- a/skeleton_chrome/content-script.js
+++ b/skeleton_chrome/content-script.js
@@ -1,48 +1,81 @@
-(function() {
+/* global chrome*/
 
+/**
+ * Content script injected into the app page by chrome, works in tandem with the
+ * background-script to coordinate messaging between EmberDebug, EmberInspector and the
+ * ClientApp.  The content-script serves as a proxy between EmberDebug
+ * and the background-script.
+ *
+ * Content scripts are loaded into every page, and have access to the DOM.  This uses that
+ * to inject the in-page-script to determine the ClientApp version onLoad.
+ */
+(function() {
   "use strict";
 
+  /**
+   * Add an event listener for window.messages.
+   * The initial message from EmberDebug is used to setup the event listener
+   * that proxies between the content-script and EmberDebug using a MessagingChannel.
+   *
+   * All events from the window are filtered by checking that data and data.type
+   * properties exist before sending messages on to the background-script.
+   *
+   * See:
+   *     https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API
+   */
   window.addEventListener('message', function(event) {
+    // received initial message from EmberDebug
     if (event.data === 'debugger-client') {
-      var port = event.ports[0];
-      listenToPort(port);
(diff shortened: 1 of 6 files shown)


skeleton_chrome/content-script.js
```diff
@@ -1,68 +1,74 @@
...
+      listenToEmberDebugPort(emberDebugPort);
     } else if (event.data && event.data.type) {
       chrome.extension.sendMessage(event.data);
     }
   });
 
-  function listenToPort(port) {
-    port.addEventListener('message', function(event) {
+  /**
+   * [listenToPort description]
```
teddyzeenny: Missing description.

teddyzeenny:
This is awesome!! Thank you @pete-the-pete! This is really well commented.

> I removed the setTimeout that called sendIframe in content-script.js

There was a case where a delay was needed before sending the iframes - I can't remember which specific case but I prefer that we keep it until more extensive testing is done to make sure it's no longer needed.

> Is camel case naming for the products alright? or should they be dasherized?

What do you mean by naming of the products?

> There are some differences between the chrome extension docs and the mehtods used throughout the ember-inspector extension. Do they need to be updated or are they there for older version of chrome?

We only support the latest Chrome. So if there are newer/better methods we should definitely update them - but probably better as a separate PR.

Adding file name at the top of the files as comments is not needed. I added the file name in https://github.com/emberjs/ember-inspector/issues/378 just to indicate which file I'm using for the example - didn't mean it as part of the file's content.

pete-the-pete:
> I removed the setTimeout that called sendIframe in content-script.js

I'll create a ticket to take this on separately.  The extension is really slow to open for my project, so was looking for an easy perf win.

> Is camel case naming for the products alright? or should they be dasherized?
I mean EmberInspector vs ember-inspector, EmberDebug vs ember-debug and ClientApp vs client-app.

> We only support the latest Chrome. So if there are newer/better methods we should definitely update them - but probably better as a separate PR.

I agree, separate effort. Just wanted to bring it up here.

teddyzeenny:
:smile: camel case seems fine.

PR #482, opened 2015-10-07
Make sure we only reopen the app once

teddyzeenny:


ember_debug/vendor/startup-wrapper.js
```diff
@@ -38,21 +38,24 @@ if (typeof env !== 'undefined') {
...
-          },
-          willDestroy: function() {
-            Ember.EmberInspectorDebugger.destroyContainer();
-            Ember.EmberInspectorDebugger.set('application', null);
-            this._super.apply(this, arguments);
-          }
-        });
+        if (isFirstBoot) {
+          // Watch for app reset/destroy
+          app.reopen({
```
stefanpenner: its worth noting, if this mixin wasn't anonymous duplicate reopens of the same mixin wouldn't result in the duplicate being applied.

teddyzeenny: Ah, nice - good to know :) That would've been a cleaner approach.

PR #491, opened 2015-11-10
Tweak Firefox skeleton again for 44 and later

jryans:
In #470, I updated some internal Firefox DevTools paths for 44 and later.

On the Firefox DevTools team, we made [one more change][1] that also landed in 44, so here's another update to cover that. No other changes like this are anticipated in the near future.

[1]: https://bugzil.la/1203159


teddyzeenny:
Thanks @jryans!

Issue #494, opened 2015-11-17
View Tree tab doesn't work with Ember 2.2.0


SaladFork:
Using Ember 2.2.0 results in the View Tree tab only showing the application row. Checking Components will only show components in the application template.

Everything seems fine with the previous version, Ember 2.1.1. I'm unsure if this is an issue with Ember or Ember Inspector.

PR #496, opened 2015-11-18
Use scope.getSelf when available to get the controller

teddyzeenny:
`controller` local has been removed from templates.

Fixes https://github.com/emberjs/ember-inspector/issues/494


rwjblue:
:+1:

PR #504, opened 2015-12-30
Prevent error when `renderNode` is null/undefined.

rwjblue:
During some portions of view teardown, when the inspector is open `renderNode` can be `undefined` (because it was already removed).  This guards to prevent an error in that case...

Thanks to @andrewtimberlake for reporting...

### DIFF
--- a/ember_debug/view-debug.js
+++ b/ember_debug/view-debug.js
@@ -694,7 +694,9 @@ export default EmberObject.extend(PortMixin, {
    * @return {Boolean}
    */
   _nodeIsView(renderNode) {
-    if (renderNode.getState) {
+    if (!renderNode) {
+      return false;
+    } else if (renderNode.getState) {
       return !!renderNode.getState().manager;
     } else {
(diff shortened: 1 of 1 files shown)


teddyzeenny:
I think we should filter out all undefined render nodes from the array returned from [this method](https://github.com/emberjs/ember-inspector/blob/c161f3ae42b60089f293716be17e5899fa8b7a9b/ember_debug/view-debug.js#L640) or return [here](https://github.com/emberjs/ember-inspector/blob/c161f3ae42b60089f293716be17e5899fa8b7a9b/ember_debug/view-debug.js#L620) if `renderNode` is undefined, because even if the node is determined as *not a view* we still try to [get its children](https://github.com/emberjs/ember-inspector/blob/c161f3ae42b60089f293716be17e5899fa8b7a9b/ember_debug/view-debug.js#L629) which will likely result in an error.

rwjblue:
Sounds good, I'll update shortly.

PR #505, opened 2016-01-07
Fix websocket adapter ES6 conversion

alexspeller:
The conversion to ES6 of this file a long time back broke it apparently as the scope of `this` was lost in this function


teddyzeenny:
Thanks @alexspeller!

PR #533, opened 2016-04-13
Skip in page script injection in Chrome PDF Viewer

### LINKED ISSUE #532: PDF links are displayed as blank pages in chrome
With the ember-inspector installed for chrome when loading most PDF links, it will display as a blank white page. I have no idea what can cause this, but only when the ember-inspector extension is this an issue.

I haven't tried this on another computer but I can reproduce it on my machine as follows.

- have ember inspector active
- visit http://ga.berkeley.edu/wp-content/uploads/2015/02/pdf-sample.pdf (sample pdf)
- a blank page is displayed  😢 
- reload page
- the PDF is displayed

Relevant information:

ember inspector version: 1.9.5
chrome version: 49.0.2623.112 (64-bit)
mac version: OS X El Capitan Version 10.11.3 (15D21)

If you need any more information let me know. Thanks!

felixbuenemann:
There's an issue in all current Chrome versions, where the built-in PDF-Viewer only displays a blank page, if the Ember Inspector extension is active.

This fix works around the issue, by skipping the injection if the document's content type is reported as PDF which should never be true for normal HTML pages.

Fixes #532.

For a test case see issue #532.

### DIFF
--- a/skeleton_chrome/content-script.js
+++ b/skeleton_chrome/content-script.js
@@ -79,7 +79,7 @@
   var script = document.createElement('script');
   script.type = "text/javascript";
   script.src = chrome.extension.getURL("panes/in-page-script.js");
-  if (document.body) {
+  if (document.body && document.contentType !== "application/pdf") {
     document.body.appendChild(script);
     script.onload = function() {
       document.body.removeChild(script);


teddyzeenny:
Thanks!

PR #535, opened 2016-04-19
Bundle multiple versions of the inspector to support all Ember versions

teddyzeenny:
This commit will allow us to drop support for old Ember versions on master so we can move faster while at the same time not breaking the inspector on old Ember apps.

### How this works

  - Whenever we want to stop supporting an Ember version we take a snapshot of the current inspector and compress and upload it to S3. Current master will only support the latest Ember versions (example: Ember 2.0+).

  - Before publishing the extension, we fetch the uploaded snapshots and bundle them with the current inspector and publish all of them together.

  - During runtime, we detect the Ember version and start the inspector version that matches this Ember version. This is all automatic and does not require the user to do anything.

TL;DR the Ember Inspector published extension will contain several bundled Ember Inspector versions and boot up the suitable one depending on the app.

### The reasoning behind this approach

- One of the biggest challenges while working on the inspector has been supporting all Ember versions (since Ember 1.0-RC). This is becoming harder with each released Ember version, especially after Ember 2.0 was released. We need to drop support for the old Ember versions.

- As with everything in Ember, although we want to move fast, we don't want to leave developers behind. People who have not kept up with the latest Ember version should still be able to inspect their apps. We need to keep support for the old Ember versions.

- There's a very good chance that when a developer opens the Ember Inspector they are usually debugging some kind of problem, so even adding one manual step could break their chain of thought or use up time/energy they'd rather spend on their app. The process needs to be automatic. 

Thanks @omarhamdan for helping with this.


locks:
@teddyzeenny how does this affect the remote inspector addon?

teddyzeenny:
@locks the remote inspector will only be able to use one inspector version at a time (which means one Ember version range, ex: Ember v1.0.0 -> v2.0.0.  But that is okay IMO because the remote inspector is *per app* so you can pick the version you need for your app, whereas the extension is *per browser* and should therefore support all apps.

rwjblue:
👏

PR #543, opened 2016-04-29
Remove usage of list-view so we can upgrade Ember.

teddyzeenny:
List-view has blocked us at Ember 1.12. We need to remove it if we want
to move on. We can look into alternatives as soon as we upgrade Ember.

### DIFF
--- a/app/controllers/container-type.js
+++ b/app/controllers/container-type.js
@@ -12,7 +12,7 @@ export default ArrayController.extend({
 
   search: null,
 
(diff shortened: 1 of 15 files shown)


stefanpenner:
Ember-collection might do the trick

teddyzeenny:
@stefanpenner I'm thinking of using https://github.com/runspired/smoke-and-mirrors. cc @runspired

stefanpenner:
sure w/e works best for the use-case.

PR #551, opened 2016-06-06
Workaround for Ember v1.10.0 to v1.12.0 update Container/Registry reset.

raido:
This fixes multiple Ember app inspection issue. 

Release v1.10.0 introduced regression, when selecting another app to inspect from `<select>` it crashed with error, that is cannot "re-register adapter:main". This commit provides workaround until we sort out multiple app instrumentation without calling app.reset() in between.

Ember Twiddle demonstrating the issue [https://ember-twiddle.com/918ee5f4d473abe786f6e5a102b7b619](https://ember-twiddle.com/918ee5f4d473abe786f6e5a102b7b619)

[Ember v1.12 reset](https://github.com/emberjs/ember.js/blob/v1.12.0/packages/ember-application/lib/system/application.js#L658)
[Ember v1.10 reset](https://github.com/emberjs/ember.js/blob/v1.10.0/packages/ember-application/lib/system/application.js#L658)

Commit that upgraded Ember v1.10.0 to v1.12.0 for Inspector - 524afb5

@nathanhammond - CC as requested.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -61,8 +61,33 @@ App.initializer({
   }
 });
 
-
 loadInitializers(App, config.modulePrefix);
 
+/**
+ * This is here a "temporary" workaround for Container reset that Ember v1.10 provided "out of the box"
+ *
+ * Problem: Ember v1.11.0 introduced ApplicationInstance concept and with this Application.reset()
+ * method changed. In v1.10.0 reset() always created new Container() and thus everything in Container was reset.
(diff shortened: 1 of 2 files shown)


raido:
buildRegistry is actually public method in recent Ember.js versions - http://emberjs.com/api/classes/Ember.Application.html#method_buildRegistry

locks:
I believe that due to how Inspector is now structured that we won't be fixing older versions? This was probably more timely at the time you submitted, sorry :\
If I'm mistaken about anything please discuss, and reopen if necessary!

PR #552, opened 2016-06-06
Better iframe embedded Ember App detection

raido:
This has been a struggle for me for months now and finally gave it a shot to implement it better instead of simple "on boot only" iframes detection.

Anyway, this commit introduces MutationObserver on document.body and onLoad event listeners on frames.

What use cases does it solve?
- Dynamically loaded iframes. For example: App A boots, then later on it injects iframe to DOM and loads another Ember App B which seamlessly integrates to App A. - my main issue detecting App B until now :)
- Dynamic SRC changes - because onLoad fires each time iframe src changes
- Works if iframe is included directly in index.html

@nathanhammond - CC as requested

### DIFF
--- a/skeletons/chrome/content-script.js
+++ b/skeletons/chrome/content-script.js
@@ -89,35 +89,34 @@
   /**
    * Send the iframes to EmberInspector so that it can run
    * EmberDebug in the context of the iframe.
+   *
+   * @param {DOMEvent} event
    */
 
-  function sendFrameToInspector(url) {
-    chrome.extension.sendMessage({type: 'iframes', urls: [url]});
+  function sendFrameToInspector(event) {
+    chrome.extension.sendMessage({type: 'iframes', urls: [event.target.src]});
+  }
+
+  function setupOnLoadListenersForIFrames() {
+    // get all current frames and attach onLoad event
+    var iframes = document.getElementsByTagName('iframe');
+    for (var i = 0, l = iframes.length; i < l; i ++) {
+      var iframe = iframes[i];
+      if (!('__inspector__on__load_listener' in iframe)) {
+        iframe.__inspector__on__load_listener = true;
+        iframe.addEventListener('load', sendFrameToInspector);
+      }
+    }
   }
 
   // create an observer instance for body tag, so if any iframes get pushed
   // to DOM we attach onLoad event listener on it and when it has finally loaded
   // we send it's URL to Inspector, otherwise we might send URL but Ember app in the
   // frame is still loading and thus will not be detected
-  var forEach = Array.prototype.forEach;
-  var bodyObserver = new MutationObserver(function(mutations) {
-    mutations.forEach(function(mutation) {
-      forEach.call(mutation.addedNodes, function(node){
-        if (node.tagName === 'IFRAME') {
-          node.addEventListener('load', function() {
-            sendFrameToInspector(node.src);
-          });
-        }
-      });
-    });
-  });
-  bodyObserver.observe(document.body, { childList: true });
+  new MutationObserver(function() {
+    setupOnLoadListenersForIFrames();
+  }).observe(document.body, { childList: true });
 
-  // get all current frames and attach onLoad event
-  var iframes = document.getElementsByTagName('iframe');
-  for (var i = 0, l = iframes.length; i < l; i ++) {
-    iframes[i].addEventListener('load', function() {
-      sendFrameToInspector(this.src);
-    });
-  }
+  // get frames already in DOM on boot
+  setupOnLoadListenersForIFrames();
 }());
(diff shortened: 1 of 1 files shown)


skeletons/chrome/content-script.js
```diff
@@ -87,22 +87,37 @@
...
+            sendFrameToInspector(node.src);
+          });
+        }
+      });
+    });
+  });
+  bodyObserver.observe(document.body, { childList: true });
 
+  // get all current frames and attach onLoad event
+  var iframes = document.getElementsByTagName('iframe');
```
nathanhammond: Probably want to make sure that you don't double-add to the events added by the mutation observer.

skeletons/chrome/content-script.js
```diff
@@ -87,22 +87,37 @@
...
+  }
+
+  // create an observer instance for body tag, so if any iframes get pushed
+  // to DOM we attach onLoad event listener on it and when it has finally loaded
+  // we send it's URL to Inspector, otherwise we might send URL but Ember app in the
+  // frame is still loading and thus will not be detected
+  var forEach = Array.prototype.forEach;
+  var bodyObserver = new MutationObserver(function(mutations) {
+    mutations.forEach(function(mutation) {
+      forEach.call(mutation.addedNodes, function(node){
```
nathanhammond: It _may_ be faster to do a `MutationObserver` + `getElementsByTagName` than iteration over all `addedNodes`.

raido: It will avoid iterating over and over again all the mutations that might happen on component heavy app. Because we actually do not care about the mutations in general, we just want to know if anything has changed in DOM and that should enough + `getElementsByTagName`.

skeletons/chrome/content-script.js
```diff
@@ -89,35 +89,34 @@
...
+  function sendFrameToInspector(event) {
+    chrome.extension.sendMessage({type: 'iframes', urls: [event.target.src]});
+  }
+
+  function setupOnLoadListenersForIFrames() {
+    // get all current frames and attach onLoad event
+    var iframes = document.getElementsByTagName('iframe');
+    for (var i = 0, l = iframes.length; i < l; i ++) {
+      var iframe = iframes[i];
+      if (!('__inspector__on__load_listener' in iframe)) {
```
raido: I am not sure about this... Maybe removeEventListener would still be better or some other way?

skeletons/chrome/content-script.js
```diff
@@ -89,35 +89,34 @@
...
-          node.addEventListener('load', function() {
-            sendFrameToInspector(node.src);
-          });
-        }
-      });
-    });
-  });
-  bodyObserver.observe(document.body, { childList: true });
+  new MutationObserver(function() {
+    setupOnLoadListenersForIFrames();
```
raido: This probably should be debounced to do less work.

nathanhammond: I don't think you can debounce it without risking missing the `load` event?

raido: Yes, there would be a high chance if missing the load event.

teddyzeenny:
Hey, thanks for working on this! Sorry for the delay in reviewing.

I have a couple of comments:

- I am concerned about the mutation observer on body and how that affects performance. Content scripts are executed on *all* apps on *every* page load/iframe (not just Ember apps or when the Ember Inspector is activated). This means we may literally slow down browser performance in general for users who install this extension.
- This is a Chrome-only fix, we should consider fixing it for FF too (if it's also broken).
- There is [already code](https://github.com/emberjs/ember-inspector/blob/942d379cdef38afc3712f9af270798752ebf3ff5/app/adapters/chrome.js#L55) that attempts to detect late iframes. Obviously the code is failing somehow which caused you to work on this fix, but my question is whether you attempted to look into fixing the existing solution since `chrome.devtools.inspectedWindow.onResourceAdded` is most likely a much less costly solution than the body mutation observer (if we can get it to work).

raido:
I'll look at onRecourceAdded first, i missed it.

Looking at current onRecourceAdded code, it is mainly missing onLoad handler registration. It sends url right away but in my case frame is first initated with about:blank, later url changes but onRecourceAdded does not detect that. 

Instead if onRecourceAdded attached onLoad event listener on iframe it would work for late frame inits and even url changes. 

I'll refactor my solution into exiating onResourceAdded part.

raido:
This is weird, Inspector from Chrome Store no longer fails on my iframe setup nor locally built version. Let's keep this on hold for some time. What's a bigger issue for me is the https://github.com/emberjs/ember-inspector/pull/551 if we could land that soon, that would be great.

raido:
Closing this as there are not many participants here and this is no longer issue probably. Well actually I just don't have use case for iframes anymore.

PR #555, opened 2016-06-17
Fix deprecated window.postMessage overload

evansims:
Resolves feature deprecation warning in Chrome Canary V53; legacy window.postMessage overload https://www.chromestatus.com/features/5719033043222528

### DIFF
--- a/ember_debug/adapters/chrome.js
+++ b/ember_debug/adapters/chrome.js
@@ -8,7 +8,7 @@ export default BasicAdapter.extend({
   connect() {
     const channel = this.get('_channel');
     return this._super(...arguments).then(() => {
-      window.postMessage('debugger-client', [channel.port2], '*');
+      window.postMessage('debugger-client', '*', [channel.port2]);
       this._listen();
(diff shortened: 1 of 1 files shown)


teddyzeenny:
Thanks!

nerdybeast:
@evansims this doesn't seemed to be fixed.  I'm currently running chrome Beta v54, ember-inspector v2.0.2, and ember v2.8.1 and I'm still seeing this error in the chrome console: `Failed to execute 'postMessage' on 'Window': The 3rd argument is neither an array, nor does it have indexed properties.`  This is preventing ember-inspector from working.  Thoughts?

evansims:
@nerdybeast Hmm, that sounds like an entirely different issue from what this commit fixed. I've not encountered that one myself, sorry. Probably worth opening an issue if you haven't already.

nerdybeast:
@evansims thanks for the response.  I lied in my last comment, I was only on Ember v2.8.0 but after "actually" upgrading to 2.8.1 the inspector is working again.

PR #553, opened 2016-06-06
WIP Upgrade Ember v1.12 > v1.13 (CLI included)

raido:
It builds, some tests even pass and it actually boots and detects your app but that's about it.

PR #563, opened 2016-07-20
Upgrade Ember 1.12 -> Ember and Ember CLI 2.6

teddyzeenny:
Thank you @raido for working on a big part of this in https://github.com/emberjs/ember-inspector/pull/553.


Serabe:
I am working on support for Glimmer 2. What do you need to land this? Anything I can help you with?

Thank you for your work!

app/components/date-property-field.js
```diff
@@ -2,31 +2,27 @@ import Ember from "ember";
...
   escape: 27
 };
 
 export default DatePicker.extend({
-  attributeBindings: ['label:data-label'],
+  /**
+   * Workaround bug of `onPikadayClose` being called
+   * on a destroyed component.
+   */
+  onPikadayClose() {
```
rwjblue: We should open an issue upstream for this

app/templates/deprecations.hbs
```diff
@@ -1,69 +1,18 @@
...
-            </div>
-          </div>
-        {{/if}}
-        {{/if}}
+{{#x-list class="js-deprecations"}}
+  {{#x-list-content contentHeight=application.contentHeight headerHeight=0}}
+    {{#each filtered as |content|}}
+      {{deprecation-item
+        model=content
+        openResource="openResource"
```
rwjblue: Lets pass these as closure actions, then we don't need the actions that just call `sendAction` in `deprecation-item` component.

app/components/deprecation-item.js
```diff
@@ -0,0 +1,26 @@
...
+  tagName: '',
+
+  hasMap: notEmpty('model.hasSourceMap'),
+
+  actions: {
+    toggleExpand() {
+      this.toggleProperty('isExpanded');
+    },
+    openResource(...args) {
+      this.sendAction('openResource', ...args);
```
rwjblue: I mentioned this below, but if we pass closure actions in here we don't need to manually bubble these actions like this.

rwjblue:
TBH, I think we should land this and just iterate on master.  

@teddyzeenny - What do you think?

teddyzeenny:
@rwjblue that seems fine. @locks is testing it out locally to make sure nothing's broken.

package.json
```diff
@@ -22,8 +22,7 @@
   },
   "repository": "https://github.com/emberjs/ember-inspector",
   "engines": {
-    "node": ">= 0.10.0",
-    "npm": ">= 2.0.0"
+    "node": ">= 0.10.0"
   },
   "emberVersionsSupported": [
```
locks: can you explain what this key and the next one do? Is it supposed to be an empty string in the list?

teddyzeenny: `emberVersionsSupported`: The range of Ember versions supported by current master. It's an array, first item is "from", second item is "to" which is an Empty string because it's 2.7.0+.
`previousEmberVersionsSupported`: all the previous builds and what Ember versions they support. These versions are only included in the final build before deployment.

app/templates/deprecations.hbs
teddyzeenny: Good point. Fixed.

locks:
I did some cursory testing and as far as I can tell everything is functional :)

teddyzeenny:
@Serabe thanks, this PR is good to go. Let me know if you need anything.
Anybody would like to merge 😄

raido:
Awesome!

Serabe:
Thank you, guys!

Issue #567, opened 2016-08-04
Persist "Current Route only" through refreshes


locks:

teddyzeenny:
Should "just work" by making `hideRoutes`a query param [here](https://github.com/emberjs/ember-inspector/blob/67404b7aea2bccd8733ad49bbaccfa7b7b5c59b2/app/controllers/route-tree.js).

teddyzeenny:
Fixed by https://github.com/emberjs/ember-inspector/pull/568

PR #568, opened 2016-08-04
Current route only persists now after changing tabs if checked

chrism:
Hi,

This adds the queryParam as requested here
https://github.com/emberjs/ember-inspector/issues/567

I haven't written a test, but have tried it locally and it appears to work.

### DIFF
--- a/app/controllers/route-tree.js
+++ b/app/controllers/route-tree.js
@@ -6,7 +6,10 @@ const { Controller, computed, inject: { controller } } = Ember;
 export default Controller.extend({
   application: controller(),
 
+  queryParams: ['hideRoutes'],
+
   currentRoute: null,
+  hideRoutes: computed.alias('options.hideRoutes'),
 
   options: {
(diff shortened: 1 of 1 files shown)


teddyzeenny:
Hey! Thanks for working on this 😄 

I think this is missing `hideRoutes: alias('options.hideRoutes')`? Are you sure it's working?

The commit message says "Current route only persists now after changing tabs if checked", however we need it to persist across refreshes.

- Go to "Routes"
- Check "Current Route Only"
- Refresh the page
- "Current Route Only" should be checked

chrism:
Hey!

Sorry you are totally right — I didn't check between refreshes, but have added the extra alias and now it does.

Hopefully now this is good to merge - unless you'd prefer me to try to squash the two commits?

Thanks

teddyzeenny:
Looks good - yes squash the commits if you don't mind.

chrism:
👍

PR #571, opened 2016-08-09
Broccoli: Replace obsolete Broccoli plugins with "broccoli-funnel"

Turbo87:
This seems needed to update the ESLint broccoli plugin. I've verified that the output in `dist` when running `ember build` stays the same but feel free to verify once more.

/cc @teddyzeenny


stefanpenner:
👍

ember-cli-build.js
```diff
@@ -131,10 +121,9 @@ module.exports = function(defaults) {
     return "'" + item + "'";
   }).join(',') + ']';
 
-  var startupWrapper = pickFiles('ember_debug', {
-    srcDir: '/vendor',
+  var startupWrapper = new Funnel('ember_debug', {
```
rwjblue: Can this use the local variable `emberDebug` instead of the static string?

ember-cli-build.js
```diff
@@ -145,16 +134,14 @@ module.exports = function(defaults) {
     }]
   });
 
-  var sourceMap = pickFiles('ember_debug', {
-    srcDir: '/vendor',
+  var sourceMap = new Funnel('ember_debug', {
```
rwjblue: Same as above?

ember-cli-build.js
```diff
@@ -145,16 +134,14 @@ module.exports = function(defaults) {
...
+    srcDir: 'vendor',
     files: ['source-map.js'],
-    destDir: '/'
   });
 
-  var loader = pickFiles('ember_debug', {
-    srcDir: '/vendor',
+  var loader = new Funnel('ember_debug', {
```
rwjblue: Same?

ember-cli-build.js
stefanpenner: can we also do `__dirname + '/ember_debug'`

teddyzeenny: We can't use the variable since it excludes `vendor/startup-wrapper.js` which we're picking here.

ember-cli-build.js
teddyzeenny: Same as https://github.com/emberjs/ember-inspector/pull/571/files#r74094059

ember-cli-build.js
teddyzeenny: ^

Turbo87:
@rwjblue I believe we can do something like that too, but I wanted to focus that PR on the funnel first before changing other stuff. Don't worry, I got more in the pipeline :wink:

rwjblue:
Seems good, sorry for the noise.

Turbo87:
@teddyzeenny any comments from your side? otherwise this seems ready for merging :)

teddyzeenny:
@Turbo87 thank you! This looks good at first glance but I prefer to try it out and double check the dist. I can't do that and merge until tomorrow morning, is this blocking you from something you're doing right now?

Turbo87:
@teddyzeenny nope, all good, thanks!

PR #572, opened 2016-08-10
Update ESLint and enable more rules

Turbo87:
/cc @teddyzeenny


ember_debug/main.js
```diff
@@ -12,13 +12,15 @@ import DeprecationDebug from "ember-debug/deprecation-debug";
...
+const { Object: EmberObject, run, Application, Namespace } = Ember;
 
 const EmberDebug = EmberObject.extend({
 
   application: null,
   started: false,
 
+  // Using object shorthand syntax here is somehow having strange side effects.
+  // eslint-disable-next-line object-shorthand
   Port: Port,
```
teddyzeenny: What were the side effects, `Port` is `undefined`? My first guess is because [the es6 module is being transpiled before Babel](https://github.com/emberjs/ember-inspector/blob/930fc378f232f85ab51c906a74556aab4ca17f46/ember-cli-build.js#L109-L115) so `Port` import is being transpiled to `Port['default']` before Babel gets the file.

Turbo87: huh, that might indeed explain things... I've got another PR waiting that ditches the es6module plugin and uses Babel for that like Ember CLI does too. Once that is merged we can remove the comment and use the shorthand syntax.

ember_debug/view-debug.js
```diff
@@ -2,7 +2,7 @@
 import PortMixin from "ember-debug/mixins/port-mixin";
 
 const Ember = window.Ember;
-const { guidFor, $, computed, run, Object: EmberObject, typeOf, Component, Controller, ViewUtils, A, LinkComponent, String } = Ember;
+const { guidFor, $, computed, run, Object: EmberObject, typeOf, Component, Controller, ViewUtils, A, LinkComponent, String, View } = Ember;
```
teddyzeenny: This is for Ember >= 2.7, there's no `Ember.View`.  It's the [old code](https://github.com/emberjs/ember-inspector/blob/930fc378f232f85ab51c906a74556aab4ca17f46/ember_debug/view-debug.js#L555) that needs to be fixed (by removing `|| View.views`).

Turbo87: just to clarify: should I remove `|| View.views` instead? the issue was that `View` is undefined since it's not imported from anywhere or defined as a global.

teddyzeenny: Right - removing `|| View.views` (which no longer works anyway) should remove the eslint error correct?

tests/ember_debug/object-inspector-test.js
```diff
@@ -27,6 +26,7 @@ function setupApp() {
 let ignoreErrors = true;
 
 module("Ember Debug - Object Inspector", {
+  // eslint-disable-next-line object-shorthand
   beforeEach: async function() {
```
teddyzeenny: This should be `async beforeEach() {` if I'm not mistaken.

teddyzeenny:
❤️ !!!!

tests/ember_debug/object-inspector-test.js
Turbo87: tried that, but my IDE screamed at me when I did it... might not be supported that well yet

ember_debug/view-debug.js
Turbo87: yes, I'll remove it

teddyzeenny:
Looks good to me. Is this ready to merge?

Turbo87:
@teddyzeenny should be ready to merge from my side. Thanks for the quick review!

teddyzeenny:
Can you squash the ESlint rules into one commit?

Turbo87:
@teddyzeenny done

PR #573, opened 2016-08-10
Update build dependencies

Turbo87:
/cc @teddyzeenny


ember-cli-build.js
```diff
@@ -102,13 +100,10 @@ module.exports = function(defaults) {
     emberDebug = mergeTrees([emberDebug, linted]);
   }
 
-  emberDebug = new ES6Modules(emberDebug, {
-    esperantoOptions: {
-      absolutePaths: true,
-      strict: true
-    }
+  emberDebug = esTranspiler(emberDebug, {
```
teddyzeenny: We need to convert relative module imports to absolute paths since the current loader doesn't support it. Either that or we fix the loader. Or we have to update imports to use the full path. This setup will have some broken imports (any import starting with `./` or `../`).

Turbo87: this should be taken care of now by 5f42100

teddyzeenny:
👍

PR #579, opened 2016-08-26
Glimmer 2 support

teddyzeenny:
- [x] Use correct API to walk the component tree. Update [these two methods](https://github.com/teddyzeenny/ember-extension/blob/a98c0e4299e522e624014d4e1dcb9fcd3d5606d9/ember_debug/libs/glimmer-tree.js#L238-L260) when Ember core is ready.
- [x] Make sure instrumentation is still working after [this PR](https://github.com/emberjs/ember.js/pull/14130) is merged. Inspector may need some minor tweaks.
- [x] Everything else 🎆 

cc @chancancode @rwjblue

### DIFF
--- a/app/adapters/chrome.js
+++ b/app/adapters/chrome.js
@@ -64,6 +64,16 @@ export default BasicAdapter.extend({
   },
 
   /**
+   * Open the devtools "Elements" tab and select a specific DOM element.
+   *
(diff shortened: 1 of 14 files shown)


rwjblue:
https://github.com/emberjs/ember.js/pull/14130 landed upstream

PR #599, opened 2016-09-21
Splits render time into separate column

locks:
Ember Inspector - Render Performance

- [x] Separate render time into column
- [x] ~~Make render time column sortable~~
- [x] Add odd/even row style

### DIFF
--- a/app/templates/components/render-item.hbs
+++ b/app/templates/components/render-item.hbs
@@ -3,8 +3,10 @@
     <span class="list__cell-arrow"></span>
     <span title="{{model.name}}">
       <span class="js-render-profile-name">{{model.name}}</span>
-      <span class="pill pill_not-clickable js-render-profile-duration">{{ms-to-time model.duration}}</span>
     </span>
+  {{/list.cell}}
+  {{#list.cell class="list__cell_value_numeric"}}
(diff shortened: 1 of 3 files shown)


app/templates/components/render-item.hbs
```diff
@@ -3,10 +3,12 @@
     <span class="list__cell-arrow"></span>
     <span title="{{model.name}}">
       <span class="js-render-profile-name">{{model.name}}</span>
-      <span class="pill pill_not-clickable js-render-profile-duration">{{ms-to-time model.duration}}</span>
     </span>
   {{/list.cell}}
   {{#list.cell class="list__cell_value_numeric js-render-profile-timestamp"}}
```
teddyzeenny: `js-render-profile-timestamp` does not apply to this column anymore. Maybe `js-render-profile-duration`.

teddyzeenny:
👍

teddyzeenny:
> Make render time column sortable

This is a bit tricky since it's a hierarchical list so I don't think you can completely sort since you'd need to preserve the hierarchy.

locks:
> This is a bit tricky since it's a hierarchical list so I don't think you can completely sort since you'd need to preserve the hierarchy.
That's a good point, removing it.

Issue #249, opened 2014-10-24
View rendering times are inaccurate, add warning


mixonic:
The times reported in the view rendering pane, though helpful for comparison between views and discovering issues, are known to be skewed and inaccurate due to the instrumentation required.

I propose adding a sticky warning to the top of the rendering pane:

![](https://api.monosnap.com/image/download?id=CgENICZvjDGmpOvJ5qYtwde6RNYvSQ)

I imagine something like the yellow block here:

![](https://api.monosnap.com/image/download?id=0FRHdKD3BU9LHIW2Nzg9oBEArogXx9)

We've seen people reporting these numbers back to the Ember repo and discuss as if they were accurate and additional profiling was not discovered, and this is to be discouraged.

/cc @stefanpenner

arjunkathuria:
@locks got some time on my hands this week, can give a shot.

locks:
@arjunkathuria yes, please! :) Reach us in on Slack for any help.

arjunkathuria:
@teddyzeenny @locks what issue ? :trollface:

locks:
Merged! :D

PR #601, opened 2016-09-24
feat(warning-msg): adds a warning message on Render pane.

arjunkathuria:
PR for issue #249 
Adds a warning message atop the render pane for the view timings being inaccurate due to overhead and environment.

- [ ] Chrome
- [ ] Firefox
- [ ] Bookmarklet (?)

cc:- @locks @teddyzeenny 

~~issue:- works fine on chrome, while on firefox, when pressing X, the warning doesn't instantly goes away like its designed to and does in chrome but needs a refresh. is it a Firefox xpi thing?~~

What it looks like :- 

screen shot 2016-09-27 at 7 53 23 pm

screen shot 2016-09-27 at 7 55 01 pm


app/components/warning-message.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  actions: {
+    setIsClosed() {
+      let ls = this.storage;
```
locks: You should always `.get`/`.set` properties:

```javascript
let ls = this.get('storage');
```

app/components/warning-message.js
```diff
@@ -0,0 +1,12 @@
...
+      ls.setItem('isClosed', true);
+      let warningMsg = event.target.parentNode;
```
locks: Where does `event` come from? It's not defined in the function.

app/components/warning-message.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  actions: {
+    setIsClosed() {
```
locks: You are manually manipulating the DOM in this action handler, which isn't a good practice in Ember. This is because Ember expects to control the DOM, and when you manually mutate it, the state of the DOM gets out of sync with what Ember thinks it is.
You already have the `isClosed` property in the controller, and are conditionally wrapping the warning-message on it, you should fully utilize it.

app/templates/render-tree.hbs
```diff
@@ -1,3 +1,7 @@
+{{#unless isClosed}}
```
locks: As mentioned above, you already have the conditional, so instead of removing the DOM element you should mutate the property.

app/templates/render-tree.hbs
```diff
@@ -1,3 +1,7 @@
...
+  {{warning-message storage=storage}}
```
locks: Pass `isClosed` instead of `storage`.

app/templates/components/warning-message.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="warning">
+    <img src="assets/images/warning.png" alt="warning icon">
+    <p>Render times may be inaccurate due to instrumentation and environment</p>
+    <button {{ action 'setIsClosed' }}> X </button>
+</div>
+{{yield}}
```
locks: Remove `yield` for now, since it isn't necessary.

locks:
Hi @arjunkathuria, I have reviewed your PR and I have two suggestions on possible ways forward:

## 1. `warning-message` as plain UI component

Make `isClosed` a [get/set computed property](https://guides.emberjs.com/v2.7.0/object-model/computed-properties/#toc_setting-computed-properties) and [pass the action](https://guides.emberjs.com/v2.7.0/components/triggering-changes-with-actions/#toc_passing-the-action-to-the-component) that sets `isClosed` from the controller to the component.

## 2. `warning-message` as generic warning component

Inject the `storage` service into the component instead. You'd have to pass in the key under which you'd like the state of a specific warning message to be saved.
Use the block form of `{{warning-message}}` to pass in the actual message, like: 

`render-tree.hbs`
```handlebars
{{#warning-message key="render-performance"}}Render times may be inaccurate due to instrumentation and environment{{/warning-message}}
```

`warning-message.hbs`
```handlebars
…
<p>{{yield}}</p>
…
```

Since the `storage` service would be inside the component, the conditional would go in the component template, or you could bind it to the `hidden` DOM attribute.

app/templates/render-tree.hbs
arjunkathuria: will that be a 2-way binding ?

locks: Sorry, I slightly misspoke. You'd pass an action that sets `isClosed` instead. The component doesn't actually need to know anything, as the controller holds the value and the conditional

arjunkathuria:
@locks Updated!

This update seems to have fixed the firefox issue as well.

app/templates/render-tree.hbs
```diff
@@ -1,5 +1,5 @@
 {{#unless isClosed}}
-  {{warning-message storage=storage}}
+  {{warning-message closed=(action "setIsClosed")}}
```
locks: If you bind `setIsClosed=(action "setIsClosed")` you shouldn't need the action inside the component.

locks: Even more nitpicky, can you name it something a bit more clear, like `closeWarning`?

app/components/warning-message.js
```diff
@@ -0,0 +1,4 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
```
teddyzeenny: This file isn't needed.

app/controllers/render-tree.js
```diff
@@ -2,16 +2,44 @@ import Ember from "ember";
...
 const { and, equal, filter } = computed;
 const get = Ember.get;
+let LOCAL_STORAGE_SUPPORTED;
+try {
+  LOCAL_STORAGE_SUPPORTED = !!window.localStorage;
+} catch (e) {
+  // Security setting in chrome that disables storage for third party
+  // throws an error when `localStorage` is accessed.
+  LOCAL_STORAGE_SUPPORTED = false;
+}
```
teddyzeenny: Can you make this make this a class property on the local storage service since we're using it in two places now?

```js
// app/services/local-storage.js
LocalStorage.reopenClass({
  SUPPORTED: LOCAL_STORAGE_SUPPORTED
});
```

```js
// app/controllers/render-tree.js, app/components/x-list.js
import LocalStorageService from 'ember-inspector/services/local-storage';
storage: service(`storage/${LocalStorageService.SUPPORTED ? 'local' : 'memory'}`)
```

app/controllers/render-tree.js
```diff
@@ -2,16 +2,44 @@ import Ember from "ember";
...
   application: controller(),
   initialEmpty: false,
   modelEmpty: equal('model.length', 0),
   showEmpty: and('initialEmpty', 'modelEmpty'),
-
+  storage: service(`storage/${LOCAL_STORAGE_SUPPORTED ? 'local' : 'memory'}`),
+  // isClosed: computed(function() {
+  //   let ls = this.get('storage');
+  //   return ls.getItem('isClosed') || false;
+  // }),
```
teddyzeenny: Needs cleanup ^

app/controllers/render-tree.js
```diff
@@ -2,16 +2,44 @@ import Ember from "ember";
...
+  isClosed: computed({
+    get() {
+      let ls = this.get('storage');
+      return ls.getItem('isClosed') || false;
```
teddyzeenny: Just a tip: you can coerce `null`/`undefined` values to `false` using `!!`.
`return !!this.get('storage').getItem('isClosed');`

app/templates/components/warning-message.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="warning">
+    <img src="assets/images/warning.png" alt="warning icon">
+    <p>Render times may be inaccurate due to instrumentation and environment</p>
+    <button {{ action closeWarning }}> close </button>
```
teddyzeenny: Can you remove spaces at the beginning and end? `{{action closeWarning}}`. Also the action name is fine but I would've named it just `close` given that we are inside the warning message :)

app/styles/warning.scss
```diff
@@ -0,0 +1,29 @@
...
+  float: right;
+  font-weight: 200;
+  color: #6E68C6;
+  text-decoration: underline;
+  background: #FFFEC2;
+  border: hidden;
+  cursor: pointer;
+  padding: 1px;
+  margin: 0 15px 0 5px;
+}
```
teddyzeenny: We use [BEM naming](http://getbem.com/naming/) for the classes. In case you're not familiar with it it's just that instead of using the HTML tags to style we would use `.warning__image .warning__text .warning__close`. It's a bit easier to maintain as we can more easily change the HTML tags or add tags without breaking the design.

app/templates/components/warning-message.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="warning">
+    <img src="assets/images/warning.png" alt="warning icon">
```
teddyzeenny: If you can easily find an svg instead of a png then it would be great since we wouldn't have to worry about resolution. Otherwise png is fine :)

app/styles/warning.scss
```diff
@@ -0,0 +1,29 @@
...
+  background: #FFFEC2;
+  font-size: 11px;
+  text-align: left;
+  padding: 5px 0 6px 0;
+}
+
+.warning img {
+  display: inline;
+  width: 15px;
+  float: left;
```
teddyzeenny: Did you try using flexbox instead of float?

app/controllers/render-tree.js
arjunkathuria: ✔️

app/controllers/render-tree.js
arjunkathuria: ✔️

app/templates/components/warning-message.hbs
arjunkathuria: ✔️

app/controllers/render-tree.js
btecu: At this time `ember-inspector` doesn't work in Safari, but if it did, this check is not sufficient.
When you use Safari in Private mode, it appears that you have `localStorage` but you don't.
Instead, something like this would work:
```
    try {
      localStorage.setItem('test', 0);
      localStorage.removeItem('test');
      return true;
    } catch(_) {
      return false;
    }
```

app/styles/warning.scss
arjunkathuria: ✔️

app/components/warning-message.js
arjunkathuria: ✔️

app/styles/warning.scss
arjunkathuria: ✔️

app/controllers/render-tree.js
arjunkathuria: @locks not sure i follow.

teddyzeenny: @btecu the inspector works with Safari if you use the bookmarklet. Thanks for the suggestion.

arjunkathuria: @teddyzeenny there isn't a file named app/services/local-storage.js currently, do i make one or use the app/services/storage/local.js ?

teddyzeenny: Ah I meant `app/services/storage/local.js`.

arjunkathuria: ✔️

arjunkathuria:
@teddyzeenny @locks Updated!

app/controllers/render-tree.js
```diff
@@ -12,6 +13,44 @@ export default Controller.extend({
...
   showEmpty: and('initialEmpty', 'modelEmpty'),
 
+  /**
+   * Service used for storage. Storage is
+   * needed for remembering if the user closed the warning
+   * as it might get mildly annoying for devs to see and close
+   * the trival warning everytime.
+   * The default storage service is local storage however we
+   * fall back to memory storage if local storage is disabled (For
+   * example as a security setting in Chrome).
```
teddyzeenny: 👍

app/controllers/render-tree.js
```diff
@@ -12,6 +13,44 @@ export default Controller.extend({
...
+   *
+   * @property storage
+   * @return {Service}
```
teddyzeenny: Properties have `@type` and methods have `@return`. This should have `@type {Service}`.

app/controllers/render-tree.js
```diff
@@ -12,6 +13,44 @@ export default Controller.extend({
...
+    get() {
+      return !!this.get('storage').getItem('isClosed');
+    },
+    set(key, value) {
+      this.get('storage').setItem('isClosed', value);
+      return value;
+    }
+  }),
+  actions: {
+    // This action when triggered, closes the warning message for redering times being inaccurate
```
teddyzeenny: s/redering/rendering

app/services/storage/local.js
```diff
@@ -62,3 +63,23 @@ export default Service.extend({
...
   }
 });
+
+try {
+  LOCAL_STORAGE_SUPPORTED = !!window.localStorage;
+} catch (e) {
+  // Security setting in chrome that disables storage for third party
+  // throws an error when `localStorage` is accessed.
+  LOCAL_STORAGE_SUPPORTED = false;
+}
```
teddyzeenny: Can you add @btecu's suggestion for Safari support?
```js
try {
  localStorage.setItem('test', 0);
  localStorage.removeItem('test');
  LOCAL_STORAGE_SUPPORTED = true;
} catch (e) {
  // Security setting in chrome that disables storage for third party
  // throws an error when `localStorage` is accessed. Safari in Private mode
  // also throws an error.
  LOCAL_STORAGE_SUPPORTED = false;
}
```

app/templates/components/warning-message.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="warning">
+    <img class="warning__image" src="assets/images/warning.png" alt="warning icon">
+    <p class="warning__text">Render times may be inaccurate due to instrumentation and environment</p>
+    <button class="warning__close" {{action close}}> close </button>
```
teddyzeenny: Can you make the text say `hide` here instead?

app/controllers/render-tree.js
```diff
@@ -12,6 +13,44 @@ export default Controller.extend({
...
+  storage: service(`storage/${LocalStorageService.SUPPORTED ? 'local' : 'memory'}`),
+
+  /**
+   * Checks if the user previously closed the warning by referencing localStorage
+   * it is a computed get/set property.
+   *
+   * @property isClosed
+   * @type {Boolean}
+   */
+  isClosed: computed({
```
teddyzeenny: Can you rename this to `isWarningClosed`?  `isClosed` alone is ambiguous in this context, as if the entire render-tree is closed.

app/controllers/render-tree.js
```diff
@@ -12,6 +13,44 @@ export default Controller.extend({
...
+    get() {
+      return !!this.get('storage').getItem('isClosed');
```
teddyzeenny: Please rename this to `is-render-tree-warning-closed`. Storage is shared across the entire app so we need to avoid naming conflicts.

app/templates/render-tree.hbs
```diff
@@ -1,3 +1,7 @@
+{{#unless isClosed}}
+  {{warning-message close=(action "closeWarning")}}
+{{/unless}}
```
teddyzeenny: Unfortunately we need to tell `x-list` the header's height (so that it can calculate its content height). The warning here adds to the header's height so the default height (31) no longer applies when the warning is opened. 

You can see the problem if you fill the render tree and scroll to the bottom while the warning is opened. You won't be able to see the last row.

The way to solve this is to do the following:

```js
// app/controllers/render-tree.js
headerHeight: computed('isWarningClosed', function() {
  return this.get('isWarningClosed') ? 31 : 56;
})
```

```js
// app/templates/render-tree.hbs
{{#x-list .... headerHeight=headerHeight}}`
```

app/controllers/render-tree.js
```diff
@@ -12,6 +13,44 @@ export default Controller.extend({
...
+    },
+    set(key, value) {
+      this.get('storage').setItem('isClosed', value);
+      return value;
+    }
+  }),
+  actions: {
+    // This action when triggered, closes the warning message for redering times being inaccurate
+    // and sets `isClosed` value to true, thus preventing the warning from being shown further.
```
teddyzeenny: Can you update this to:
```js
/**
 * This action when triggered, closes the warning message for redering times being inaccurate
 * and sets `isClosed` value to true, thus preventing the warning from being shown further.
 *
 * @method closeWarning
 */
```

app/controllers/render-tree.js
arjunkathuria: ✔️

app/controllers/render-tree.js
arjunkathuria: ✔️

app/templates/components/warning-message.hbs
arjunkathuria: ✔️

app/services/storage/local.js
arjunkathuria: ✔️

app/controllers/render-tree.js
arjunkathuria: ✔️

app/controllers/render-tree.js
arjunkathuria: ✔️

app/controllers/render-tree.js
arjunkathuria: ✔️

app/templates/render-tree.hbs
arjunkathuria: ✔️

arjunkathuria:
@teddyzeenny Updated!

teddyzeenny:
Thanks for working hard on this! It's good to go, can you squash the commits?

teddyzeenny:
💃

Issue #611, opened 2016-10-25
Inspector seems to alter child route loading substate


GavinJoyce:
Opening the inspector on the routes tab seems to alter the behaviour of the child route loading substate. See https://github.com/GavinJoyce/glammer/pull/8 for an app which demonstrates the issue.

With the inspector closed:

![before](https://cloud.githubusercontent.com/assets/2526/19685156/fb307812-9ab2-11e6-8332-00437b65494e.gif)

With the inspector open on the routes tab (notice that the child `loading.hbs` route never displays) :

![after](https://cloud.githubusercontent.com/assets/2526/19685165/08f8d304-9ab3-11e6-94a9-11a5d408de46.gif)

locks:
Hm, this one's going to be a doozy.

teddyzeenny:
Looks like `router.getHandler('parent.child_loading')` actually generates a `parent.child_loading` route if it doesn't exist. So as the inspector loops over the route names and calls `getHandler` on them it is unintentionally creating the (empty) route `parent.child_loading` which takes precedence over `parent/loading`.

We should maybe consider using container lookup instead of the route's `getHandler` method to get the route handlers because this should definitely not have side effects.

teddyzeenny:
Ok the solution is to add [this check](https://github.com/emberjs/ember.js/blob/77c51f6e80322c9f0bd189be6c3c7bb5054a1bd1/packages/ember-routing/lib/system/router.js#L1184-L1188)  on all `_loading` and `_error` sub routes.

teddyzeenny:
@GavinJoyce thanks for reporting this, it must have been a tough one to debug! I just published a new version with the fix, can you confirm it's working? (Make sure you're on Ember Inspector 2.0.4).

GavinJoyce:
@teddyzeenny thanks for fixing, I just tried `2.0.4` and I can confirm that it has fixed the issue

PR #612, opened 2016-10-28
Only list subroutes if they are defined

teddyzeenny:
Fixes https://github.com/emberjs/ember-inspector/issues/611


locks:
Thank yyoouu :D

Issue #214, opened 2014-09-17
> $E  should print the toString of the object before letting chrome decide what to print


stefanpenner:
this will result in more helpful output.

maheshsenni:
@stefanpenner @teddyzeenny something like this?

![screen shot 2015-04-04 at 6 41 33 pm](https://cloud.githubusercontent.com/assets/1472437/6992770/fa5a9a54-dafa-11e4-859c-1b2769506bd4.png)

arjunkathuria:
@locks ping! 
got some time on my hands next 1-2 days, can you please define the issue for me, like you do. 😼

locks:
@arjunkathuria you need to find the place where the logic of the `$E` binding to the console is, and make it also log the `toString` of the object, so we have a better idea what it is.

teddyzeenny:
Fixed by https://github.com/emberjs/ember-inspector/pull/614

PR #614, opened 2016-11-05
'$E' now also prints the name of object sent to console.

arjunkathuria:
### PR for issue #214 

Clicking the '$E' now also prints the name of the object sent to console along with the object for easier identification and clarity, resulting in more helpful output.

It does so by making use of the `inspect` function

Looks like this:- 
- includes a controller, route and a model 
screen shot 2016-11-05 at 11 12 04 pm

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -207,7 +207,11 @@ export default EmberObject.extend(PortMixin, {
     if (value instanceof Error) {
       value = value.stack;
     }
-    this.get("adapter").log('Ember Inspector ($E): ', value);
+    let args = [value];
+    if (value instanceof EmberObject) {
+      args.unshift(inspect(value));
+    }
+    this.get("adapter").log('Ember Inspector ($E): ', ...args);
   },
 
   digIntoObject(objectId, property) {


ember_debug/object-inspector.js
```diff
@@ -207,7 +207,7 @@ export default EmberObject.extend(PortMixin, {
     if (value instanceof Error) {
       value = value.stack;
     }
-    this.get("adapter").log('Ember Inspector ($E): ', value);
+    this.get("adapter").log('Ember Inspector ($E): ', inspect(value), value);
```
teddyzeenny: I'm thinking that although it is uncommon as there would be no reason to send it to the console, `$E` could also be a string, number, boolean, null, or undefined (for example when you `$E` a value in the object inspector that is not an object). In this case there would be no need to print the `toString()` (and possibly misleading since the number `1` would then appear as `"1"`). A more common case is when it's an `Error` and the value is the stack trace string of that error (see line 208).

So I suggest we first check if it's truthy (to avoid nulls because `typeof null === 'object'`) and if it's an object before the inspected value.

teddyzeenny:
Thanks!

PR #617, opened 2016-11-20
Remove Ember.K.

locks:


app/components/drag-handle.js
```diff
@@ -1,5 +1,8 @@
 import Ember from "ember";
-const { computed, String: { htmlSafe }, K } = Ember;
+
+const { computed, String: { htmlSafe } } = Ember;
+const K = function() { return this; }
```
rwjblue: IMHO, the replacement here is simply to use empty functions where needed. We don't need another "special" Emberism, let's just use functions.

stefanpenner:
👍

app/mixins/row-events.js
```diff
@@ -36,7 +36,7 @@ export default Mixin.create({
    * @type {Function}
    * @default Ember.K
```
sduquej: 🔪

app/mixins/row-events.js
```diff
@@ -45,7 +45,7 @@ export default Mixin.create({
...
```
sduquej: 🔪

app/mixins/row-events.js
```diff
@@ -27,7 +27,7 @@ export default Mixin.create({
...
```
sduquej: 🔪

app/components/drag-handle.js
```diff
@@ -45,7 +47,7 @@ export default Ember.Component.extend({
    * @type {Function}
    * @default Ember.K
```
sduquej: 🔪

locks:
Done!

PR #620, opened 2016-12-07
package.json: Add missing ember-cli-sass dependency

mriska:
Thanks @teddyzeenny, this worked perfectly.


locks:
Thanks for the contribution, but it seeems like `ember-cli-sass` got added meanwhile.

PR #638, opened 2017-03-03
Avoid error in some environments

pablobm:
There seems to be a problem, in very specific circumstances, with the code that detects support for session storage:

screen shot 2017-03-03 at 18 36 55

To reproduce follow these steps:

  1. On Google Chrome
  2. Disable 3rd-party cookies
  3. Open the browser inspector
  4. Open on the Ember inspector
  5. Visit a page showing an Ember app, on a frame
     loaded from a different domain

It's important that the Ember inspector is already open when you land on the page (hence step 4 before 5). Reloading the iframe page with the Ember inspector open also reproduces the problem.

This PR appears to fix the problem for me.

### DIFF
--- a/ember_debug/services/session.js
+++ b/ember_debug/services/session.js
@@ -7,8 +7,29 @@ const Session = EmberObject.extend({
   getItem(/*key*/) {}
 });
 
+let SESSION_STORAGE_SUPPORTED = false;
+
+try {
+  if (typeof sessionStorage !== 'undefined') {
+    SESSION_STORAGE_SUPPORTED = true;
+  }
+} catch (e) {
+  // This can be reached with the following succession of events:
+  //
+  //   1. On Google Chrome
+  //   2. Disable 3rd-party cookies
+  //   3. Open the browser inspector
+  //   4. Open on the Ember inspector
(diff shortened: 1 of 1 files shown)


pablobm:
Thought of that, but opted for not doing it simply to avoid adding code whose purpose I don't really know. If you think it's worth doing, I can add that.

pablobm:
I rebased this today, and the build failed. Looks like three environments timed out before even trying to run the tests. Would it be possible to rerun the build, please?

rwjblue:
Done!

ember_debug/services/session.js
```diff
@@ -7,8 +7,29 @@ const Session = EmberObject.extend({
   getItem(/*key*/) {}
 });
 
+let SESSION_STORAGE_SUPPORTED = false;
+
+try {
+  if (typeof sessionStorage !== 'undefined') {
+    SESSION_STORAGE_SUPPORTED = true;
+  }
+} catch (e) {
```
rwwagner90: Do we want to log any sort of message here or does just catching the error retain the functionality, so everything appears good?

pablobm: Catching the error retains the functionality, so everything appears good.

PR #647, opened 2017-04-04
Opera Add Bower Install

krazylegz:
Adds `bower install` to the Opera directions, as it is necessary and included in the installation instructions for Chrome and Firefox.


locks:
Appreciated!

PR #654, opened 2017-04-22
ember-inspector shows wrong tooltip - fixed

siwalikm:
When Ember-Inspector is running Browser and you visit a website not running Ember, on hovering the extension logo, the tooltip says "this webpage is running Ember.js" which is wrong.

demo-url : https://url.siwalik.in/emberPRgif


teddyzeenny:
Thanks!

PR #671, opened 2017-06-10
Remove all usages of jQuery in tests

cibernox:
Before removing jQuery in the inspector itself, we have to update ember and ember-cli to 2.13

Since this repo is a bit unidiomatic, I might use some help with that.

@teddyzeenny Is there a reason for the inspector to be on an old version?

### DIFF
--- a/app/components/action-checkbox.js
+++ b/app/components/action-checkbox.js
@@ -13,7 +13,7 @@ export default Component.extend({
   },
 
   _updateElementValue() {
-    this.set('checked', this.$().prop('checked'));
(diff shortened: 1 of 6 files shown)


teddyzeenny:
@cibernox, no reason, just the challenge of finding the time to do it. However all code inside the `ember_debug` directory should support Ember 2.7+ as this is the part of the code that runs in client apps, not in the Ember Inspector env.

cibernox:
@teddyzeenny In that case this can be merged. I'll try to update the version of ember-cli in another PR.

PR #675, opened 2017-06-19
Try updating bit by bit

cibernox:
I tried updating directly to the latest Ember and Ember-cli but I ran into problems, so I decided to start things bit by bit. Expect more PRs updating Ember and Ember-cli soon.


app/index.html
```diff
@@ -3,22 +3,22 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <title>Ember Inspector</title>
+    <title>EmberInspector</title>
```
teddyzeenny: s/EmberInspector/Ember Inspector

cibernox: Done.

PR #676, opened 2017-06-19
Upgrade to glimmer2

cibernox:


tests/acceptance/info-test.js
```diff
@@ -32,11 +33,6 @@ module('Info Tab', {
 });
 
 test("Libraries are displayed correctly", async function t(assert) {
-  let infoRoute = App.__container__.lookup('route:info');
```
cibernox: Per slack conversation, in glimmer2 ember the `service:-document` is not injected until an application instance is created with the first visit, so we cannot lookup a route like that.

With this workaround we don't need to.

tests/ember_debug/view-debug-test.js
```diff
@@ -131,7 +131,6 @@ test("Simple View Tree", async function t(assert) {
   let value = tree.value;
   assert.equal(tree.children.length, 1);
   assert.equal(value.controller.name, 'App.ApplicationController');
-  assert.equal(value.viewClass, '(unknown mixin)');
```
cibernox: In glimmer seems that there is no views anymore. I don't know if it's intended or an oversight. Can anyone confirm?

cibernox: In glimmer seems that there is no views anymore. I don't know if it's intended or an oversight. Can anyone confirm?

PR #679, opened 2017-06-23
Update to 2.13 family

cibernox:
I had to fix more broken stuff dure to usage of heavy private functions related with route-recognizer. 

The only thing I didn't update was babel because I found some errors that I want to deal with on their own.


ember_debug/route-debug.js
```diff
@@ -200,11 +200,23 @@ function getURL(container, segments) {
   for (let i = 0; i < segments.length; i++) {
     let name = null;
 
-    try {
-      name = segments[i].generate();
-    } catch (e) {
-      // is dynamic
-      name = `:${segments[i].name}`;
+    if (typeof segments[i].generate !== 'function') {
+      // After changes in RouteRecognizer in >= 2.12
```
cibernox: This changed **completely** from 2.11 to 2.12.

rwjblue:
The changes here look good, happy to merge once CI is passing...

PR #681, opened 2017-06-26
Use yarn instead of npm

Turbo87:
/cc @locks @cibernox


cibernox:
@Turbo87 I did this: https://github.com/runspired/ember-run-raf/pull/16
I don't know if that is the cause if your error tho.

Turbo87:
using `--ignore-engines` has apparently fixed the issue

PR #685, opened 2017-06-30
Start removing jQuery

cibernox:
There is more work to do, but i'd rather keep merging advances as I make them.

### DIFF
--- a/ember_debug/adapters/basic.js
+++ b/ember_debug/adapters/basic.js
@@ -1,7 +1,7 @@
 /* globals requireModule */
 /* eslint no-console: 0 */
 const Ember = window.Ember;
-const { $, A, computed, RSVP, Object: EmberObject } = Ember;
+const { A, computed, RSVP, Object: EmberObject } = Ember;
 const { Promise, resolve } = RSVP;
 
 export default EmberObject.extend({
@@ -113,7 +113,7 @@ export default EmberObject.extend({
   */
   connect() {
     return new Promise((resolve, reject) => {
-      $(() => {
+      document.addEventListener('DOMContentLoaded', () => {
         if (this.isDestroyed) { reject(); }
         this.interval = setInterval(() => {
           if (document.documentElement.dataset.emberExtension) {
--- a/ember_debug/general-debug.js
+++ b/ember_debug/general-debug.js
@@ -1,7 +1,7 @@
 /* eslint no-empty:0 */
 import PortMixin from "ember-debug/mixins/port-mixin";
 const Ember = window.Ember;
-const { $, computed, Object: EmberObject, A } = Ember;
+const { computed, Object: EmberObject, A } = Ember;
 let { libraries } = Ember;
 const { oneWay } = computed;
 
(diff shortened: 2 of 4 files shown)


locks:
Tidy PRs make for tidy code reviews!

ember_debug/adapters/basic.js
```diff
@@ -113,7 +113,7 @@ export default EmberObject.extend({
   */
   connect() {
     return new Promise((resolve, reject) => {
-      $(() => {
+      document.addEventListener('DOMContentLoaded', () => {
```
bloodyowl: I think `DOMReady` equivalents probably needs a small helper, given that this event will not fire if the script is loaded after the DOM is ready.

It would require something like the following to handle the case:

```javascript
const subscribers = [];

function handleDOMContentLoaded() {
  document.removeEventListener('DOMContentLoaded', handleDOMContentLoaded)
  let func
  while ((func = subscribers.shift()) {
    setTimeout(func, 0)
  }
}

function scheduleDomReady(func) {
  if (document.readyState === "interactive" || document.readyState === "complete") {
    setTimeout(func, 0)
  } else {
   if (subscribers.length === 0) {
     document.addEventListener('DOMContentLoaded', handleDOMContentLoaded)
   }
    subscribers.push(func)
  }
}
```

cibernox: Interesting caveat. Can you make a PR, @bloodyowl ?

bloodyowl: where to? if other repositories depend on it, it might just be as interesting to use a npm package for this, don't you think?

bloodyowl: also what browsers does this target? a few IE versions have a bug where an `interactive` `document.readyState` that's not actually mean it's ready to access and manipulate

cibernox: On that recard we're lucky, we can be sure this is only going to run in chrome and firefox.

About creating a separate npm package, I think it would be overkill really. A simple utility function would be good. This inspector is not something that is going to affect people's apps.

ember_debug/general-debug.js
```diff
@@ -19,17 +19,18 @@ const GeneralDebug = EmberObject.extend(PortMixin, {
   // Keep an eye on https://github.com/ember-cli/ember-cli/issues/3045
   emberCliConfig: computed(function() {
     let config;
-    $('meta[name]').each(function() {
-      const meta = $(this);
-      let match = meta.attr('name').match(/environment$/);
+    let metas = document.querySelectorAll('meta[name]');
+    for (let i = 0; i < metas.length; i++) {
```
c0urg3tt3: to encourage best practice and avoid performance issues `metas.length` could be cached even if it's a small loop. 

```
for (let i = 0, j = metas.length; i < j; i++) {
```

bloodyowl: most JS runtimes (if not all) now inline this pattern

cibernox: Last time I checked this kind of low-lever optimizations had a despicable impact on performance, but feel free to create a PR to improve it.

ember_debug/adapters/basic.js
teddyzeenny: The basic adapter code can actually run on all browsers, including IE, Safari, and mobile browsers.

PR #687, opened 2017-06-30
Remove all usage of jQuery from `/ember_debug` folder.

cibernox:
This folder is the one that actually runs along with the user's app, so in
theory this is enough to make the inspector work on apps without jQuery.

However in following PRs I'd like to completely remove jQuery from the
extension and ensure tests are run without it, so nobody inadvertently
reintroduces it by mistake.


cibernox:
I guess this could be merged

PR #688, opened 2017-07-02
Include ember-chrome-devtools

dwickern:
The chrome extension injects [ember-chrome-devtools](https://github.com/dwickern/ember-chrome-devtools#ember-chrome-devtools) onto the page.

I used rollup to pull the formatters from my repo into the `in-page-script.js`. I didn't transpile at all so there are things like generator functions, `for-of` and `...spread`, which could be problem depending on how far back in chrome versions we want to support.

For the formatters to work you still need to "Enable custom formatters" in the DevTools settings. Maybe we could log a message about that into the console.


locks:
I assume it's not possible to programmatically ask the user to enable it?

dwickern:
Maybe... I'd start by looking at the devtools sources, find out where settings are stored, and see if there's an extension API for it

locks:
@dwickern can you look into the conflict?

rwwagner90:
@locks @dwickern what would you think about adding the functionality of ember-chrome-devtools directly into inspector, rather than trying to install the addon and use it?

dwickern:
This PR pulls in ember-chrome-devtools at build-time. It seems reasonable to keep the addon separate so users have the option of installing it.

rwwagner90:
@dwickern the idea would be that people using inspector would get this stuff automatically, thus not needing to install the addon in any of their apps. I think the main concern we had with this was the addition of rollup, and we were trying to think of ways to simplify things. That being said, I think we can probably still make this work.

sandstrom:
Since [ember-chrome-devtools](https://github.com/dwickern/ember-chrome-devtools) seems stale (no activity since 2017) and Ember is moving towards vanilla ES6 classes, does this PR still make sense?

I'm doing some PR gardening 🌱🌿 🌷 and came upon this one. If it's no longer relevant, maybe we can close it?

rwwagner90:
I agree @sandstrom. This has been open awhile, and I think there is less of a need for it, so going to go ahead and close for now.

PR #693, opened 2017-08-05
Remove / from the root URL since we're using relative URLs

teddyzeenny:
Part 2 (of 2) to fix the Ember Inspector on master.
cc @cibernox

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -5,7 +5,7 @@ module.exports = function(environment) {
   var ENV = {
     modulePrefix: 'ember-inspector',
     environment: environment,
-    rootURL: '/',
+    rootURL: '',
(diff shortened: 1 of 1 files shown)


cibernox:
I wasn't aware that there was a difference between `'/'` and `''`

teddyzeenny:
@cibernox `rootURL` is prepended to the assets in [index.html](https://github.com/emberjs/ember-inspector/blob/master/app/index.html#L12). A `/` makes the assets URL absolute, whereas an empty string keeps them relative to the current URL. Generally you'll want a `/` prepended to assets, which is probably why it's the default, especially when rewriting history to reflect Ember routes - but that's not the case in the inspector.

PR #702, opened 2017-08-13
Fix event delegation change

teddyzeenny:
Continues the work in https://github.com/emberjs/ember-inspector/pull/698 by simulating mouseleave as well and fixes the failing test.

cc @cibernox


cibernox:
I'll be back home tonight to check it.

PR #704, opened 2017-08-15
Don't let mixin names wrap in the object inspector

teddyzeenny:
Some mixins' `toString` just prints the entire function - so we don't let it wrap.

Before:
screen shot 2017-08-15 at 2 20 51 pm

After:
screen shot 2017-08-15 at 2 20 30 pm

### DIFF
--- a/app/styles/mixin.scss
+++ b/app/styles/mixin.scss
@@ -14,6 +14,9 @@
   background-color: #f0f0f0;
   cursor: default;
   user-select: none;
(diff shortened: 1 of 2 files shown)


cibernox:
@rwjblue @teddyzeenny I've created https://github.com/emberjs/ember.js/issues/15586 to track the fix of this in Ember. A comment to link that workaround with the issue would be nice so it can be removed eventually.

PR #667, opened 2017-06-02
Firefox WebExtension

clarkbw:
This PR fixes #666 

There is still more work to do to migrate the build / packaging of Firefox over to the WebExtension, this PR is focused on updating the chrome WebExtension so that it works in Firefox.

PR #715, opened 2017-09-04
[WIP] Firefox WebExtension

teddyzeenny:
Continues the work started by @clarkbw in https://github.com/emberjs/ember-inspector/pull/667. cc @rpl 

Remaining things:
- [x] Late Iframe detection (which Chrome does via `chrome.devtools.inspectedWindow.onResourceAdded` but FF doesn't support it at the moment). [REMOVED SUPPORT IN FF].
- [x] Transition the old inspector snapshot to use FF WebExtension as well so it doesn't break for users of old Ember versions. DONE IN https://github.com/emberjs/ember-inspector/pull/716.
- [x] Cleanup


app/adapters/chrome.js
```diff
@@ -1,5 +1,21 @@
 import WebExtension from "./web-extension";
 
 export default WebExtension.extend({
-  name: 'chome'
+  name: 'chome',
```
locks: should this be `chrome`?

teddyzeenny: 👍

topaxi:
Woohoo thanks everyone for the awesome work! 💯  🎉 ❤️

Issue #677, opened 2017-06-20
Search View Tree


shoxter:
Sometimes we have hundreds of components in the view tree at a given time and it would be extremely useful to be able to search the contents of that view tree for a specific component.

Thoughts?

kamikaz1k:
+1

kamikaz1k:
I put up a PR to address this issue. If you are still interested in this functionality please drop into the PR.

rwwagner90:
We have search in both the view tree and components tree, so this is done.

PR #720, opened 2017-10-07
View tree search filter

kamikaz1k:
When dealing with a complicated view tree, it can be cumbersome scrolling through all the views trying to identify the one that is of interest.

To help make finding it easier I added an input field to filter the list that is visible in the scroll view.
It naively looks for the substring to be present in the name of the view to determine a match.
The filter also works with the components included in the view tree.

Demo:
view-tree-filter

Refs Issue https://github.com/emberjs/ember-inspector/issues/677

Looking for feedback around contribution guidelines, and if this solution fits the pattern being used in the application. Otherwise also accepting suggestions for improving the implementation.

### DIFF
--- a/tests/acceptance/view-tree-test.js
+++ b/tests/acceptance/view-tree-test.js
@@ -21,6 +21,10 @@ module('View Tree Tab', {
   }
 });
 
+function textFor(selector, context) {
+  return find(selector, context).textContent.trim();
+}
+
 let treeId = 0;
 function viewNodeFactory(props) {
   if (!props.template) {
@@ -125,10 +129,6 @@ test("It should correctly display the view tree", async function(assert) {
   let viewClassNames = [];
   let durations = [];
 
-  function textFor(selector, context) {
-    return find(selector, context).textContent.trim();
-  }
-
   [...treeNodes].forEach(function(node) {
     templateNames.push(textFor('.js-view-template', node));
     controllerNames.push(textFor('.js-view-controller', node));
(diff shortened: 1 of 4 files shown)


tests/acceptance/view-tree-test.js
```diff
@@ -186,6 +187,71 @@ test("It should correctly display the view tree", async function(assert) {
...
+  await visit('/');
+  run(() => {
+    port.trigger('view:viewTree', { tree: viewTree });
+  });
+  await wait();
+
+  let treeNodes = findAll('.js-view-tree-item');
+  assert.equal(treeNodes.length, 3, 'expected some tree nodes');
+
+  fillIn('.js-filter-views input', 'post');
```
teddyzeenny: `await`

tests/acceptance/view-tree-test.js
```diff
@@ -186,6 +187,71 @@ test("It should correctly display the view tree", async function(assert) {
...
+  treeNodes = findAll('.js-view-tree-item');
+  assert.equal(treeNodes.length, 1, 'expected filtered tree nodes');
+
+  let controllerNames = [];
+  let templateNames = [];
+  let modelNames = [];
+  let viewClassNames = [];
+  let durations = [];
+
+  function textFor(selector, context) {
```
teddyzeenny: Can you move this function outside the tests callbacks so we can reuse the same one to avoid duplication?

kamikaz1k:
Thanks for being so quick to provide feedback :)
I made the changes requested to address them.

tests/acceptance/view-tree-test.js
kamikaz1k: I didn't realize we had to use `await` for async helpers because I thought the test framework automatically took care of making it all synchronous. I read the docs again, and is it that the awareness comes from preceding async tasks, rather than ensuring downstream logic is synchronous? (I hope I said that clearly...)

In other words:
```
asyncFunction();
// Async helper is aware of the above async behaviour and will wait
fillIn();
// But it doesn't make sure the stuff below gets executed in a synchronous way
some = synchronousLogic(); 
```
So to fix it, I would have to use either `await` or `andThen`?
```
await fillIn();
// or
andThen(() => { some = synchronousLogic() });
```

teddyzeenny: Yes precisely, you got it. `await` is used to avoid using `andThen` down the line.

PR #729, opened 2017-10-25
Fix Iframe detection.

Serabe:
Iframe was not working because of wrong context in callback.


nathanhammond:
Hey @teddyzeenny! What's the release cadence on ember-inspector? (When might we expect this to be distributed?)

teddyzeenny:
@nathanhammond either this weekend or the next at the latest.

PR #742, opened 2017-12-21
preserve model type hiding

### LINKED ISSUE #741: [feature discussion] ability to configure value for "Hide Empty Model Types"
# use case
As a developer in a large app, we use a number of different ember data models. Often when developing one section, I only care about seeing one's loaded in the store. Thus I would like to be able to save a preference to toggle this on.
I would see this as simple checkbox navigated to from Ember Inspector's options link:
screen shot 2017-12-20 at 1 50 27 pm

# prior work
https://github.com/emberjs/ember-inspector/pull/458/files

efx:
Resolve #741.
This is a bad PR in that it does not test the added functionality. I tried adding a test as you can see [here](https://github.com/efx/ember-inspector/tree/fix-741-broken-tests) but `moduleFor` blew up with
screen shot 2017-12-21 at 5 54 55 pm

Given end of day before a break, I thought I'd just PR the functionality for whenever you get a chance to review. Thanks for the pointers in the issue. It made this trivial.

### DIFF
--- a/app/controllers/model-types.js
+++ b/app/controllers/model-types.js
@@ -1,14 +1,28 @@
-import Ember from "ember";
+import Ember from 'ember';
+import LocalStorageService from 'ember-inspector/services/storage/local';
 const { Controller, computed, get, inject } = Ember;
 const { sort } = computed;
-const { controller } = inject;
+const { controller, service } = inject;
 
 export default Controller.extend({
   application: controller(),
   navWidth: 180,
   sortProperties: ['name'],
+  storage: service(`storage/${LocalStorageService.SUPPORTED ? 'local' : 'memory'}`),
   options: {
-    hideEmptyModelTypes: false
+    hideEmptyModelTypes: computed({
+      get() {
+        return !!this.get('storage').getItem('are-model-types-hidden');
+      },
+      set(key, value) {
+        if (!value) {
+          this.get('storage').removeItem('are-model-types-hidden');
+        } else {
+          this.get('storage').setItem('are-model-types-hidden', value);
+        }
+        return value;
+      }
+    })
   },
 
   sorted: sort('filtered', 'sortProperties'),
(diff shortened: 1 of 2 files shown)


app/controllers/model-types.js
```diff
@@ -1,14 +1,30 @@
...
   application: controller(),
   navWidth: 180,
   sortProperties: ['name'],
+  storage: service(`storage/${LocalStorageService.SUPPORTED ? 'local' : 'memory'}`),
   options: {
-    hideEmptyModelTypes: false
+    hideEmptyModelTypes: computed({
+      get() {
+        // TODO fix test resolution lookup
+        if (!this.get) { return false; }
```
teddyzeenny: This is not needed.

app/controllers/model-types.js
```diff
@@ -1,14 +1,30 @@
...
+const { controller, service } = inject;
 
 export default Controller.extend({
   application: controller(),
   navWidth: 180,
   sortProperties: ['name'],
+  storage: service(`storage/${LocalStorageService.SUPPORTED ? 'local' : 'memory'}`),
   options: {
-    hideEmptyModelTypes: false
+    hideEmptyModelTypes: computed({
```
teddyzeenny: This it is now a CP, you'll need to move it outside the `options` hash into its own property on the component. When you do that, make sure to update all references to it [in the controller](https://github.com/emberjs/ember-inspector/blob/825f3de9244b952f564789d9ba2241e912e51eb8/app/controllers/model-types.js#L16-L18) and [in the toolbar](https://github.com/emberjs/ember-inspector/blob/825f3de9244b952f564789d9ba2241e912e51eb8/app/templates/model-types-toolbar.hbs#L3).

efx:
@teddyzeenny comments make sense, thank you. I can squash once things look good.

app/controllers/model-types.js
```diff
@@ -1,21 +1,34 @@
...
+    set(key, value) {
+      if (!value) {
+        this.get('storage').removeItem('are-model-types-hidden');
+      } else {
+        this.get('storage').setItem('are-model-types-hidden', value);
+      }
+      return value;
+    }
+  }),
+  options: {},
```
teddyzeenny: This is not needed anymore.

efx:
grand, thanks @teddyzeenny !

PR #743, opened 2018-01-16
Update Inspect Views Icon

nummi:
The magnifying glass was confusing me so I created an icon that is similar to the inspect icons in Chrome and Firefox.

Original:
original

New Chrome:
new - chrome

New Firefox:
new - firefox


locks:
I did the `sudo: required` thing and we are good to go!

PR #748, opened 2018-02-22
Fix loading main outlet

rwwagner90:
I was trying running the dev mode of this and got errors on my site. I am assuming `getRoot` changed in Ember 3.1 or something in glimmer, but this should support both.


teddyzeenny:
Thanks! We still need to fix the View Tree components since [getTemplate](https://github.com/emberjs/ember-inspector/blob/66e7ec3d830cfbfbf2294bca0e868bf363fa8742/ember_debug/libs/glimmer-tree.js#L356) seems to have been removed from glimmer environment.

PR #753, opened 2018-02-23
Allow for empty mixins

### LINKED ISSUE #752: Empty mixins break component selector
If an app includes an empty mixin, selecting a component using either the magnifying glass and clicking on the component name or clicking the View / Component in the View Tree causes a call stack in ember-inspector in `propertiesForMixin(mixin)` where it expects all mixins to have a mixins array.

Version 0.6.2 of ember-component-css includes such an empty mixin and so breaks ember-inspector for apps that use it.

You can see the behaviour when running https://github.com/XuluWarrior/ember-component-css-test

I have already raised https://github.com/ebryn/ember-component-css/issues/274 but it seems that ember-inspector should be robust enough to deal with empty mixins.

XuluWarrior:
Fixes #752

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -429,11 +429,13 @@ export default EmberObject.extend(PortMixin, {
 function propertiesForMixin(mixin) {
   let properties = [];
 
-  mixin.mixins.forEach(mixin => {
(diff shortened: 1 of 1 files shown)


rwwagner90:
Thanks!

PR #757, opened 2018-02-25
Fix file paths displayed in the Routes tab

omarhamdan:
- `emberCliConfig` was always false, using `getAttribute` instead of `attr` fixes the issue.
- From ember 2.7.0 the libraries are always found in `libraries._registry`, so we remove support for older versions.
- Documented `ember_debug/general-debug.js`.

### DIFF
--- a/ember_debug/general-debug.js
+++ b/ember_debug/general-debug.js
@@ -1,24 +1,41 @@
 /* eslint no-empty:0 */
 import PortMixin from "ember-debug/mixins/port-mixin";
 const Ember = window.Ember;
-const { computed, Object: EmberObject, A } = Ember;
+const { computed, Object: EmberObject } = Ember;
 let { libraries } = Ember;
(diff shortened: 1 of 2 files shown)


skeletons/web-extension/scripts/in-page-script.js
```diff
@@ -29,17 +29,9 @@
...
-      // but now it's found under _registry.
-      if (libraries._registry) {
-        libraries = libraries._registry;
-      }
-
-      var versions = Array.prototype.slice.call(libraries, 0);
       window.postMessage({
         type: 'emberVersion',
-        versions: versions
+        versions: libraries._registry.slice()
```
teddyzeenny: This script runs for all Ember versions (even < 2.7), so let's keep the backwards compatibility here.

teddyzeenny:
😍

PR #762, opened 2018-02-27
Ember 3.0, dep updates, testing updates, and codemods

rwwagner90:
We did a lot of work here to update things to newer standards and fix the usages of the `__container__` private API to use `this.owner` API. We also converted `ember-native-dom-helpers` code to `@ember/test-helpers` code, removed some dead code and wrote our own custom logic in test helpers to setup and destroy the app.

We still have one failing test, but there is a lot to review here, so I wanted to get it up for people to review while we poke at the one remaining failure. Thanks to @rwjblue and @teddyzeenny for their extensive help figuring out some of the issues I encountered!


ember_debug/view-debug.js
```diff
@@ -221,8 +220,9 @@ export default EmberObject.extend(PortMixin, {
         if (this.glimmerTree) {
           this.glimmerTree.highlightLayer(viewElem.id);
         } else {
-          this.highlightView(viewElem[0]);
+          this.highlightView(viewElem);
```
teddyzeenny: 👍

PR #765, opened 2018-03-03
Fix inspector reset on client app's reset and destroy

teddyzeenny:


app/initializers/setup.js
```diff
@@ -0,0 +1,46 @@
...
+    instance.adapter = '{{EMBER_DIST}}';
+
+    // register and inject adapter
+    let Adapter;
+    if (typeOf(instance.adapter) === 'string') {
+      Adapter = instance.resolveRegistration(`adapter:${instance.adapter}`);
+    } else {
+      Adapter = instance.adapter;
+    }
+    register(instance, 'adapter:main', Adapter);
```
rwjblue: I know these manual registrations are not new, but I’m curious if we actually need them? I would think that putting a file at `app/adapters/main.js` (Or `app/adapter.js`) would “just work”?

ember_debug/main.js
```diff
@@ -12,11 +12,40 @@ import DeprecationDebug from 'ember-debug/deprecation-debug';
...
+   * @type {Application}
+   */
+  _application: null,
 
   owner: null,
+
+  /**
+   * We use the application's id instead of the owner's id so that we use the same inspector
+   * instance for the same application even if it was reset (owner changes on reset).
+   *
```
rwjblue: Thank you for this write up, I was wondering why we needed it...

app/initializers/setup.js
teddyzeenny: Good point. I think these should be turned into services, though it's probably not as simple as putting a file in the correct place since the `adapter` is registered as a different class depending on the environment (Chrome/FF/bookmarklet/tests).

rwwagner90: Are these adapters like normal adapters for ember-data in an application? If so, I would think they would be automatically registered.

teddyzeenny: I'm going to merge the PR for now as we need this fix to release 3.0, we can look into refactoring this later.

Issue #590, opened 2016-09-10
Remove jQuery


Turbo87:
@rwjblue is working on making jQuery an optional dependency of Ember. Once that has progressed far enough we should investigate what we need to do to drop our dependency on jQuery too.

https://github.com/emberjs/ember-inspector/blob/master/shared/in-page-script.js for example will need to be adjusted to a world without jQuery.

rwjblue:
FWIW, it is already possible to use Ember without jQuery (you just need a custom event dispatcher at the moment).

locks:
A custom event dispatcher by ours truly: https://github.com/rwjblue/ember-native-dom-event-dispatcher

cibernox:
I will try to work on this as soon as I can put a couple hours in a row

acorncom:
This was resolved by #768

rwwagner90:
@acorncom technically, this isn't 100% done. While our jQuery usages have been removed, there is something still using it, and if you remove it from vendor, things break.

cibernox:
@rwwagner90 When I removed jQuery, I did it from the parts of the app that ran along with the app, so the addon could be useful in apps without jQuery, but I didn't care about usages of jQuery INSIDE the addon itself. 

That said, personally I'd remove it completely and ensure CI don't have it, so we can be sure we don't use it, not even by mistake.

rwwagner90:
@cibernox yeah, that's what I mean. Inspector itself, not the part that runs in the consuming app. Mainly want to remove it to align with the removal we have been doing across the learning team.

acorncom:
Reopening in the interests of tracking the remaining work needed to remove internal jQuery usage

PR #768, opened 2018-03-08
Remove jQuery from app

rwwagner90:
Also had to switch to a new context menu plugin. This should get us closer to #590

screen shot 2018-03-08 at 10 45 34 am

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,6 +14,9 @@ module.exports = {
   env: {
     browser: true
   },
+  globals: {
+    basicContext: false
+  },
   rules: {
     /* Possible Errors */
 
@@ -121,6 +124,8 @@ module.exports = {
     'prefer-template': 2,
     'require-yield': 2,
     'template-curly-spacing': 2,
+
+    'ember/no-jquery': 'error',
 
     // Temporarily turn these off
     'ember/avoid-leaking-state-in-ember-objects': 'off',
--- a/app/components/drag-handle.js
+++ b/app/components/drag-handle.js
@@ -1,4 +1,3 @@
-import $ from 'jquery';
 import { equal } from '@ember/object/computed';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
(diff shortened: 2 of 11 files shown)


.eslintrc.js
```diff
@@ -14,6 +14,9 @@ module.exports = {
   env: {
     browser: true
   },
+  globals: {
+    basicContext: false
```
rwjblue: 🤔 what is `basicContext`?

app/components/drag-handle.js
```diff
@@ -51,32 +50,18 @@ export default Component.extend({
...
-      position -= this.get('left');
-      if (position >= this.get('minWidth') && position <= this.get('maxWidth')) {
-        this.set('position', position);
-        this.get('on-drag')(position);
-      }
-    })
-      .on(`mouseup.${namespace} mouseleave.${namespace}`, () => {
-        this.stopDragging();
-      });
+    document.body.addEventListener(`mousemove.${namespace}`, this._mouseMoveHandler);
```
rwjblue: These also need to be cleaned up in`willDestroy` or `willDestroyElement`

package.json
```diff
@@ -36,6 +36,7 @@
     "aws-sdk": "^2.3.3",
     "babel-eslint": "^8.2.2",
     "babel-plugin-transform-es2015-modules-amd": "^6.24.0",
+    "basiccontext": "^3.5.1"
```
rwjblue: missing a comma here

teddyzeenny:
Thanks! Can you please add a screenshot of the new context menu?

.eslintrc.js
rwwagner90: It is the new contextMenu plugin we are using.

app/components/drag-handle.js
rwwagner90: `willDestroyElement` calls `this.stopDragging();` which has the `removeEventListener` calls. Did I mess them up?

package.json
rwwagner90: 👍

rwwagner90:
@teddyzeenny I added a screenshot 👍

app/components/drag-handle.js
rwjblue: Ahh, I see, thank you for explaining!

Issue #771, opened 2018-03-12
Implement dark theme


rwwagner90:
Copying over data from slack:

it looks like we actually can detect the devTools theme from `chrome.devtools.panels.themeName`. @teddy where would be the best spot to check this?

teddy [6:26 AM]
we can add a CP `isDarkTheme` or `theme` on the adapter
https://github.com/emberjs/ember-inspector/blob/1aa8f03e1f00a6166042b61914a0ccdf03a4512b/app/adapters/web-extension.js
https://github.com/emberjs/ember-inspector/blob/1aa8f03e1f00a6166042b61914a0ccdf03a4512b/app/adapters/chrome.js

Then check the theme in `x-app`
https://github.com/emberjs/ember-inspector/blob/1aa8f03e1f00a6166042b61914a0ccdf03a4512b/app/components/x-app.js

We should use CSS variables to switch the colors based on this.

pbishop16:
@rwwagner90 I would like to tackle this.

rwwagner90:
@pbishop16 sure thing! We would love the help 😃 . Please feel free to ping me with questions on slack. I'm happy to help however I can.

pbishop16:
@rwwagner90 Greatly appreciated. Will do.

nummi:
@rwwagner90 close?

rwwagner90:
This is done now 👍

PR #782, opened 2018-04-10
Implement dark theme

pbishop16:
Addresses #771 
- Implement support for chrome devtools dark theme
- Minor CSS color refinements

It is my hope that the approach implemented may provide support for additional themes in the future. Will gladly address any and all issues/concerns, as the approach may not be ideal.


rwwagner90:
@teddyzeenny @locks what do you think about this approach? I think we initially were thinking of hard coded overrides with CSS variables, but @pbishop16 did some cool stuff to calculate inverse colors here. I'm not sure if we want to keep that or hard code some colors, but this stuff is looking great!

app/constants/theme-colors.js
```diff
@@ -0,0 +1,83 @@
...
+  spec01: '#ffffee',
+  spec02: '#ff0000',
+  spec03: '#f23818',
+  spec04: '#990099',
+  spec05: '#990000',
+  spec06: '#6e68c6',
+  spec07: '#4281eb',
+  spec08: '#3879d9',
+  spec09: '#2b7fb3',
+  spec10: '#000099',
```
nummi: Wow. We need to wrangle the colors.

rwwagner90: Definitely open to some help or suggestions on this stuff 😄

pbishop16:
@sivakumar-kailasam That sounds like a great idea. Since adjusting the color palette was not the focus, I definitely made sure to only use the existing colors. As @nummi pointed out, there are quite a few colors.

rwwagner90:
@pbishop16 I just merged some style updates, which resulted in a bunch of conflicts, sorry about that 😞

pbishop16:
@rwwagner90 No worries. I will get this cleaned up.

nummi:
Quick thought: I think inverting the colors is a good first pass but eventually we'll probably end up creating component specific variables where simple inversion doesn't work. For example, in the navigation and toolbar. I think it looks better to actually keep the blue and white text rather than inverting it:

current:

current

better?
proposed

``` javascript
export const LightTheme = {
  navLinkActiveBackground: 'thatbluecolor',
  navLinkActiveText: white,
};
export const DarkTheme = {
  navLinkActiveBackground: 'thatbluecolor',
  navLinkActiveText: white,
}
```

rwwagner90:
@nummi what happens in the normal dark Chrome DevTools? We may want to closely match that, but we also wouldn't necessarily have to. We should likely consider accessibility and choose colors good for color blind people.

nummi:
> @nummi what happens in the normal dark Chrome DevTools? We may want to closely match that, but we also wouldn't necessarily have to. We should likely consider accessibility and choose colors good for color blind people.

Good point. In Chrome they seem to invert, Firefox they do not.

Chrome:

Color contrast ratio: 6.78:1

screen shot 2018-04-12 at 1 07 37 pm

Firefox:

Color contrast ratio: 8.33:1

screen shot 2018-04-12 at 1 07 02 pm

We could just darken our blue a bit like Firefox does. I guess it's really a matter or personal taste — I think white on blue looks better than black on orange.

Here's what the colors look like with the most common form of color blindness:

screen shot 2018-04-12 at 1 12 25 pm

rwwagner90:
We likely want the highest contrast we can do, but we could always get a dark them in and iterate from there.

rwwagner90:
@pbishop16 I just merged some fixes for our failing builds. Do you mind rebasing with master please? Hopefully our builds should be green after that!

pbishop16:
@rwwagner90 Ok thanks for the heads up. I will be sure to rebase.

rwwagner90:
@pbishop16 @nummi what do you guys think about the color inversion and accessibility? Do we want to just add a dark class and switch CSS variables instead and provide explicit color choices?

nummi:
I think the current inversion of colors has plenty of contrast. Over time we can improve on the design.

First pass should be reducing the amount of colors.

rwwagner90:
I'm okay with merging this, if we are okay with using JS to invert colors, rather than having CSS variables we override to a specific dark theme. @locks @teddyzeenny what do you guys think? If we are okay getting this in and iterating from there, I am in favor of merging. Having a dark theme is better than not having one 😄

app/adapters/web-extension.js
```diff
@@ -55,6 +61,23 @@ export default BasicAdapter.extend({
     this.onResourceAdded();
   },
 
+  _setThemeColors() {
+    const rootElement = document.querySelector('html');
+    let theme;
+
+    if (chrome.devtools.panels.themeName === 'dark') {
```
nummi: How do browsers other than Chrome handle this statement?

rwwagner90: I'm not sure, but it must be okay because we are doing several `chrome.devtools` things in several methods in this file.

pbishop16: @nummi @rwwagner90 I did fail to verify this. Thank you for the wakeup call on this. I have now verified Opera and Firefox. Opera handles this statement without issues. It works in Firefox but the add-on requires a manual reload. The Devtool extensions in both Opera and Chrome seem to be automatically forced to reload when the settings are adjusted. 

I will work on fixing the issue in Firefox.

pbishop16: I just noticed that the React-Devtools requires the same manual reload in order to respond to theme changes within Firefox's devtool settings. Firefox might be a bit tricky to deal with regarding this.

rwwagner90: @pbishop16 is Firefox okay if you start with normal or dark themes? It just has issues if you try to switch themes?

pbishop16: @rwwagner90 Yeah, it sure is. It displays light or dark accordingly.

rwwagner90: @pbishop16 I think that is fine then. We should do whatever the React ones do.

nummi:
@pbishop16, what work is remaining? Anything I can help with?

pbishop16:
@nummi The PR is pretty much good to go, pending any other concerns you or @rwwagner90 may have. 

Also, I would be more than happy to take on any additional clean up or fixes you all may desire regarding implementation. (ie. adjust the color pallet, adjust for accessibility, etc.) I know my solution is not ideal and I would be glad readjust to adding a dark class and switching CSS variables as you all originally envisioned.

nummi:
🚢 🎉

pbishop16:
@rwwagner90 @nummi My good gentlemen, I went ahead and took some time this morning to implement CSS variables and theme class toggling. Please let me know if there are any concerns. Everything tested well on my end; no impact on functionality.

rwwagner90:
@pbishop16 awesome! Thanks for doing that. Did we factor in accessibility and contrast to the colors or should we follow up with that later?

pbishop16:
@rwwagner90 I was thinking about saving that for the followup. I didn't want to break too much at this time and I understand that finding the right accessibility color mix can be tricky.

app/adapters/web-extension.js
```diff
@@ -55,6 +57,19 @@ export default BasicAdapter.extend({
...
+    const element = document.querySelector('body');
+    let theme;
+
+    if (chrome.devtools.panels.themeName === 'dark') {
+      theme = 'theme--dark';
+    } else {
+      theme = 'theme--light';
+    }
+
+    element.setAttribute('class', theme);
```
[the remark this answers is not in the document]
pbishop16: Definitely makes sense. Thank you.

tryane:
Hi guys, how enabled dark theme for ember inspector, because is still light for me 
chrome 69

rwwagner90:
@tryane you have to enable dark theme for your entire dev tools, then ember inspector will turn dark to match

tryane:
Sorry, but I use dark theme and ember inspector for a long time and I never see the dark theme on the ember inspector
After another test, when I go on the Ember time for the first time, I can see the dark theme, but it switch immediately on the light theme 😢

rwwagner90:
@tryane what browser and version? Everything is working perfectly for me in Chrome and Firefox. Try a few hard refreshes with the inspector open.

ncoquelet:
(I have switch on my real github account)
I use Chrome 69, Inspector is install from the web store, id = bmdblncegkenkacieihfhpjfppoconhi, version 3.3.2
In my file system, inside the extension I have two directories : panes-0-0-0 and panes-2-7-0
And if I inspect the Ember Inspector tab, I can see I use the panes-0-0-0 code

ncoquelet:
Okay, I got it.
We use Ember 1.13 !
If I go on the Discourse forum, Ember Inspector is dark ... 😢

rwwagner90:
@ncoquelet yeah, if you're pre 2.7, you do not get any of the new inspector features. It's too hard to support all the old stuff. If you need help updating past 1.13, Ship Shape would love to help you 😄. https://shipshape.io

ncoquelet:
Thank you @rwwagner90 
We work on a huge project : 694 js files (33411 loc) and 329 HtmlBars files (17358 loc). And as we started with ember 1.5.0, all is build with  grunt. 
The full migration to 2.0 is on the track but we need time, as usual.
Regards

rwwagner90:
@ncoquelet sometimes it is helpful to have an outside firm come in and help with updates, while your team continues to work on feature development. Ship Shape would love to work with you, if you are interested 😄

PR #784, opened 2018-04-11
search-field  Component Bug Fix and Acceptance Tests

nummi:
Changed clear button from DOM removal to 0 opacity because this causes problems in testing. When clicking on a button, it gets focus. But if you remove a focused element from the DOM, the entire window loses focus.

### DIFF
--- a/app/components/search-field.js
+++ b/app/components/search-field.js
@@ -1,10 +1,9 @@
 import Component from '@ember/component';
 
 export default Component.extend({
-  tagName: '',
-
   actions: {
     clear() {
+      this.element.querySelector('input').focus();
       this.set('value', '');
     }
   }
--- a/app/templates/components/search-field.hbs
+++ b/app/templates/components/search-field.hbs
@@ -1,6 +1,6 @@
 {{input value=value type="text" placeholder="Search"}}
-{{#if search}}
-  <button {{action "clear"}} class="toolbar__icon-button toolbar__search-clear-button" title="clear">
+{{#if value}}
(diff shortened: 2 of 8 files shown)


rwwagner90:
@nummi did this not cause issues before or were we just not testing before? Wouldn't it be shown in tests as well, when the if condition is true?

rwwagner90:
Oh, yeah, should it be `if searchValue`? I think that was the name, right?

nummi:
> Oh, yeah, should it be if searchValue? I think that was the name, right?

Correct. The component is called like: `{{search-field value=searchValue}}`

rwwagner90:
So can we put in the appropriate `if` statement? Trying to determine why we want to style opacity, rather than using an `if`.

rwwagner90:
@teddyzeenny any thoughts on this? If you are okay with the opacity approach, we can keep it as is. I'd like to figure out how to use an `if`, as I think that is more clear to people jumping into the code later, but getting this in and iterating later is also fine with me.

teddyzeenny:
@rwwagner90 @nummi I'll take a look during the weekend. Ideally we can fix this at the source of the error on [`x-app`](https://github.com/emberjs/ember-inspector/blob/317f4d1b1297053aa384570761fa84fc122a3ee1/app/components/x-app.js#L27-L37) by maybe scheduling or debouncing the `active` toggle using the run loop? If it works it would be better than removing the `if` as a workaround IMO.

rwwagner90:
@nummi can you explore that option?

rwwagner90:
That could work. @teddyzeenny would that work for you?

teddyzeenny:
That works! :shipit:

rwwagner90:
Thanks @nummi!

PR #785, opened 2018-04-13
Initial component tree implementation

Bestra:
This is our first iteration of a new component tree view.
### Initial features
- Nodes can be opened or closed
- Uses the existing object inspector.  Clicking a component's tag will open that component. 
- Clicking the 'eye' icon will scroll the corresponding element into view.
screen shot 2018-04-13 at 8 59 42 am

- The search bar will filter components.  Matched components will be indented at the top level, keeping indent levels for their children
screen shot 2018-04-13 at 8 57 23 am

The component tree shares some common behavior with the view tree for the time being:
- The preview layer 'mostly' works.  If a user clicks the 'component=' link it will highlight the corresponding element in the tree.  As of this PR the tree does not expand itself to show the selected item if it's not already visible (play around and you'll see what I mean).
screen shot 2018-04-13 at 9 07 35 am
Pretty shortly we'll probably write a more specific `component-debug` implementation to make the preview layer play better with the component tree feature set.

### Some stuff that probably needs to happen
- [X] Currently there's no "expand all/collapse all" feature, which we'll probably need
- [X] It'd be good to make the entire component row clickable for selecting, not just the component name itself
- [X] We show the views in the tree but they don't actually respond to clicks.  Aside from lacking curlies, they shouldn't have an 'eye' icon, hover state, etc.


rwwagner90:
This is awesome. Thanks @Bestra! I will try to do an in depth review and play with it some this weekend. As for the things you mentioned that "probably need to happen", do you think we should have an issue to track the next steps, to make sure we tackle all those? I am okay getting this initial work in and iterating further later, if @teddyzeenny and @locks are okay with it.

Bestra:
@rwwagner90 I'd say if we decide to go forward with this PR before fixing those issues we'll definitely want to properly track them.  @nummi is a good sounding board for design/UX problems and I'll get his feedback along the way. 

I'm totally comfortable trying to tie up loose ends before we land this in master, too.  It'll be good to have another set of eyes triaging things for severity - I've looked at a few different sites a sample size of 2 is pretty low.

rwwagner90:
@Bestra going ahead and addressing concerns before merging might be the safest. Let's sync up this weekend sometime.

ef4:
This looks really nice, I will be very happy when this can release.

lifeart:
Maybe we should create tree view using new syntaxis?

Instead of this:
```
{{power-select}}
    {{basic-dropdown}}
        {{basic-dropdown/trigger}}
```
use this:
```
<PowerSelect>
    <BasicDropdown>
        <BasicDropdown::trigger>
```
?

rwwagner90:
@lifeart we definitely will need to add some sort of check and change the display based on if you are using glimmer components or not. As of now, this is the correct format, and we need to support all Ember versions.

Bestra:
- Note the icons next to the search box for expand/contract all
- Views are no longer clickable
- When inspecting a component, the tree will expand to show it if necessary.
screen shot 2018-04-16 at 5 03 10 pm

rwwagner90:
@Bestra @nummi I just got the builds back to green. Can you rebase with master to try to get this build green as well please?

Bestra:
@rwwagner90 I need to add unit tests for the 'expand all / contract all' behavior but I don't think that should stop folks from reviewing now.  I'll make sure the tests are in before we merge (or shortly after if somehow the review process goes incredibly fast)

app/controllers/component-tree.js
```diff
@@ -0,0 +1,281 @@
...
+  reads
+} from '@ember/object/computed';
+import {
+  isEmpty
+} from '@ember/utils';
+
+/**
+ * ComponentViewItem is used to represent the flattened nodes in the component tree
+ */
+let ComponentViewItem = EmberObject.extend({
```
rwwagner90: Should we maybe have a separate file for `ComponentViewItem` and import it? This file is a bit long, and might help with separation of concerns if we split the component item and the controller into separate files.

app/routes/component-tree.js
```diff
@@ -0,0 +1,47 @@
...
+    this.get('port').off('view:inspectDOMElement', this, this.inspectDOMElement);
+    this.get('port').off('view:inspectComponent', this, this.inspectComponent);
+
+  },
+
+  setViewTree(options) {
+    this.set('controller.viewTree', options.tree);
+  },
+
+  startInspecting() {
```
rwwagner90: Could we combine `startInspecting` and `stopInspecting` into one, perhaps `toggleInspecting` and pass a boolean for true or false?

app/routes/component-tree.js
```diff
@@ -0,0 +1,47 @@
+import TabRoute from "ember-inspector/routes/tab";
+
+export default TabRoute.extend({
+  model() {
+    return [];
```
rwwagner90: I'm curious, do we need to return an empty array here?

rwwagner90:
@Bestra 

> Remove unused model hook in component-tree route

I'm not sure if it was unused, just was asking if we needed that empty array. If we don't we should remove, but if we do, and use it to push items into the empty array or something, it's fine to keep.

app/routes/component-tree.js
Bestra: I think it was copy-pasta from the `view-tree` based on my recollection, but I'll check.

app/routes/component-tree.js
```diff
@@ -0,0 +1,47 @@
...
+    this.get('port').on('view:inspectDOMElement', this, this.inspectDOMElement);
+    this.get('port').on('view:inspectComponent', this, this.inspectComponent);
+    this.get('port').send('view:setOptions', { options: this.get('controller.options') });
+    this.get('port').send('view:getTree');
+  },
+
+  deactivate() {
+    this.get('port').off('view:viewTree', this, this.setViewTree);
+    this.get('port').off('view:stopInspecting', this, this.stopInspecting);
+    this.get('port').off('view:startInspecting', this, this.startInspecting);
```
rwwagner90: Sorry, I wasn't clear on my feedback for this. I mean instead of having a `startInspecting` and `stopInspecting` function, have a single `toggleInspecting` function. Then here, instead of doing `this.startInspecting` you would do `this.toggleInspecting(true)` etc. Just trying to have one function to do starting and stopping, rather than duplicating, but we definitely don't want a third function. Maybe we should just keep the two, and not worry about it.

Bestra: Ha! Easily done.  I like that version much better than the one I wrote

stefanpenner:
🎉

PR #787, opened 2018-04-16
[WIP] Trying to fix beta and canary

rwwagner90:
Our builds are all failing now, due to ember-beta and ember-canary changes. I'm working to resolve some of the issues here, but if anyone knows of specific things I should be on the lookout for and definitely change, let me know!

cc @rwjblue @teddyzeenny

### DIFF
--- a/app/components/drag-handle.js
+++ b/app/components/drag-handle.js
@@ -52,14 +52,14 @@ export default Component.extend({
   startDragging() {
     let namespace = `drag-${this.elementId}`;
 
-    this.sendAction('action', true);
+    this.setIsDragging(true);
 
     document.body.addEventListener(`mousemove.${namespace}`, this._mouseMoveHandler);
     document.body.addEventListener(`mouseup.${namespace} mouseleave.${namespace}`, this.stopDragging);
   },
 
   stopDragging() {
-    this.sendAction('action', false);
+    this.setIsDragging(false);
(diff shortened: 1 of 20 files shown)


rwjblue:
See https://github.com/emberjs/ember.js/pull/16478 for details. I believe that PR was closed and reopened as https://github.com/emberjs/ember.js/pull/16499#pullrequestreview-112383911 (which was just merged).

Can you double check if that was the only change required? I’ll get it cherry picked into beta before the next one gets cut...

rwwagner90:
@rwjblue I tried latest canary and I see another issue:

```
not ok 14 Chrome 65.0 - Ember Debug - Object Inspector: Property grouping can be customized using _debugInfo
    ---
        actual: >
            null
        stack: >
            TypeError: Cannot add property __NAME_KEY__ember15238817701531233187778522__, object is not extensible
```

There are also a couple more failures, related to things like `this.get(...).on is not a function`, but I think maybe those are just things we need to update?

rwjblue:
That failure is the thing reported in https://github.com/emberjs/ember.js/issues/16479, I'm working on a PR to fix it now...

rwwagner90:
@rwjblue canary is passing now, so we should be green, or mostly green when a new beta release is cut. Thanks so much for everything you do! 😄

PR #789, opened 2018-04-19
Sticky Mixin Headers

nummi:
See animated GIF below (ignore the colors).

sticky

Also made an adjustment to an SVG that was throwing an svg-jar warning in the command line.

### DIFF
--- a/app/styles/mixin.scss
+++ b/app/styles/mixin.scss
@@ -9,7 +9,9 @@
 $mixin-left-padding: 22px;
 
 .mixin__name {
-  position: relative;
+  position: -webkit-sticky;
+  position: sticky;
+  top: 0;
   padding: 4px 2px 4px $mixin-left-padding;
   border-bottom: 1px solid #cecece;
   background-color: #f5f5f5;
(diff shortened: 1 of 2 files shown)


app/styles/mixin.scss
```diff
@@ -9,7 +9,9 @@
 $mixin-left-padding: 22px;
 
 .mixin__name {
-  position: relative;
+  position: -webkit-sticky;
```
rwwagner90: Rather than trying to get all the prefixes in our styles, should we just use https://github.com/kimroen/ember-cli-autoprefixer?

nummi: Your call. We use `-webkit` twice in the CSS (including this PR).

rwwagner90: I suppose it's no big deal then.

nummi:
> This looks good to me! Is it ready to merge? Will it conflict with any of the other work in the other PRs?

Ready to merge — it shouldn't cause any issues with the other PRs.

rwwagner90:
Merged, thanks! 🎉

PR #790, opened 2018-04-25
Use `qunit-dom` for DOM assertions

Turbo87:
see https://github.com/simplabs/qunit-dom


rwjblue:
Yes, it should be part of the default ember-cli@3.2 blueprint..

Issue #129, opened 2014-03-12
Feature request: route search 


okolehao:
Nearly every page would be nice to have a generic text search or filter option.   As the amount of 
data collects, it would be nice. 

![image](https://f.cloud.github.com/assets/3718695/2401445/24153276-aa16-11e3-99ef-64ca9c33a2e3.png)

lauraannwilliams:
Just adding my +1 for this. Finding the route/components used is a little difficult when there are a lot of items in the list.

stefanpenner:
Ya this is a great idea, just a reminder to folks. Ember-inspector is itself an ember-app, if you can write and ember-app you may actually have fun adding this to the ember-inspector.

nummi:
@rwwagner90 close?

rwwagner90:
@nummi we do not have search in the routes tab. I'll see if I can add it real quick

rwwagner90:
This is done now. Thanks @nummi!

PR #799, opened 2018-05-02
Search Routes

nummi:
https://github.com/emberjs/ember-inspector/issues/129

screen shot 2018-05-01 at 9 32 29 pm


lifeart:
@nummi @rwwagner90 I think it will be nice to have button 'goToRoute', near each route, to have ability move application into expected state. What do you think? Route may require some params, an we can ask youser provide it (id, slug, etc)

Issue #800, opened 2018-05-03
Property icon floating issue


lifeart:
![image](https://user-images.githubusercontent.com/1360552/39563595-84402d94-4eb9-11e8-8f99-0c1d7cf731a6.png)
uri: https://monegraph.com/m5692335728

Prop, named 'OWNER [id=__ember1525330949094135535295738]' broke Icon

lifeart:
@nummi looks like it's an css issue, do we need to add 'text-overflow'?

PR #801, opened 2018-05-03
Object Inspector Large Key Name Fix

nummi:
https://github.com/emberjs/ember-inspector/issues/800

croppercapture 1

Flexbox was allowing the property key to push other elements out of the way.


lifeart:
can we add "title=propName" to this container to get full property name on hover?

PR #805, opened 2018-05-21
Don't render excluded columns in tables.

### LINKED ISSUE #803: Data tab - change visible columns not working
In "Data" tab when I change visible columns, it looks like columns are added/removed, but data in the table are still the same. 

E.g. in the example below, when I hide ID column, it looks like just header is removed - data in the table are still showing "ID" column

ember-inspector

Bestra:
Fixes #803 

This PR attempts to fix issues we've been having with displaying the correct content in tables after users have filtered out some columns.
This PR threads the column schema down through the `x-list` into the `x-list-cell` component.  Consumers can then pass a `column` property to the `x-list-cell` that corresponds to one of the column ids defined in the schema.  Now when the columns change (when a user filters them out via the context menu) we don't render the `td` elements at all. 

This PR also fixes a problem on the data tab where the filters wouldn't work correctly after a user had selected a different model to view.  See f4f2283 for more details.

### DIFF
--- a/app/components/x-list-cell.js
+++ b/app/components/x-list-cell.js
@@ -18,6 +18,7 @@ import Component from '@ember/component';
 
 import { computed } from '@ember/object';
 import { htmlSafe } from '@ember/string';
+import { isEmpty } from '@ember/utils';
 export default Component.extend({
   /**
    * Defaults to a table cell. For headers
@@ -28,31 +29,7 @@ export default Component.extend({
    * @type {String}
    * @default 'td'
    */
-  tagName: 'td',
-
-  /**
-   * @property classNames
-   * @type {Array}
-   */
-  classNames: ['list__cell'],
-
-  /**
-   * `highlight` and `clickable` or class modifiers.
-   *
-   * @property classNameBindings
-   * @type {Array}
-   */
-  classNameBindings: ['highlight:list__cell_highlight', 'clickable:list__cell_clickable'],
-
-  /**
-   * Style passed through the `style` property
-   * should end up as the DOM element's style.
-   * Same applies to the `title` attribute.
-   *
-   * @property attributeBindings
-   * @type {Array}
-   */
-  attributeBindings: ['safeStyle:style', 'title'],
+  tagName: '',
 
   /**
    * Avoid unsafe style warning. This property does not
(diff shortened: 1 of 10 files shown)


Bestra:
@rwwagner90 @teddyzeenny I've made this a WIP pr temporarily to get feedback on the tests I've got.  I wrote a quick component test that runs fine in isolation but seems to run into a conflict of sorts with our test setup when I run it with the whole suite.

The suite eventually times out.  If I pause things beforehand, I normally see it spinning on 
https://github.com/emberjs/ember-inspector/blob/master/ember_debug/adapters/basic.js#L119

I'm going to try writing acceptance tests instead for the time being, but I figured it was worth bringing up to see if either you had seen something like this happen before.

app/components/x-list-cell.js
```diff
@@ -104,6 +81,32 @@ export default Component.extend({
    */
   highlight: false,
 
+  /**
+   * The list of possible columns
+   * @property columns
+   * @type {Array<{id: String, name: String, maxWidth: Number, width: Number}>}
+   */
+  columns: computed(() => []),
```
rwwagner90: Can we just set `columns` in `init`? I tend to prefer to set properties on `init` rather than having computeds that are just intended to get around having an array as a default prop.

app/components/x-list.js
```diff
@@ -154,7 +158,7 @@ export default Component.extend({
    * @method setupContextMenu
    */
   setupContextMenu() {
-    let menu = this.resizableColumns.getColumnVisibility().reduce((arr, { id, name, visible }) => {
+    this.menuContent = this.resizableColumns.getColumnVisibility().reduce((arr, { id, name, visible }) => {
```
rwwagner90: Shouldn't we just have `getColumnVisibility` figure out when to show/hide columns, rather than passing a column name to every list-cell?

Bestra: @rwwagner90 I think the problem we'd run into is that we don't use a data-driven approach to render the cell contents in most cases, except for in the `record` view.  https://github.com/emberjs/ember-inspector/pull/805/files#diff-1f81a1f049b22867a84780b0a7f9885eR21 is pretty much how we do it -- we manually wrap all the columns we want to render with an `x-list-cell`.

No matter what as the responsibility of showing which columns to show, we need some way of feeding that information down to our individual templates (at least with how they're implemented now)
We can't fall back to just doing things positionally either.  I'm afraid one way or another we need a way to say "this td belongs to this column" in the templates.

rwwagner90: Sounds logical. I will defer the decision to @teddyzeenny as he likely knows a lot more about it than me 😄

rwwagner90:
Was showing/hiding cells always bugged? I'm having a hard time figuring out how this ever worked. We know the visible columns, so I would generally associate the data with the column in a nested data structure or something, rather than manually passing keys to show and hide.

PR #811, opened 2018-05-29
Fix inspector on FF for Ember <= 2.7

teddyzeenny:
Fixes the same issue we fixed on master but for the old inspector.


rwwagner90:
@teddyzeenny did you mean to close this without merging?

teddyzeenny:
@rwwagner90 yes I realized this doesn't fix the issue I'm trying to solve.

PR #825, opened 2018-07-04
Adding security policy to see if this fixes #819

### LINKED ISSUE #819: Can't select the Ember inspector tab in Firefox
I am currently on firefox version 61.0b11 and I can't seem to select the Ember inspector tab. I have tried this with multiple pages that use Ember, but it seems to be a firefox issue. I don't get any error in the console when trying to go to the Ember inspector tab. I currently have version 3.1.3 of the inspector installed.

jrock2004:
Trying to see if setting CSP will help at all with the issue. This hopefully would fix #819

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -25,7 +25,17 @@ module.exports = function(environment) {
     APP: {
       // Here you can pass flags/options to your application instance
       // when it is created
-    }
+    },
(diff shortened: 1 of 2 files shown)


jrock2004:
This mad things worse

PR #826, opened 2018-07-04
Adding CSP to app to see if it fixes issue

### LINKED ISSUE #819: Can't select the Ember inspector tab in Firefox

jrock2004:
This is an attempt to fix #819

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -27,15 +27,14 @@ module.exports = function(environment) {
       // when it is created
     },
     contentSecurityPolicy: {
-      'default-src': ["'none'"],
+      'default-src': ["'self'"],
       'script-src': ["'self'"],
       'font-src': ["'self'"],
       'connect-src': ["'self'"],
       'img-src': ["'self'"],
       'style-src': ["'self'"],
       'media-src': ["'self'"]
-    },
-    contentSecurityPolicyMeta: true
+    }
   };
 
   if (environment === 'development') {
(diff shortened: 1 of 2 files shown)


rwwagner90:
@jrock2004 did this help? How can we test this?

cc @pbishop16

jrock2004:
I need to see why that test is failing. Again since when running locally this all works, I am not sure if this will fix it. But since its complaining about CSP I am thinking by adding the default it should work

jrock2004:
@rwwagner90 you can take my branch and compile it locally and try it yourself. If your asking if there are test cases we can write, 🤔

rwwagner90:
@jrock2004 I don't mean writing tests, I mean running in FF and seeing if it works. The locally running version has always worked, and we only hit the error from the store version, so trying to figure out how to see if this helps.

Any ideas @teddyzeenny @pbishop16?

pbishop16:
@rwwagner90 Agreed. Based on what I have experienced so far, to test we would have to upload the version to the add-on store.

rwwagner90:
It was a bug in FF, so closing this.

Issue #834, opened 2018-07-24
Ember.copy deprecation warnings


bgentry:
Using the ember-inspector on my v3.3 app spews an endless series of `Ember.copy` deprecation warnings:

```
15:00:44.656 telemetry.js:357 DEPRECATION: Use ember-copy addon instead of copy method and Copyable mixin. [deprecation id: ember-runtime.deprecate-copy-copyable] See https://emberjs.com/deprecations/v3.x/#toc_ember-runtime-deprecate-copy-copyable for more details.
        at logDeprecationStackTrace (http://localhost:4200/assets/vendor.js:17354:29)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:17456:17)
        at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:17378:17)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:17456:17)
        at invoke (http://localhost:4200/assets/vendor.js:17465:17)
        at deprecate (http://localhost:4200/assets/vendor.js:17437:34)
        at copy (http://localhost:4200/assets/vendor.js:52794:47)
        at Class.updateOrCreate (<anonymous>:1978:22)
        at Class.fulfill (<anonymous>:1992:26)
```

#### Screenshot

<img width="1347" alt="screen shot 2018-07-24 at 3 00 49 pm" src="https://user-images.githubusercontent.com/114033/43168525-7d21fdc4-8f52-11e8-9e62-7acd9b60ba9b.png">

Might be as simple as just changing to import for [this line](https://github.com/emberjs/ember-inspector/blob/1c34df9e8e7f521b380d09d6f7d3af15c650cc78/ember_debug/libs/promise-assembler.js#L98) to use ember-copy?

bgentry:
Hopefully #835 is a suitable fix! :pray:

PR #835, opened 2018-07-24
replace all uses of Ember.copy, Ember.merge w/ Object.assign

bgentry:
Attempted fix of #834. Not sure if I have to do something weird for the import in [`ember_debug/libs/promise-assembler.js`](https://github.com/emberjs/ember-inspector/pull/835/files#diff-c50ce0bdd9f93deddf67c4549ed160f6) or not…

### DIFF
--- a/ember_debug/libs/promise-assembler.js
+++ b/ember_debug/libs/promise-assembler.js
@@ -7,7 +7,7 @@
 
 import Promise from 'ember-debug/models/promise';
 const Ember = window.Ember;
-const { Object: EmberObject, Evented, A, computed, RSVP, copy, isNone } = Ember;
+const { Object: EmberObject, Evented, A, computed, RSVP, isNone } = Ember;
 
 let PromiseAssembler = EmberObject.extend(Evented, {
   // RSVP lib to debug
@@ -95,7 +95,7 @@ let PromiseAssembler = EmberObject.extend(Evented, {
     if (entry) {
       entry.setProperties(properties);
     } else {
-      properties = copy(properties);
+      properties = Object.assign({}, properties);
       properties.guid = guid;
       entry = this.createPromise(properties);
     }
--- a/app/libs/promise-assembler.js
+++ b/app/libs/promise-assembler.js
@@ -1,5 +1,4 @@
 import { assert } from '@ember/debug';
-import { copy } from '@ember/object/internals';
 import { later } from '@ember/runloop';
 import EmberObject, { computed } from '@ember/object';
 import EventedMixin from '@ember/object/evented';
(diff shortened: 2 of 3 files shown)


bgentry:
FYI, these same test failures appear to also happen on `master`. I'll open a separate issue for that.

ember_debug/libs/promise-assembler.js
```diff
@@ -6,8 +6,9 @@
  */
 
 import Promise from 'ember-debug/models/promise';
+import { copy } from 'ember-copy';
```
rwwagner90: @teddyzeenny should this import work here?

teddyzeenny: I don't think it would, no. Not unless the app being debugged has `ember-copy` installed as an addon.

rwwagner90: So what is your recommendation here?

teddyzeenny: There's no way to use Ember addons in `ember-debug` at the moment. We may be able to eventually work this out by custom-building and including the addons in the `ember-debug` final build, but I doubt it's going to be simple. 

For the sake of this PR, my recommendation would just be to not use `Ember.copy` in this file as it's not even doing a deep clone. So either `properties = {...properties}` if our Babel setup supports it, or a simple manual shallow copy by iterating over the properties. Either way, I don't think this function requires an Ember addon.

bgentry:
Would Obect.assign do the trick?

teddyzeenny:
@bgentry yes `Object.assign` works! 👍

rwwagner90:
@teddyzeenny with that in mind, does that work everywhere? Could we completely remove `ember-copy` in favor of that?

teddyzeenny:
`ember-copy` can do deep clones which `Object.assign` and `{ ...spread }` cannot, I think that's the difference? If they're all shallow copies we should be able remove `ember-copy`.

bgentry:
I updated this PR. It no longer uses ember-copy, instead just using `Object.assign` everywhere that `copy` was previously. I'm not certain whether this is ok (if everything is just a shallow copy) so somebody w/ more knowledge should verify that :smile:

rwwagner90:
@bgentry sorry to keep making you change things, but I think we should keep `ember-copy` in the `app` code, but just use `Object.assign` in the `ember_debug` code, just to be safe. With those changes, I would be fine with merging, since @teddyzeenny said `Object.assign` was okay for that one case.

rwwagner90:
As I was typing that, Teddy approved the PR 😂. Let me sync with him offline.

rwwagner90:
@teddyzeenny thinks this is good to go, and we don't need deep copies, so works for me 👍

PR #842, opened 2018-07-27
Remove jQuery, use flatpickr instead of pikaday

### LINKED ISSUE #590: Remove jQuery

rwwagner90:
ember-pikaday was using jQuery internally, so we had to move away from it. I chose flatpickr, but we could use anything else. I'm having trouble getting the tests to pass currently.

Resolves #590


rwwagner90:
@teddyzeenny the one test failure I am getting here is related to https://github.com/emberjs/ember-inspector/blob/master/tests/acceptance/object-inspector-test.js#L564.

```
not ok 37 Chrome 67.0 - Object Inspector: Date fields are editable
    ---
        actual: >
            view:getTree
        expected: >
            objectInspector:saveProperty
        stack: >
```

It seems `objectInspector:saveProperty` is not being called, although it does seem to save the new dates when I manually click through it. I was thinking maybe `fillIn` does not trigger the same things on flatpickr as it did on pikaday. Any ideas?

nummi:
Have you tried calling other events manually? I see a `keydown` — what about `keyup`?

rwwagner90:
@nummi I fixed that, everything works, I just need someone to review the PR.

Bestra:
I'll review

PR #846, opened 2018-07-29
Check for owner before calling isGlimmerTwo

### LINKED ISSUE #816: Seeing failing tests on `/tests` route with Ember Inspector open
When running the tests in an Ember app using the `/tests` route, I'm seeing tests fail as a result of Ember Inspector's `getOwner` function throwing an exception (specifically, [this line](https://github.com/emberjs/ember-inspector/blob/master/ember_debug/main.js#L164)).

After digging into the stack trace for a bit, it appears that one of Ember Inspector's initializers ([this one](https://github.com/emberjs/ember-inspector/blob/1aa8f03e1f00a6166042b61914a0ccdf03a4512b/ember_debug/vendor/startup-wrapper.js#L119-L138)) is falling over when the second test in the suite beings (so, presumably when `application.reset` is being called).

It almost seems like my application instance isn't even getting passed into the initializer, so that when Ember Inspector tries to call methods on it, it immediately fails.

Have you ever run into this before, or do you have any ideas about what might be causing this?

rwwagner90:
Resolves #816 

This stops the errors from happening, when there is no owner, because now we are checking if we have an owner before trying to use the owner.

I am not sure if this is the correct solution or if there is a way to actually get the real owner again here, rather than just doing nothing if it does not exist.

@teddyzeenny do we want to support running the inspector on `localhost:4200/tests` anyway?

### DIFF
--- a/ember_debug/view-debug.js
+++ b/ember_debug/view-debug.js
@@ -153,7 +153,8 @@ export default EmberObject.extend(PortMixin, {
     };
     window.addEventListener('resize', this.resizeHandler);
 
-    if (this.isGlimmerTwo()) {
+    // We do not have an `owner` when testing, so check for an owner here before calling isGlimmerTwo
+    if (this.getOwner() && this.isGlimmerTwo()) {
       this.glimmerTree = new GlimmerTree({
         owner: this.getOwner(),
         retainObject: this.retainObject.bind(this),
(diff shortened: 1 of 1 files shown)


Bestra:
@rwwagner90 I'd say what we've got in this PR is at least an improvement in the interim even we want to settle on different semantics longterm.  I'd assumed that we _always_ had an owner before this, so making us aware of the possible null (thanks for the code comments!) is a good thing.  I vote :shipit:  for now.

PR #855, opened 2018-08-03
Fix Ember Inspector during tests for Ember >= 3

### LINKED ISSUE #816: Seeing failing tests on `/tests` route with Ember Inspector open

### LINKED ISSUE #818: Failing in Ember 3.1
I'm seeing the following error in new greenfield Ember application using Ember Inspector 3.1.1:

screen shot 2018-06-04 at 11 17 31 am

Which is found: https://github.com/emberjs/ember-inspector/blob/02351d34b3b2a645e4ad9ced50606e3fb7486592/ember_debug/view-debug.js#L334-L336

Clearly:

screen shot 2018-06-04 at 11 25 43 am

is the problem.

I'm more than willing to help with a solution.

teddyzeenny:
The primary difference between acceptance tests for Ember < 3 vs Ember >= 3 is that for Ember < 3, every test created a new application + one application instance, whereas in Ember > 3 all the tests use the same application, but a different application instance for each test. The inspector was relying on the application boot, (which only happens once during tests in Ember >=3, and it happens even before the first test starts which was causing the inspector to fail and throw errors as it was starting even before the first application instance was created).

On the plus side, making the inspector only boot on application instance boot (instead of on application boot) also works for prior versions, and feels more "correct", so this should resolve everything everywhere.

Resolves #816 
Resolves #818 
Closes #846

### DIFF
--- a/ember_debug/general-debug.js
+++ b/ember_debug/general-debug.js
@@ -66,11 +66,6 @@ export default EmberObject.extend(PortMixin, {
   port: readOnly('namespace.port'),
 
   /**
-   * @type {Application}
-   */
-  application: readOnly('namespace.owner.application'),
-
-  /**
    * Sends a reply back indicating if the app has been booted.
    *
    * `__inspector__booted` is a property set on the application instance
@@ -79,7 +74,7 @@ export default EmberObject.extend(PortMixin, {
    */
   sendBooted() {
     this.sendMessage('applicationBooted', {
-      booted: this.get('application.__inspector__booted')
+      booted: this.get('namespace.owner.__inspector__booted')
     });
   },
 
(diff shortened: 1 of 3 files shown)


dandehavilland:
Any chance of a patch-level release to get this in?

Issue #821, opened 2018-06-09
What's New


nummi:
A quick mockup for what we discussed in the meeting. The design and content can be anything — I'm interested in thoughts on tucking _What's New_ under the Info tab.

@rwwagner90, @pbishop16 

![info](https://user-images.githubusercontent.com/3692/41193905-2ff9702a-6bc8-11e8-90b6-2fb0f1431186.jpg)

![info-2](https://user-images.githubusercontent.com/3692/41193906-32b33968-6bc8-11e8-9fe3-4bb76f9abdd3.jpg)

rwwagner90:
This seems like a good first step. I would like to default to showing the release notes whenever you grab a new version probably. We could just route people to that tab to handle that.

pbishop16:
@nummi Sorry for the late replay, but this looks great.

PR #858, opened 2018-08-10
What’s New screen

nummi:
Resolves https://github.com/emberjs/ember-inspector/issues/821

I added a new route and made it the default, `launch`. This route pulls the last loaded version number of Inspector out of local storage and compares it to the current version of Inspector. If the current version is more recent, jump to the What's New screen.

- [x] Load and parse CHANGELOG
- [x] Find _first load_ solution for Chrome (local storage only supported in FF)

Thank for @leongersing for mad regex.

screen shot 2018-08-10 at 8 33 25 pm


app/components/whats-new-content.js
```diff
@@ -0,0 +1,50 @@
...
+export default Component.extend({
+  error: false,
+  safeHTML: '',
+
+  didInsertElement() {
+    this.element.addEventListener('click', this.handleClick);
+    this.fetchMarkdown.perform();
+  },
+
+  handleClick(e) {
```
pbishop16: Would it be viable to use the native component `click` event handler here?

```
didInsertElement() {
  this.fetchMarkdown.perform();
},

click(e) {
  if (e.target.tagName.toLowerCase() === 'a') {
    e.preventDefault();
    window.open(e.target.href);
  }
},
```

nummi:
> @nummi could we use the chrome storage api to store the info we need?

That seems like a good solution. Should I write a wrapper around local?
https://github.com/emberjs/ember-inspector/blob/master/app/services/storage/local.js

nummi:
I'm starting to wonder if Chrome support is going to be more work than it's worth. The Chrome extension storage service is async so I can't write a simple wrapper that can match the synchronous style of the localstorage and memory storage services.

in `launch.js` I was using the `redirect` method of the Route to decide where to go on app launch:

``` javascript
Route.extend({
  redirect() {
    if(newVersion > storage.get('last-version')) {
      transitionTo('whats-new')
    }
  }
})
```

Would it be too hacky to take advantage of the async handling of the `model` method?

``` javascript
Route.extend({
  model() {
    return new Promise((resolve) => {
      if(Chrome) {
        chromeStorage.get('last-version').then(() => {
          // do check
          resolve();
        })
      } else {
        // localstorage check
        resolve();
      }
    });
  }
})
```

rwwagner90:
@nummi I think perhaps `beforeModel` would be a good place to do these checks and do the appropriate `transitionTo` calls.

nummi:
Okay. Changes made.

Fun fact I learned _after_ finishing the work: Firefox (45 and later) and Edge (14 and later) support  the chrome.storage API. Also the Firefox guides recommend against using localstorage for extensions.

rwwagner90:
This looks mostly good to me. Any chance we could get some basic test coverage in place for this?

PR #859, opened 2018-08-23
Fix issue with Proxy services

### LINKED ISSUE #814: Ember Inspector has errored with ember-feature-flags
Hi,

Not sure if this is the right place, but thought of reporting this

Ember Inspector: `3.1.1`
Ember: `3.1.2`
Ember CLI: `3.1.4`
Ember Feature Flags: `5.0.0`

In our app, we inject the `features` service in routes, components, etc using an initializer. Trying to click on any component, route, etc triggers the following in console and fails to open the Inspector tab
screen shot 2018-05-31 at 2 12 00 pm

**NOTE:**

- if we remove the initializer and inject the `features` service manually the tab works when clicking on components/routes... but still in Ember Inspector clicking on Container in the sidebar then on the features service same thing still happens

vladucu:
fix #814

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -475,7 +475,10 @@ function addProperties(properties, hash) {
     let options = { isMandatorySetter: isMandatorySetter(hash, prop) };
 
     if (typeof hash[prop] === 'object' && hash[prop] !== null) {
-      options.isService = hash[prop].type === 'service';
+      options.isService = !('type' in hash[prop]) && typeof hash[prop].unknownProperty === 'function' ?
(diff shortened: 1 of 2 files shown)


rwwagner90:
Thanks for the PR @vladucu! Does this apply only to services or should we also apply this fix to other entities?

vladucu:
sorry for taking so long to get to this @rwwagner90 

tbh, didn't looked too much, but can do....any suggestions what else to look for?

rwwagner90:
@vladucu I'm not sure, I was just curious if there were other proxies that may need this treatment. We'll assume there are not for now.

Issue #832, opened 2018-07-20
current route is not displayed if `resetNamespace:true`


luxferresum:
If the current route has set `resetNamespace: true` then it is not listed as current route in `ember-inspector` when it should be.

pbishop16:
@rwwagner90 I will take this one on.

PR #863, opened 2018-08-29
Fix routes reset namespace display

pbishop16:
Resolves: #832 
reset-namespace-demo-2

### DIFF
--- a/tests/acceptance/route-tree-test.js
+++ b/tests/acceptance/route-tree-test.js
@@ -35,6 +35,15 @@ module('Route Tree Tab', function(hooks) {
     port = this.owner.lookup('port:main');
   });
 
+  hooks.afterEach(async () => {
+    const checkbox = find('.js-filter-hide-routes input');
+
+    checkbox.checked = false;
+    await triggerEvent(checkbox, 'change');
+
+    port = null;
+  });
+
   function routeValue(name, props) {
     let value = {
       name,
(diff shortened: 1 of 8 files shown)


tests/acceptance/route-tree-test.js
```diff
@@ -129,13 +143,16 @@ module('Route Tree Tab', function(hooks) {
       "PostRoute",
       "application",
       "application",
+      "comments",
```
rwwagner90: Do we need `"comments"` twice here?

pbishop16: @rwwagner90 I double checked and the test seems to want "comments" listed twice similar to "application" and "post".

tests/unit/check-current-route-test.js
```diff
@@ -3,14 +3,152 @@ import { module, test } from 'qunit';
...
+        currentRoute: {
+          url: '/posts/show/comments',
+          name: 'posts.show.comments',
+        },
+        routeValue: {
+          url: '/comments',
+          name: 'comments',
+        },
+        message: 'resource match fails when current route does not match the root route',
+      },
```
pbishop16: @rwwagner90 You are very right. I did run into this edge case. The above test two cases roughly model this scenario.

**Edge case Failure**
example-fail

Adding the following URL check `const hasMatchingURL = routeURL === currentRouteURL;` resolved this edge case.

**Edge Case Success**
example-success

Please let me know if I failed to address your concern.

pbishop16:
@rwwagner90 You are very right. I did run into this edge case. The

PR #865, opened 2018-09-01
Add test for hideEmptyModelTypes

nlfurniss:
I couldn't find a test for the hiding empty models toggle, so I added one.


rwwagner90:
Awesome, thanks @nlfurniss!

PR #866, opened 2018-09-01
Add ability to sort models by record count

nlfurniss:
It would be nice to be able to sort models by their record count as well as alphabetically (current option).

### This PR
- Adds ability to sort models by record count

- Adds test for new sort functionality

- DRYs up the code to get, set, and remove  model sorting keys from Storage

### Feature in action
screen shot 2018-09-01 at 12 13 05 pm

### DIFF
--- a/app/controllers/model-types.js
+++ b/app/controllers/model-types.js
@@ -3,27 +3,40 @@ import { get, computed } from '@ember/object';
 import LocalStorageService from 'ember-inspector/services/storage/local';
 import { sort } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
+import { HIDE_EMPTY_MODELS_KEY, ORDER_MODELS_BY_COUNT_KEY } from 'ember-inspector/utils/local-storage-keys';
 
 export default Controller.extend({
   application: controller(),
   navWidth: 180,
-  sortProperties: ['name'],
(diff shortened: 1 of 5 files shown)


app/controllers/model-types.js
```diff
@@ -3,27 +3,36 @@ import { get, computed } from '@ember/object';
 import LocalStorageService from 'ember-inspector/services/storage/local';
 import { sort } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
+import { HIDE_EMPTY_MODELS_KEY, ORDER_MODELS_BY_COUNT_KEY } from 'ember-inspector/utils/local-storage-keys';
 
 export default Controller.extend({
   application: controller(),
   navWidth: 180,
-  sortProperties: ['name'],
+  sortByNameProp: Object.freeze(['name']),
```
rwwagner90: I'm curious what this `Object.freeze` is for?

nlfurniss: Sure I can make those changes. There prob should be some standardize, see [Ember.sort](https://www.emberjs.com/api/ember/release/functions/@ember%2Fobject%2Fcomputed/sort) docs using Object.Freeze

nlfurniss: Updated

PR #869, opened 2018-09-12
Expose storage type to use via LocalStorage service

nlfurniss:
The ternary to determine the type of the storage to use is repeated several times, this cleans that up.


rwwagner90:
Thanks!

PR #872, opened 2018-09-30
Attempt to support dynamically added iframes

rwwagner90:
This does seem to at least notice when an iframe is added and attempt to inject the scripts. I found that this does not work with ember-twiddle because they use a sandboxed iframe, and Chrome will not let us inject a script into the iframe unless we have both the frameId and permissions to its origin.

It ends up having an origin of `"null"`, which we don't have permissions to, so I've been unable to get much further here.

If anyone has some better ideas or knows how we could make this work, please let me know!

cc @thorsteinsson

### DIFF
--- a/app/adapters/web-extension.js
+++ b/app/adapters/web-extension.js
@@ -1,7 +1,7 @@
 /* globals chrome */
 import { computed } from '@ember/object';
 
-import BasicAdapter from "./basic";
+import BasicAdapter from './basic';
 import config from 'ember-inspector/config/environment';
 
 let emberDebug = null;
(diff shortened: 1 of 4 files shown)


rwwagner90:
@teddyzeenny @thorsteinsson  any ideas on how to get this over the finish line? I believe this is called when new iframes are added, however it doesn't seem to add those iframes to the dropdown. Perhaps one of you can see where I might have messed up?

We can worry about the sandboxed iframe permissions as a separate follow up issue after getting this to work.

rwwagner90:
Closing this PR, but keeping the branch around for reference, for now.

PR #873, opened 2018-10-03
Use ember-table for the Ember Data inspector

### LINKED ISSUE #562: Alphabetically sort data view
It would be nice to have the ability to sort alphabetically on any of the table columns within the data view, and also on the `attributes`, `belongs to`, `has many`, etc. lists when viewing a specific object. Especially with lots of data and attributes, it gets hard to find a specific item. Thanks!

thorsteinsson:
This enables horizontal scrolling, reordering of columns and sorting. It no longer supports right click to hide/show columns.

I would like to land this for only the Data Tab to begin with and then start using it in other places. I've styled it close to what we have today but I'd like to spend more time on improving the UI.

image

Three unrelated minor things I fixed while working on this
- Removed ember-async-image that was not being used and causing a build warning
- Removed CSS for dealing with bugs in Phantom (we're using Chrome now)
- Added ember-test-selectors to avoid using class names with `js-` prefix

/cc @Bestra 

Resolves #562


nummi:
Looks great! How's performance relative to `vertical-collection`?

thorsteinsson:
@rwwagner90 the latest version of ember-table fixed the bug I was seeing.

app/controllers/records.js
```diff
@@ -33,48 +31,33 @@ export default Controller.extend({
...
    * @property schema
    * @type {Object}
    */
-  schema: computed('columns', function() {
-    let columns = this.get('columns').map(({ desc }, index) => ({
-      id: dasherize(desc),
-      name: desc,
-      visible: index < this.get('columnLimit')
+  columns: computed('modelType.columns', function() {
+    return this.get('modelType.columns').map(({ desc }) => ({
```
rwwagner90:
```suggestion
    return this.get('modelType.columns').map(({ desc, name }) => ({
```

app/controllers/records.js
```diff
@@ -33,48 +31,33 @@ export default Controller.extend({
...
+      valuePath: `columnValues.${dasherize(desc)}`,
```
rwwagner90:
```suggestion
      valuePath: `columnValues.${name}`,
```

PR #874, opened 2018-10-09
Use [class='ember-view'] selector to exclude anything with extra classes

### LINKED ISSUE #419: Only application present in View Tree
I have been working on upgrading to 1.13 and have noticed that my view tree is quite off. I tried cloning and building from source but the same issue

screenshot from 2015-07-08 13 09 43

rwwagner90:
This should exclude liquid-fire elements or various others that add classes to the application container ember-view.

If you were adding your own classes to the application container, this would break, but I think this supports the majority of cases, and application wrappers are disabled by a lot of people these days anyway.

Resolves #419

@emberjs/inspector what do you guys think of this? It would hide liquid-fire etc from the view tree, but I think that is okay?

### DIFF
--- a/ember_debug/view-debug.js
+++ b/ember_debug/view-debug.js
@@ -334,7 +334,7 @@ export default EmberObject.extend(PortMixin, {
     if (!emberApp) {
       return false;
     }
-    let applicationView = document.querySelector(`${emberApp.rootElement} > .ember-view`);
+    let applicationView = document.querySelector(`${emberApp.rootElement} > [class='ember-view']`);
     let applicationViewId = applicationView ? applicationView.id : undefined;
     let rootView = this.get('viewRegistry')[applicationViewId];
     // In case of App.reset view is destroyed


ember_debug/view-debug.js
```diff
@@ -334,7 +334,7 @@ export default EmberObject.extend(PortMixin, {
     if (!emberApp) {
       return false;
     }
-    let applicationView = document.querySelector(`${emberApp.rootElement} > .ember-view`);
+    let applicationView = document.querySelector(`${emberApp.rootElement} > [class='ember-view']`);
```
rwjblue: Hmm, I think this is still somewhat broken. Specifically when the `application-template-wrapper` optional feature is enabled, there will be no root wrapping div at all.

https://github.com/emberjs/rfcs/pull/280 for references

rwjblue: Hmm, wait, is this only supporting legacy cases? Seems like everything 2.10+ (including anyone witht he optional feature) would be hitting the this.glimmerTree branch, right?

rwwagner90: @rwjblue you are correct, anything glimmer 2 and up should be hitting that `if`. I happen to be working in an app in a version afflicted by this issue, due to liquid-fire. I think for these legacy cases, this should be sufficient. We'll probably want to focus more on the no application wrapper format going forward anyway.

rwwagner90: We could also follow up with another PR to the 0.0.0-2.7.0 branch to apply this same fix there https://github.com/emberjs/ember-inspector/blob/ember-0.0.0-2.7.0/ember_debug/view-debug.js#L255

rwwagner90: I think the only case that may throw things off here, is if you were to be manually applying extra classes to the application wrapper div, but I think I am okay with that? I'm also okay with closing this PR and closing #419 as wontfix 😄

rwjblue: > manually applying extra classes to the application wrapper div, but I think I am okay with that?

Ya, that seems very suspect and not something I'm terribly worried about supporting by default.

rwwagner90: @rwjblue are we good to go here then? I will open a PR for the 0.0.0-2.7.0 branch later as well.

rwjblue: yep, seems good to me! sorry for my confusion before...

rwwagner90: @rwjblue no worries! Mind approving real quick then?

PR #875, opened 2018-10-09
Use [class='ember-view'] to avoid liquid-fire etc

rwwagner90:
Brings #874 to legacy code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: node_js
 
 node_js:
-  - "0.12"
+  - "4"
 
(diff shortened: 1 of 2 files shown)


rwwagner90:
@teddyzeenny I have not tried to update anything in the 0.0.0-2.7.0 branch before. Any ideas why the tests are failing here?

rwwagner90:
@teddyzeenny it looks like we have some errors with strict mode, since there is a `let` in there. Perhaps we should bump up the node version in travis?

PR #883, opened 2018-11-06
Fix test failures by passing function references to `off`

rwwagner90:
We need to allow beta and canary to fail, until we can find a way to run all Inspector tests as part of Ember PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
 jobs:
   fail_fast: true
   allow_failures:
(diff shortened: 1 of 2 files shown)


rwjblue:
Please create an issue in Ember for this failure, and mention it in a comment here so we know when we can remove.

rwwagner90:
@rwjblue we already don't need to allow the failures, got a quick turnaround on a fix from @pzuraq, so we're good to go. Updating this PR with the changes to fix things, if someone could add a quick approval 😃

PR #885, opened 2018-11-10
Collapse deprecations by default

### LINKED ISSUE #454: Collapse deprecations by default
It's kind of annoying that deprecations are expanded when opening the tab. Makes it tough to just zip through the list to get a sense of them all, and there is already a badge for the number of instances of a particular deprecation. 

Along with that, there is a weird viewing issue where if the instances list is longer than the window, it (somewhat) hides the list, making it look like you are at the end of the deprecation list. Probably worthy of a separate issue...

thorsteinsson:
Collapse deprecations by default, fixes #454.

I think it's better to get an overview of all the deprecations in the app and be able to expand each one.


rwwagner90:
@thorsteinsson it's not the promise tests, it's deprecation tab tests that are failing here. They expect things to be expanded and are looking for elements that are not there. We need to tweak the tests to toggle expanding first.

thorsteinsson:
Good point, I fixed the tests.

PR #888, opened 2018-12-01
Apply classes to get back record colors

rwwagner90:
With the switch to ember-table, we lost the text colors for new, modified, etc records. This adds them back.


app/styles/ember-table.scss
```diff
@@ -43,10 +43,22 @@
     border-right: 1px solid var(--base03);
   }
 
+  .blue {
+    color: #174fff
```
pbishop16: Minor: `;`

rwwagner90: Will fix in a follow up, and add stylelint to catch these things 👍

PR #889, opened 2018-12-02
Enforce stylelint

rwwagner90:
I'm using styleline-config-ship-shape here, but open to other configs, if we don't want this one. One of the main changes to note is things are alphabetized.

### DIFF
--- a/app/styles/component_tree.scss
+++ b/app/styles/component_tree.scss
@@ -1,13 +1,12 @@
 .component-tree-item {
-  position: relative;
-  display: flex;
-  min-height: 18px;
   align-items: center;
-  margin: 3px 3px 0 3px;
   border-radius: 4px;
-  font-size: 12px;
   color: var(--base12);
-
+  display: flex;
+  font-size: 12px;
+  margin: 3px 3px 0;
+  min-height: 18px;
+  position: relative;
 }
 
 .component-tree-item--component {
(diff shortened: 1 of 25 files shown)


app/styles/component_tree.scss
```diff
@@ -16,50 +15,58 @@
...
+}
+
+.component-tree-item__bracket:before {
   color: var(--base07);
+  content: '{{';
 }
-.component-tree-item__bracket::after {
-  content: '}}';
+
+.component-tree-item__bracket:after {
```
thorsteinsson: Minor: psuedo elements in CSS should use `::`. It's in more places, used for `before` and `after`.

app/styles/pill.scss
```diff
@@ -4,17 +4,17 @@
...
 
 .pill {
-  font-size: 12;
-  position: relative;
+  background: transparent;
+  background: var(--base06);
+  border-radius: 8px;
   cursor: default;
   display: inline-block;
+  font-size: 12;
```
pbishop16: *(Present before your edits)* Missing `px`

app/styles/component_tree.scss
rwwagner90: The stylelint rules have a rule to use a single `:`. I prefer the single, but not a big deal either way to me.

Issue #180, opened 2014-07-16
Multiple Ember apps on the same page


ramybenaroya:
My web app is instantiating 3  Ember apps (on the same page) and the inspector can inspect only the app that is created last. Is there any way to force the inspector to inspect a specific app by modifying some global variables on the page? Or in another way?
Thanks,
Ramy

hbrysiewicz:
+1 working in another code base with several apps - would be nice to have an option to pick which app in the inspector

meirish:
I'd love to help with this, but don't have any knowledge of the codebase. Would you be willing to give an outline of how you see it working / list of things that would need to be done and then I could take a pass at it?

rwwagner90:
Does anyone still need this and have the bandwidth to help implement it? I think that it might be easier to just run your apps separately, and inspect that way.

rwwagner90:
We feel that engines are the path forward for running multiple apps in one page, so we are going to focus our efforts there.

PR #898, opened 2018-12-18
[multiple apps] Support multiple apps on the same page and toggling between them

alexhancock:
This allows you to switch back and forth between two (or more) Ember apps running on the same page. I saw there is history of people wanting functionality like this (https://github.com/emberjs/ember-inspector/issues/180). We also have a use case for it at our company, so I implemented something.

I replace the `iframe-picker` component with a new `app-picker` component which I wrote. It presents a select with a list of all the Ember applications running on the page, and you can toggle between them.

I realize replacing `iframe-picker` outright may not be the desired approach. I'm open to other ideas, such as detecting when there are multiple apps on the page and only using `app-picker` in that situation, falling back to `iframe-picker` otherwise. If people are OK with replacing `iframe-picker`, I can delete that component entirely.

Happy to make any changes or clean up code as desired.

(Hat tip to @ginomiglio who helped in the early stages of this work)


rwwagner90:
@thorsteinsson @teddyzeenny what do you guys think of this? Is there a use case for iframes specifically over this functionality? I would think having an app switcher should work.

@alexhancock does this work for apps running in iframes I assume?

alexhancock:
@rwwagner90 I just tested and it does not work for apps in iframes - only apps that are rendered on the same page together.

It's not immediately clear to me why apps in iframes don't show up in the new picker, but if it's important to keep that functionality I am more than happy to look into/add it.

alexhancock:
@rwwagner90 I looked into this today. It's going to require some more investigation and better management of the `detectedApplications` list in `app/port.js`. It's possible to make it work, but I need more time to look at it than I have today. I leave soon for a couple weeks of holiday travel, but I can look again in the new year.

In the meantime, I will follow along with what others think about the approach in discussion on this PR.

Happy holidays everyone 🎄

nummi:
Could we merge the iframes and multiple apps together?

image


```
<select>
  <optgroup label="http://...">
    <option value="app.name">app.name</option>
    <option value="app.name">app.name</option>
  </optgroup>
  <optgroup label="http://...">
    <option value="app.name">app.name</option>
  </optgroup>
</select>
```

rwwagner90:
@nummi that sounds like a good idea 👍

alexhancock:
I pushed an update that gets apps running in iframes into the list with apps directly on the page.

Couple caveats...

1) I noticed an existing issue where if you don't have the Ember Inspector open when the page with the iframe in it loads, an app running in that iframe won't show up in the list of apps when you open the Ember Inspector. Apps running in iframes only show up if you load the page with the Ember Inspector already open. This is because the only events which add apps running in iframes to the list of `detectedApplications` in `app/port.js` are related to the app booting, which you obviously miss if the Ember Inspector isn't running at the time apps boot. We could find a fix for this, but it's somewhat separate from my changes.

2) I like the idea of showing the list of apps running in iframes under the url of the iframe (or the page) using optgroups (as mentioned here https://github.com/emberjs/ember-inspector/pull/898#issuecomment-449875783), but I don't know how I would do that given the data we currently have available. The data is currently just a list of application IDs without much information about how and where they are running (iframe vs in the page directly). If someone with more knowledge of the project would want to pair on this with me, I'd be happy and willing to do a Google hangout. Or I could try to look into making bigger changes to support it on my own.

alexhancock:
Thoughts on my above post? @rwwagner90 @nummi

rwwagner90:
@alexhancock I can take a look and see if I can find a way to get more meaningful info out of it. The iframes are supposed to have a listener added, so they should update as added to the page, I think.

ginomiglio:
Even without the optgroups, I think this is still strictly an improvement over the existing functionality. Can we move forward with this now and address the other issues separately?

rwwagner90:
@alexhancock @ginomiglio do either of you have some examples of pages with multiple apps and/or iframes I can test with?

alexhancock:
@rwwagner90 When I was testing for iframes I used two newly created apps and put an iframe to `app b` in `app a`'s template. I didn't save the code though.

For the two apps in one page case, our production website exhibits this behavior: https://squareup.com/dashboard/. You need to sign up for an account and log in to see it, but that will probably be less work than setting up a second mock testing setup.

If it's a requirement to have a really clean setup for testing, I can create a mini project with three apps. Let me know!

nummi:
For testing iframes I just throw emberjs.com/depreciations into the source of an iframe tag.

rwwagner90:
Thanks for testing that out @nummi! Just updated the branch. Once the tests finish, assuming they pass, we'll :shipit:

alexhancock:
Thanks for closing the loop on this. I will keep an eye out for when it lands in a release.

Have a good weekend everyone!

rwwagner90:
@alexhancock just released 3.5.0, which should include the changes. It may take awhile for it to make it through the Chrome store, but should be out shortly 😃

ember_debug/vendor/startup-wrapper.js
```diff
@@ -75,15 +72,25 @@ var EMBER_VERSIONS_SUPPORTED = {{EMBER_VERSIONS_SUPPORTED}};
...
+
+          if (!Ember.EmberInspectorDebugger._application) {
+            bootEmberInspector(instance);
+          }
         }
       });
     }
   });
 
+  function bootEmberInspector(appInstance) {
```
rwwagner90: @alexhancock I am concerned moving this boot stuff may have broken some apps. I need to investigate further, but inspecting https://shipshape.io no longer works with inspector 3.5.0.

alexhancock: Shoot! I'm pretty busy today but I could look tomorrow or definitely Monday.

Any more context on why you think specifically it might be the moving of the boot logic? If so, it would be helpful in debugging.

rwwagner90: @alexhancock I haven't had a chance to look yet, but it worked on https://shipshape.io before I merged the last couple PRs, so thought this was a logical first place to look, since we moved the boot logic around, maybe it's happening at the wrong time for newer Ember versions or something.

alexhancock: I dug into this just now. The boot logic actually seems OK. It ends up having more to do with how I changed the `uniqueId` logic in `ember_debug/port.js`

https://github.com/emberjs/ember-inspector/blob/master/ember_debug/port.js#L17

```
this.get('namespace._application.name');
```

... is not set to anything in the case of https://shipshape.io. We could do something like...

```
  uniqueId: computed('namespace._application.name', function() {
    return this.get('namespace._application.name') || `app${RANDOM_NUMBER}`;
  }),
```

... but it feels like a better idea to think about what it really should be if `namespace._application.name` is only going to be set sometimes.

Any ideas?

nummi: I'm okay with `app${RANDOM_NUMBER}` for the moment. We can work out something better later.

rwjblue: https://github.com/emberjs/ember-inspector/pull/929 should fix...

PR #901, opened 2018-12-24
Update to 3.6

cibernox:


rwwagner90:
@cibernox just want to confirm the only updates were to packages?

rwwagner90:
@cibernox there are lots of deprecation warnings. It would be good if we could resolve those. I think the tests are failing because of linting.

PR #911, opened 2019-01-04
Ember 3.7 beta, bump other deps

rwwagner90:
ember-route-action-helper was causing deprecation warnings for handlerInfos, so I moved to using controller actions and passing the target instead.

Closes #901


nummi:
👍🏼👍🏼

PR #917, opened 2019-01-21
WIP: Button to view element in browser Elements panel

nummi:


app/controllers/component-tree.js
```diff
@@ -258,14 +258,19 @@ export default Controller.extend({
...
-    inspectElement({
-      objectId,
-      elementId
-    }) {
-      this.get('port').send('view:inspectElement', {
-        objectId,
-        elementId,
-      });
-    },
+    inspectElement(item) {
```
nummi: I grabbed this code from the view tree components — since we are only dealing with components is there ever a case where `objectId` _isn't_ available?

PR #918, opened 2019-01-23
x-list -> list

rwwagner90:
Also, converted some things to be angle bracket component syntax. This was just a proof of concept, to ensure things work before we do a larger refactor of the inspector.


app/templates/components/list.hbs
```diff
@@ -50,7 +49,7 @@ as |content|
       maxWidth=column.maxWidth
       on-drag=(action "didResize" column.id)
       position=(one-way column.width)
-      setIsDragging=setIsDragging
+      setIsDragging=@setIsDragging
```
pbishop16: It seems like some of the test failures point to `@setIsDragging`. Should it remain as `setIsDragging`?

rwwagner90: I think it's that this stuff is incompatible with older Ember versions, not sure how to best handle this

rwjblue: We need to add ember-named-arguments-polyfill I believe...

rwjblue: gah, looks like https://github.com/rwjblue/ember-named-arguments-polyfill only goes down to Ember 2.10 😩

rwwagner90: @rwjblue so adding that polyfill gets things passing on 2.10+. It seems like versions before that still have the error. Thoughts?

rwwagner90: Heh, I needed to refresh my page, you noticed the same things first 😛

rwwagner90:
@rwjblue what should we do for supporting angle bracket components here? Should we drop the tests for anything pre 2.10?

PR #925, opened 2019-01-27
alt/option click to toggle children in component tree

nummi:
We can currently expand/collapse ALL items but not in any granular fashion without a lot of clicking. This can be a pain to explore large component trees. With this PR I'm proposing we copy the MacOS style (I'm sure other platforms have something similar) of expanding/collapsing children nodes by alt/option clicking on a disclosure triangle.

image

Option click `{{es-header}}` and it will fold and so will all the children.


app/components/disclosure-triangle.js
```diff
@@ -0,0 +1,16 @@
+import Component from '@ember/component';
+
+export default Component.extend({
```
locks: if you are invoking this component with angle brackets, can't some of this be moved to the template instead?

rwwagner90: We cannot currently use angle bracket components in inspector. It will not build.

rwjblue:
Seems neat!

app/templates/components/component-tree-item.hbs
```diff
@@ -4,17 +4,12 @@
...
-      {{action toggleExpanded bubbles=false}}
-    >
-      {{svg-jar
-        "disclosure-triangle"
-        title="Toggle visibility"
-        width="9px"
-        height="9px"
-      }}
-    </button>
+    <DisclosureTriangle
```
locks: See https://github.com/emberjs/ember-inspector/pull/925/files#r251897592 !

rwwagner90: We need to avoid angle bracket components for now, unfortunately.

Issue #927, opened 2019-01-28
[v3.5] No longer detects my ember-app


urbany:
The inspector stopped detecting my app as an Ember App since last update (v3.5).
First time this has happened in years, any way I can debug this?
I have the `Deprecations were detected, see the Ember Inspector deprecations tab for more details.` message in the console, but the inspector tab says: `Ember application not detected!`

@rwjblue pointed me to this pull where there are some comments that might be related to this issue in the end: https://github.com/emberjs/ember-inspector/pull/898

Let me know if I can do anything to help debug this!

Edit: forgot to say I'm using latest stable releases of ember-cli and ember-source.

trenpixster:
Also having issues with version 3.5.0 of the inspector: it stopped detecting my ember app. I have no warnings whatsoever in Chrome. In Firefox it says `Ember Inspector Active`, but it doesn't detect the app.

Ember version: 2.18.2

urbany:
@efx I can send you a link in Discord.

```EmberENV: {
  EXTEND_PROTOTYPES: {
    Date: false,
    Function: false,
    String: false,
  },
},```

**Comment by efx** [comment]
Per some investigation and discussion, we can rule out #926 as being related. So next step is to bisect to find what part of the recent changes caused this break per Robert's guidance here: https://github.com/emberjs/ember-inspector/pull/898/files#r251179496

**Comment by gfmartinez** [comment]
Having the same issue. Looking at two ember apps Ember v3.5 and v3.6. Chrome and firefox: neither extensions work.

**Comment by guangda-prosperworks** [comment]
Same issue here with ember 2.18.

**Comment by aandis** [comment]
+1 on Ember 3.4

**Comment by rwwagner90** [comment]
We are aware of this issue and working on a fix.

**Comment by urbany** [comment]
@rwjblue and @rwwagner90 found a quick fix while this is not properly fixed. Just add a name property to to your app/app.js:

```js
// app/app.js
const App = Application.extend({
  name: 'anything-here', // add this property with any name
  ...
```

PR #929, opened 2019-01-29
Make Port#uniqueId a bit more resilient.

rwjblue:
Fixes app detection issues introduced in #898 by using a layered fallback system to control the uniqueId:

* Use `name` specified in `app/app.js` (the easiest for users to customize to control how their app is displayed in the inspector)
* Use `modulePrefix` specified in `app/app.js` (all ember-cli apps would have this, and is _generally_ still representative of what each app owner calls their app).
* Use `(unknown app - <RANDOM_GUID>)` as a last chance fallback to ensure **all** detected applications at least show up

Fixes https://github.com/emberjs/ember-inspector/issues/927

### DIFF
--- a/ember_debug/port.js
+++ b/ember_debug/port.js
@@ -1,5 +1,5 @@
 const Ember = window.Ember;
-const { Object: EmberObject, computed, run } = Ember;
+const { Object: EmberObject, computed, run, get } = Ember;
 const { oneWay } = computed;
 
 export default EmberObject.extend(Ember.Evented, {
@@ -13,8 +13,20 @@ export default EmberObject.extend(Ember.Evented, {
    * @property uniqueId
    * @type {String}
    */
-  uniqueId: computed('namespace._application.name', function() {
-    return this.get('namespace._application.name');
+  uniqueId: computed('namespace._application.{name,modulePrefix}', 'namespace.applicationId', function() {
+    let application = this.get('namespace._application');
+
+    // can be set by users to provide a "pretty" name for a given app when
+    // multiple applications are detected
+    let name = get(application, 'name');
+
+    // good default for "standard" ember-cli apps
+    let modulePrefix = get(application, 'modulePrefix');
+
+    // fallback to do "something" when neither name or modulePrefix are found
+    let fallback = `(unknown app - ${this.get('namespace.applicationId')})`;
+
+    return name || modulePrefix || fallback;
   }),
 
   init() {


ember_debug/port.js
```diff
@@ -13,8 +13,20 @@ export default EmberObject.extend(Ember.Evented, {
    * @property uniqueId
    * @type {String}
    */
-  uniqueId: computed('namespace._application.name', function() {
-    return this.get('namespace._application.name');
+  uniqueId: computed('namespace._application.{name,modulePrefix}', 'namespace.applicationId', function() {
```
rwwagner90: As we discussed briefly in Discord, I think this should be an `applicationName` computed, and we should revert to the old `uniqueId` generation. We then need to pass both the name and id around and look up the apps by the id somehow, rather than name, I think.

rwjblue: Sounds good to me @rwwagner90, I'll leave that to you (happy to review and otherwise help out as needed though)...

alexhancock: > As we discussed briefly in Discord, I think this should be an `applicationName` computed, and we should revert to the old `uniqueId` generation. We then need to pass both the name and id around and look up the apps by the id somehow, rather than name, I think.

This makes sense to me as well. I was going to leave a comment to this effect after I thought about the idea of doing something like ``app${randomNum}`` last night, but then realized that **probably** wouldn't work for app lookups anymore when switching between apps.

rwwagner90: @alexhancock I started trying to do this, but I've been so far unsuccessful. Maybe you could take a look and help get this over the finish line? https://github.com/emberjs/ember-inspector/pull/930

rwwagner90: @alexhancock are you on Discord? It would be good to sync up there.

alexhancock: @rwwagner90 I am not on discord, but I have an old account. Username: alexhancock.

I have some Q's for you about the changes in https://github.com/emberjs/ember-inspector/pull/930 

How do I connect with you all on Discord?

rwwagner90: @alexhancock the Ember community has a discord. I believe you can join via https://discordapp.com/invite/zT3asNS

alexhancock: Cool! I see it now, but I don't have permissions to send messages in the `dev-ember-inspector` channel

screen shot 2019-01-29 at 11 42 48 am

alexhancock: Heading into a meeting for an hour, but I really only have one question about the changes in #930 so I will just post it here! :P 

Why all the changes to the logic in `app/port.js` and why remove the `sendAppList` logic? 

As far as I can see that code being altered/removed is connected to why populating the list of apps and switching apps no longer works in that changeset. Are the changes born from a desire to clean things up or implement the original app list population logic in a different way?

Happy to help on a revised version of this change after my lunch meeting.

rwwagner90: @alexhancock the app list was using the `uniqueId` and the names as the same thing. We need a reworking of it to support sending both `applicationId` and `applicationName`, like the PR is doing. We already have a `for` loop that loops through the applications, so if we can find a way to use that to populate the apps, that would be ideal. Ping me on discord I am `rwwagner90` there as well. I see two `alexhancock` so I am not sure which is you. You have to go through the `#welcome` and `#setup-profile` channels to get permissions to everything.

alexhancock: Also, another option would be to revert my original change, release a new version and then re-introduce the changes once we get the issues resolved.

Not sure how close we feel we are on fixing the issues - I'll leave it up to you.

rwwagner90: @alexhancock let's discuss via discord, it will make things quicker

PR #935, opened 2019-01-31
Remove "info" nav title

nlfurniss:
This is a very low priority, but maybe I'm not the only one who find the `Info` heading hard to distinguish from the links underneath. It also doesn't add anything, as the user can already see that `Info` is highlighted in the left menu.

screen shot 2019-01-30 at 10 20 46 pm

If you disagree, feel free to close 😊

### DIFF
--- a/app/templates/info.hbs
+++ b/app/templates/info.hbs
@@ -5,9 +5,6 @@
     width=180
   }}
     <div class="split__panel__bd">
-      <div class="nav__title">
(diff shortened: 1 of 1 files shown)


rwwagner90:
Removing is fine with me. Thoughts @nummi?

nummi:
Fine with me. I do intend to bold the headings in the future but in this case the heading feels redundant.

image

image

PR #938, opened 2019-01-31
Simplify layout of Routes Tree

nummi:
(From discussion with @rwjblue and @rwwagner90)

## 🛠 What Changed
- Replaced route and controller class names with simple label
  - Users usually do not (and should not) customize the class names for routes and controllers.
- Removed templates column
  - No functionality lost other than displaying the template name. Same note as above with following convention.
- Combined routes and controllers column
- Used a monospaced font in the URL column. I feel like a list of urls like this is easier to scan when everything lines up as best as possible.

## ❓Question for Reviewers

Does the "Objects" table heading work for you?

## 🏞 Screenshots

<details>
    <summary>Original</summary>
    screen shot 2019-01-31 at 11 39 44 am
</details>

<details>
    <summary>New</summary>
    screen shot 2019-01-31 at 11 39 31 am
</details>


nummi:
@pbishop16, I requested you to review since you were not part of the original conversation. Thanks :)

PR #939, opened 2019-02-01
Default theme for bookmarklet and tests

nummi:
The bookmarklet was not loading a theme so I added one to `app/index.html`. If Inspector is loaded in a browser extension `app/adapters/web_extension.js` will override based on browser chrome settings.


nummi:
I just realized I can do the same for tests. Does anyone oppose?

skeletons/bookmarklet/load_inspector.js
```diff
@@ -53,7 +53,7 @@
     if (e.data.name === 'version-mismatch') {
       injectEmberDebug('panes-' + e.data.version.replace(/\./g, '-') + '/ember_debug.js');
     }
-  });  
+  });
```
nummi: Not sure what these characters are but they are causing some weirdness when deployed to S3.

image

PR #942, opened 2019-02-24
UI Components Add-On

nummi:
The goal of this PR is to move as much generic UI logic to another location and keep the main repo for _business logic_. This is a huge commit so I'll leave comments where I think they will be useful.

Step two could include: more tests and documentation, usage of a component playground like _Storybook_ or _Ember CLI Addon Docs_.

This will also make life easier if we go down the _Inspector Add-ons_ route.


nummi:
_explanation coming soon..._

rwwagner90:
This is awesome work! I'll give it a thorough review this week at some point. Pretty swamped with some other things currently.

app/styles/app.scss
```diff
@@ -25,22 +25,9 @@
...
-@import "nav";
-@import "notice";
 @import "object_inspector";
-@import "pill";
 @import "route_tree";
-@import "send_to_console";
-@import "sidebar_toggle";
-@import "split";
-@import "toolbar";
-@import "warning";
```
nummi: Most of these were moved to the add-on — a few were no longer being used.

app/styles/component_tree.scss
```diff
@@ -70,27 +70,6 @@
...
-  svg {
-    transition-duration: 0.1s;
-    transition-property: transform;
-  }
-
-  &.collapsed svg {
-    transform: rotate(-90deg);
-  }
-}
-
```
nummi: All of this was rolled into the `disclosure-triangle` component

app/templates/components/view-item.hbs
```diff
@@ -53,13 +53,13 @@
...
-      {{send-to-console action=sendObjectToConsole param=model.value.objectId}}
+      {{ui/send-to-console action=sendObjectToConsole param=model.value.objectId}}
     </div>
   {{else}}
     --
   {{/if}}
 {{/list.cell}}
 
 {{#list.cell class="list__cell_size_small list__cell_value_numeric"}}
-  <span class="pill pill--not-clickable pill--small js-view-duration">{{ms-to-time model.value.duration}}</span>
```
rwwagner90: No need for the `not-clickable` class?

app/templates/container-type-toolbar.hbs
```diff
@@ -1,18 +1,18 @@
...
 
   <div class="divider" aria-hidden="true"></div>
 
-  {{send-to-console action=(action "sendContainerToConsole")}}
+  {{ui/send-to-console action=(action "sendContainerToConsole")}}
 
   <div class="divider" aria-hidden="true"></div>
 
   <div
-    class="toolbar__search toolbar__search--small js-container-instance-search"
```
rwwagner90: No need for the `toolbar__search--small` class anymore?

lib/ui/addon/components/sidebar-toggle.js
```diff
@@ -1,21 +1,14 @@
...
 
+export default Component.extend({
+  layout,
   tagName: 'button',
-
-  side: 'right',
-
-  isExpanded: false,
-
-  isRight: equal('side', 'right'),
```
rwwagner90: No need for this right stuff anymore?

nummi: As far as I can tell we never had a sidebar toggle for the left sidebar. We can add this back if we ever need it.

app/templates/container-type-toolbar.hbs
nummi: It only made the search smaller by `25px`. ¯\_(ツ)_/¯

app/templates/components/view-item.hbs
nummi: It didn't seen to have any effect. The class name didn't really describe what it did anyway. Also, I just changed the base `.pill` class to have that "not-clickable" functionality which is just inheriting the parents cursor.

Issue #950, opened 2019-03-18
Remove usage of .readOnly()


gabz75:
Remove usage of `.readOnly()` as it is now deprecated.

PR #951, opened 2019-03-18
Remove usage of chainable functions .readOnly()

gabz75:
Fixes https://github.com/emberjs/ember-inspector/issues/950

### DIFF
--- a/ember_debug/adapters/web-extension.js
+++ b/ember_debug/adapters/web-extension.js
@@ -1,10 +1,16 @@
 import BasicAdapter from "./basic";
 const Ember = window.Ember;
-const { computed, run, typeOf } = Ember;
+const { run, typeOf } = Ember;
 const { isArray } = Array;
 const { keys } = Object;
 
 export default BasicAdapter.extend({
+  init() {
+    this._super(...arguments);
+    this.set('_channel', new MessageChannel());
+    this.set('_chromePort', this.get('_channel.port1'));
+  },
+
   connect() {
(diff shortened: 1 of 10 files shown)


ember_debug/adapters/web-extension.js
```diff
@@ -1,10 +1,16 @@
...
-const { computed, run, typeOf } = Ember;
+const { run, typeOf } = Ember;
 const { isArray } = Array;
 const { keys } = Object;
 
 export default BasicAdapter.extend({
+  init() {
+    this._super(...arguments);
+    this.set('_channel', new MessageChannel());
+    this.set('_chromePort', this.get('_channel.port1'));
```
gabz75: I doubled check that no code is setting these two variables and since we expect variables starting with `_` we should be good?

tests/ember_debug/object-inspector-test.js
```diff
@@ -419,20 +419,6 @@ module('Ember Debug - Object Inspector', function(hooks) {
     assert.equal(serializedComputedProperty.dependentKeys[1], "bar");
   });
 
-  test('Read Only Computed properties mush have a readOnly property', function(assert) {
```
gabz75: This test seems silly as it's just testing core ember stuff. Removing it.

ember_debug/adapters/web-extension.js
rwwagner90: I agree 👍

PR #954, opened 2019-04-02
Ensure we do not add quotes twice to strings when edited

### LINKED ISSUE #952: Editing an attributes adds quotes to the value
When editing an attribute in the Inspector object side panel, quotes are added to the value being edited. If this value is then submitted, the quotes are escaped:

1. Screenshot 2019-04-02 at 15 21 57
2. Screenshot 2019-04-02 at 15 22 01
3. Screenshot 2019-04-02 at 15 22 09

rwwagner90:
Fixes #952

### DIFF
--- a/tests/unit/utils/parse-text-test.js
+++ b/tests/unit/utils/parse-text-test.js
@@ -1,0 +1,13 @@
+import parseText from 'ember-inspector/utils/parse-text';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | parseText', function() {
+
+  // Replace this with your real tests.
+  test('parseText returns the correct output', function(assert) {
+    assert.equal(parseText('foo'), 'foo', 'does not add quotes for string');
+    assert.equal(parseText('"foo"'), 'foo', 'removes quotes when simple string is already wrapped in quotes');
+    assert.equal(parseText('1'), '1');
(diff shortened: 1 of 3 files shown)


rwwagner90:
@locks can you think of any reason we would want to be adding these quotes? It seems like the logic has been around a long time, and presumably worked before? Perhaps it is Ember version specific?

rwwagner90:
@rwjblue sure, we can do that. Which part would you like extracted out?

rwjblue:
I was thinking that we could extract the `if (type === 'type-string')` body into a separate function (ala `formatString(value)`) then we could unit test with various input values (with and without quotes, etc).

rwwagner90:
@rwjblue there was actually already a parseText function, so I used that and tweaked some things. Mind reviewing the PR?

tests/unit/utils/parse-text-test.js
```diff
@@ -0,0 +1,13 @@
+import parseText from 'ember-inspector/utils/parse-text';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | parseText', function() {
+
+  // Replace this with your real tests.
```
nummi: ❌?

nummi:
Should we handle single quotes as well?

rwwagner90:
> Should we handle single quotes as well?

I don't think we need to do anything with single quotes, as it wasn't handled before.

Issue #958, opened 2019-04-11
Remove grunt for compress tasks


rwwagner90:
We currently use grunt for some build tasks, but some of our packages are way out of date and not supported. We should modernize our build and use gulp, broccoli, rollup, etc.

KamiKillertO:
I've some experiences with gulp, I'll have a look 👍

KamiKillertO:
Ok so grunt compress files and publish on s3. It's not gonna be an easy thing to migrate from grunt to something else 😞. The compress tasks can be tested in local but for the s3 tasks I'm not sure how to test ><.

rwwagner90:
The S3 tasks have actually been not working anyway. I think we should use the built in Travis functionality to upload to S3 https://docs.travis-ci.com/user/deployment/s3/

PR #964, opened 2019-04-23
Replace grunt with gulp

KamiKillertO:
Related to #958.

I've decided to use [gulpjs](https://gulpjs.com).
This PR replace the `grunt compress:*` tasks with gulp tasks and the `grunt clean-tmp` task with a `gulp clean-tmp` one.


rwwagner90:
Thanks so much for the PR @KamiKillertO!

PR #966, opened 2019-04-23
Use isComputed when available

rwwagner90:
This uses `Ember.Debug.isComputed` for Ember 3.10+ to determine if values are computed. It also fixes some related test failures.


rwjblue:
Thank you for working through this @rwwagner90!

rwwagner90:
You're welcome! Thank you for all your help @rwjblue 😃

Issue #965, opened 2019-04-23
Use Travis for S3 Uploads


rwwagner90:
Travis support deploying to S3, so rather than using grunt for this, it would be ideal to deploy directly from Travis. https://docs.travis-ci.com/user/deployment/s3/

rwwagner90:
@teddyzeenny I need the S3 credentials. I am not sure where `config/secrets.json` is supposed to come from. Please ping me on Discord and we can get this figured out 😄

rwwagner90:
This works now 🎉

nummi:
🎉

PR #967, opened 2019-04-24
Use Travis for S3 Uploads

KamiKillertO:
Related to #965 

I've setup a basic travis deploy task using https://docs.travis-ci.com/user/deployment/s3. It's not complete yet, s3 secret information `access_key_id` and `secret_access_key` are missin and I can't set them because I don't have the information 😅 .


rwwagner90:
This looks good to me! I will get the keys and update this 👍

PR #969, opened 2019-04-25
[port] Change the way we set up the port property to be used by logic in the port-mixin

alexhancock:
This fixes a bug with the new application switcher functionality where `this.get('port')` was not set at times other code was trying to use it.

I removed the `port: readOnly('namespace.port')` in all the things that use the `port-mixin` and set the property in the init method of the mixin itself instead.


rwwagner90:
Thanks so much for debugging and fixing this @alexhancock! I really appreciate it! 😃

PR #984, opened 2019-06-06
Highlight children of selected component

nummi:
Looking for feedback.

When looking at the tree sometimes I find it would be helpful to know the children without carefully looking down the list and lining things up in my mind.

highlight


rwwagner90:
This seems cool to me!

app/controllers/component-tree.js
```diff
@@ -298,19 +301,19 @@ export default Controller.extend({
...
     },
 
-    inspectElement(item) {
-      let elementId;
-      let objectId = item.get('view.objectId');
+    /**
+     * View DOM node in browser Elements panel
+     */
+    viewInElementsPanel(item) {
+      const objectId = item.get('view.objectId');
```
rwwagner90: @nummi did you talk to @rwjblue about what the difference is between `objectId` and `elementId`? It would be good to figure that out, and if we only need one or the other, remove one.

rwjblue:
Totally!

rwwagner90:
@rwjblue did @nummi talk to you about the difference between `objectId` and `elementId` here? I know he had questions around it and we are not sure if we need both.

app/controllers/component-tree.js
```diff
@@ -298,19 +301,19 @@ export default Controller.extend({
     /**
      * Scrolls the main page to put the selected element into view
      */
-    scrollToElement(elementId) {
+    scrollToElement(objectId) {
       this.get('port').send('view:scrollToElement', {
-        elementId
+        elementId: objectId // TODO: what?
```
rwwagner90: @nummi did we come to a consensus on the elementId vs objectId thing here?

app/templates/components/component-tree-item.hbs
```diff
@@ -31,18 +33,18 @@
       }}
     </button>
 
-    {{#if hasElement}}
+    {{#if (eq @item.view.tagName "")}}
```
rwwagner90: Why the switch here?

app/templates/components/component-tree-item.hbs
```diff
@@ -1,28 +1,30 @@
...
-    {{item.view.name}}
+    {{@item.view.name}}
   </code>
 
-  {{#if item.view.isComponent}}
+  {{#if @item.view.isComponent}}
     <button
       class="component-tree-item__action js-scroll-into-view"
       title="Scroll into view"
-      {{action scrollToElement item.view.objectId bubbles=false}}
```
rwwagner90: Shouldn't we still pass the objectId here? If not, how does it know what to scroll to?

nummi: https://github.com/emberjs/ember-inspector/pull/984/commits/87c19898a1d8dfc89794956107327eb6eec7a5dc#diff-6a9fa9ada69626e649ae996c3753adbbR7

^ In the `component-tree-item` template, the `scrollToElement` attribute has the objectId already rolled into it from the invocation of `<ComponentTreeItem />`

app/templates/components/component-tree-item.hbs
nummi: I did this so I could kill `component-tree-item.js`.

app/controllers/component-tree.js
nummi: Not in this specific case. This seems like a boo boo that happens to work (most of the time?).

app/templates/components/component-tree-item.hbs
rwwagner90: Ahhh, gotcha 👍

app/templates/components/component-tree-item.hbs
rwwagner90: That was my concern, that this might not work later, and it's not clear, to the casual observer what this check is for necessarily. I would prefer to keep something like `hasElement` and heavily document what we are checking for in the JS. That way, we can keep things from getting murky as we rewrite things. I want it to all be very well documented 😃

app/controllers/component-tree.js
rwwagner90: Gotcha. Perhaps @rwjblue could jump on a quick call to discuss this sometime this week? I want to make sure we get all the `objectId` vs `elementId` stuff right and try to future proof and get things ready for glimmer components and such.

rwjblue: ya, happy to chat in higher bandwidth

app/templates/components/component-tree-item.hbs
nummi: I created a helper, is that cool?

rwwagner90: A helper sounds good 👍

rwjblue:
FYI - I think we clarified the `objectId` / `elementId` stuff in todays call.

PR #992, opened 2019-06-21
Component Tree: Angle Brackets and Classified Names

nummi:
Screen Shot 2019-06-21 at 1 33 07 PM

### DIFF
--- a/app/helpers/console.js
+++ b/app/helpers/console.js
@@ -1,7 +1,0 @@
-import { helper } from '@ember/component/helper';
-
-export function console(_) {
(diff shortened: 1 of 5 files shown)


app/helpers/console.js
```diff
@@ -1,7 +0,0 @@
-import { helper } from '@ember/component/helper';
-
-export function console(_) {
-  return console.log(_);
-}
-
-export default helper(console);
```
nummi: There is a built-in helper called `log`.

PR #993, opened 2019-06-24
Update a bunch of deps

rwwagner90:


rwwagner90:
@stefanpenner are you familiar with how to fix this error? `Using removed Babel 6 option: .resolveModuleSource - Use `babel-plugin-module-resolver@3`'s 'resolvePath' options`

PR #996, opened 2019-06-26
Refactor render debug

rwwagner90:
I started renaming and commenting things and trying to split things out a bit. Hopefully this is step one of many of making this easier to follow!


rwwagner90:
@sivakumar-kailasam that sounds like a good idea, but can we go ahead and get this in first? Trying to do things incrementally.

Issue #999, opened 2019-07-15
ember inspector fails to inspect certain components


AbhinavVishak:
It looks like when the inspector tries to figure out the computed property's dependent keys, it fails.

```
Ember Inspector has errored.
This is likely a bug in the inspector itself.
You can report bugs at https://github.com/emberjs/ember-inspector.
Error message: Cannot read property '_dependentKeys' of undefined
Stack trace: TypeError: Cannot read property '_dependentKeys' of undefined
    at addProperties (<anonymous>:3159:78)
    at <anonymous>:3120:11
    at Array.forEach (<anonymous>)
    at propertiesForMixin (<anonymous>:3118:20)
    at Class.mixinsForObject (<anonymous>:3008:22)
    at Class.sendObject (<anonymous>:2927:26)
    at Class.inspectById (<anonymous>:2851:16)
    at sendEvent (http://localhost:4200/assets/vendor.js:37100:14)
    at Class.trigger (http://localhost:4200/assets/vendor.js:51786:28)
    at <anonymous>:3502:14
```

I think it might be related to my use of `ember-redux` and connect's `stateToComputed` paradigm. 

I tried following through the code and it looks like the `descriptorForDecorator` call in https://github.com/emberjs/ember-inspector/blob/aede612ca6698b83300f3b54ceeae3606dde3613/ember_debug/utils/type-check.js#L33 is the issue. Could somebody clarify if this is something that needs to be fixed on my side, or just not currently supported in the inspector ?

rwwagner90:
@AbhinavVishak What version of Ember are you on?

AbhinavVishak:
@rwwagner90 I'm running `v3.10.1` on linux

AbhinavVishak:
I'm guessing that since the computed property is being set through an 3rd party addon `ember-redux`, the `DECORATOR_DESCRIPTOR_MAP` is not being populated properly, causing the `undefined ` error.

rwwagner90:
@AbhinavVishak would it be possible to share your code or create a small reproduction?

AbhinavVishak:
I tried creating a simple hello-world program for ember-redux, and was able to reproduce it.

```js
import Component from '@ember/component';
import { connect } from 'ember-redux';

const stateToComputed = (state) => ({
  computedField: 'hardcoded'
});
const Qwerty = Component.extend({
});
export default connect(stateToComputed)(Qwerty);
```

```html
<h2>foo</h2>
[...]

PR #1002, opened 2019-08-01
Add undefined check for getDescriptorFor()

AbhinavVishak:
Fixes https://github.com/emberjs/ember-inspector/issues/999.

`getDescriptorFor()` returns undefined for certain properties ( like computed properties through _ember-redux_, for example), so wrap it with a logical OR to prevent the _dependentKeys accessor from throwing an exception.

### DIFF
--- a/ember_debug/utils/type-check.js
+++ b/ember_debug/utils/type-check.js
@@ -1,4 +1,5 @@
 const Ember = window.Ember;
+const { descriptorForDecorator, descriptorForProperty } = Ember.__loader.require('@ember/-internals/metal');
 const { ComputedProperty } = Ember;
 
 /**
@@ -30,7 +31,7 @@ export function isDescriptor(value) {
  */
 export function getDescriptorFor(object, key) {
   if (Ember.Debug.isComputed) {
-    return Ember.__loader.require('@ember/-internals/metal').descriptorForDecorator(object[key]);
+    return descriptorForDecorator(object[key]) || descriptorForProperty(object, key);
   }
 
   return object[key];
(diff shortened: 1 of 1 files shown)


rwwagner90:
@pzuraq is this the best fix for this?

pzuraq:
We should replace: https://github.com/emberjs/ember-inspector/blob/master/ember_debug/utils/type-check.js#L33

with:

```
let { descriptorForDecorator, descriptorForProperty } = Ember.__loader.require('@ember/-internals/metal');

return descriptorForProperty(object[key]) || descriptorForDecorator(object[key]);
```

rwwagner90:
@AbhinavVishak do you mind updating the PR to do the changes @pzuraq suggested?

AbhinavVishak:
Sure. Let me try that and see how it behaves.

rwwagner90:
@AbhinavVishak did those changes work well for your use case?

AbhinavVishak:
@rwwagner90, sorry I haven't got a chance to test them out yet. Shall I leave a comment when I do so ?
Also, shouldn't the conditionals be flipped ? `descriptorForProperty(object, key) || descriptorForDecorator(object[key]);`

Check for a normal decorator first, and if not available then check for a property ?

AbhinavVishak:
seems to work as expected. shows the appropriate dependent key for a computed property too. 👍

rwwagner90:
> Also, shouldn't the conditionals be flipped to to
> 
> `descriptorForDecorator(object[key]) || descriptorForProperty(object, key);` ?

@pzuraq does the ordering matter? Should we flip them or leave them as is?

PR #1003, opened 2019-08-02
Upgrade to Ember Table 2.1.0

mixonic:
Bump to Ember Inspector 2.1.0. I don't think there are major fixes associated with bump that apply here (most recent fixes are around selection) but the row buffer is dropped to 0 upstream which makes things feel snappy.

And in general it gets off the pinned SHA.

I did smoke test this locally in addition to running the suite. Looks good.

We need a new beta of `vertical-collection` to be released since the `yarn.lock` now has both a SHA and version pinned which is a bit messy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "ember-source": "~3.10.0",
     "ember-source-channel-url": "^1.1.0",
     "ember-svg-jar": "^1.2.2",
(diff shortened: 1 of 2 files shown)


rwwagner90:
@mixonic do you want me to cut a release of vertical-collection?

PR #1005, opened 2019-08-07
Guard against reading a property on undefined, fixes #1004

### LINKED ISSUE #1004: Cannot read property 'reopen' of undefined
Ran into this today:

```
"after each" hook for "Scenario: Applying a coupon with a free product"‣
TypeError: Cannot read property 'reopen' of undefined
    at Class.dropObject (<anonymous>:2978:18)
    at Class.willDestroy (<anonymous>:2951:18)
    at Class.superWrapper (assets/vendor.js:31574:22)
    at invoke (assets/vendor.js:54425:16)
    at Queue.flush (assets/vendor.js:54317:13)
    at DeferredActionQueues.flush (assets/vendor.js:54521:21)
    at Backburner._end (assets/vendor.js:55097:34)
    at Backburner.end (assets/vendor.js:54783:12)
    at Backburner._run (assets/vendor.js:55152:16)
    at Backburner.run (assets/vendor.js:54832:19)
```

Took me quite a while to figure out that the issue is caused by Ember Inspector. How freaky is that the Inspector breaks your tests! 😱 

Found a report of this issue from four 😱😱 years ago: https://github.com/emberjs/ember.js/issues/11863#issuecomment-123888938 @stefanpenner's reply is very interesting, though irrelevant to the cause of this particular problem.

Inspector crashes here: https://github.com/emberjs/ember-inspector/blob/f1ab36afb6abcdba53fb0180af5ed0203432dd73/ember_debug/object-inspector.js#L328-L334

Simply closing and reopening the inspector resulted in the issue not happening again, so I'm unable to share a reproduction.

But the fix is obvious, natural and free from side effects. I'll make a PR.

lolmaus:
Fixes #1004

`!= null` checks for both `null` and `undefined`. It's faster than simply `if (object)`.

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -328,7 +328,7 @@ export default EmberObject.extend(PortMixin, {
   dropObject(objectId) {
     let object = this.sentObjects[objectId];
 
-    if (object.reopen) {
+    if (object && object.reopen) {
       object.reopen({ willDestroy: object._oldWillDestroy });
       delete object._oldWillDestroy;
     }


rwwagner90:
@lolmaus would it be that slow to check `if(object && object.reopen)`? I feel like that's a more straightforward syntax, if it is not too much of a performance hit.

lolmaus:
@rwwagner90, changed.

PR #1006, opened 2019-08-11
Add ES6 Class support

patricklx:


ember_debug/object-inspector.js
```diff
@@ -220,7 +220,7 @@ export default EmberObject.extend(PortMixin, {
   },
 
   canSend(val) {
-    return (val instanceof EmberObject) || typeOf(val) === 'array';
+    return val && ((val instanceof Object) || typeOf(val) === 'array');
```
rwwagner90: Should we check both `Object` and `EmberObject` to account for everything?

patricklx: done, also pushed an update the better check if `class extends` or `Ember.*.extend` was used by checking if protoype has `_super`

rwwagner90:
@pzuraq @rwjblue do you mind giving this a quick look? We definitely should add support for classes, but I know we have various different flavors of everything, and wanted to make sure we support all the things here, as well as simplify anything we can.

rwwagner90:
@patricklx do you mind fixing the conflicts real quick? Sorry about that!

PR #1022, opened 2019-09-08
Modernize UI Components

nummi:
Use @, this, `on` in place of `action`, angle brackets, `tagName=''`, etc. for Inspectors UI add-on.


app/templates/application.hbs
```diff
@@ -28,11 +28,12 @@
           <div class="split__panel">
             <div class="split__panel__hd">
               {{outlet "toolbar"}}
-              {{ui/sidebar-toggle
```
nummi: Used only once so I removed it.

app/templates/promise-tree-toolbar.hbs
```diff
@@ -60,16 +48,18 @@
     Fulfilled
   </button>
 
-  <div class="divider" aria-hidden="true"></div>
+  <Ui::ToolbarDivider />
 
   <div class="toolbar__checkbox js-with-stack">
     <label for="instrument-with-stack">
-      {{ui/action-checkbox
```
nummi: Used only once so I removed it.

lib/ui/addon/components/error-page.js
```diff
@@ -3,12 +3,5 @@ import layout from '../templates/components/error-page';
 
 export default Component.extend({
   layout,
-
-  didInsertElement() {
```
nummi: Just use CSS for this, duh.

lib/ui/addon/templates/components/error-page.hbs
```diff
@@ -2,18 +2,14 @@
...
       <div class="error-page__title js-error-page-title">
-        {{description}}
+        {{@description}}
       </div>
       <img class="tomster" src="assets/images/fishy_tomster.png" alt="Fishy Tomster">
     </div>
 
     <div class="error-page__body">
       <div class="error-page__body-title">
-        {{#if bodyTitle}}
```
nummi: `bodyTitle` is never used by the app.

lib/ui/addon/components/toolbar-search-field.js
```diff
@@ -1,12 +1,20 @@
+import { action } from '@ember/object';
 import Component from '@ember/component';
 import layout from '../templates/components/toolbar-search-field';
 
+let uuid = 0;
```
nummi: _Robert Jackson Suggested_ ™️

lib/ui/addon/components/toolbar-divider.js
```diff
@@ -0,0 +1,7 @@
+import Component from '@ember/component';
```
rwwagner90: Could we make this a template-only-glimmer-component and remove this JS file?

lib/ui/addon/templates/components/toolbar-search-field.hbs
```diff
@@ -1,11 +1,19 @@
-{{input value=value type="text" placeholder="Search"}}
+<div class={{concat "toolbar__search " @wrapperClass}}>
```
rwwagner90: Instead of `@wrapperClass` I think we can use normal `class`. Then I think if you pass `class="foo"` instead of `@wrapperClass="foo"` you should get both classes added to the element.
```suggestion
<div class="toolbar__search">
```

lib/ui/addon/templates/components/toolbar-search-field.hbs
```diff
@@ -1,11 +1,19 @@
...
+    placeholder="Search"
+  />
+
+  {{#if @value}}
+    <button
+      class="toolbar__icon-button toolbar__search-clear-button js-search-field-clear-button"
+      title="clear"
+      {{on "click" this.clear}}
+    >
+      {{svg-jar "clear" width="14px" height="14px"}}
```
rwwagner90: Is `svg-jar` angle bracket compatible? If so, could we use `<SvgJar`?

lib/ui/addon/components/error-page.js
rwwagner90: Can you explain more about this?

app/templates/promise-tree-toolbar.hbs
rwwagner90: 🔥

lib/ui/addon/components/toolbar-divider.js
nummi: I'm not sure we can do that until 3.14. 🤷‍♀

lib/ui/addon/templates/components/toolbar-search-field.hbs
nummi: I was torn on this. If you were the end-user of `<ToolbarSearchField />` would you think `class` would be applied to the `<input>`?

lib/ui/addon/components/toolbar-divider.js
rwwagner90: We can, it has been a feature of https://github.com/emberjs/ember-optional-features for awhile. I think we need to run `ember feature:enable template-only-glimmer-components` then delete the JS file.

lib/ui/addon/templates/components/toolbar-search-field.hbs
rwwagner90: I think I would look at the template before assuming, so I think it should be fine to use `class` directly, rather than introducing a new variable.

lib/ui/addon/templates/components/toolbar-search-field.hbs
nummi: `svg-jar` is a helper.

lib/ui/addon/components/error-page.js
nummi: I just use CSS to animate Tomster in place of manual DOM stuff.

rwwagner90: Gotcha, sounds good 👍

lib/ui/addon/templates/components/toolbar-search-field.hbs
nummi: Change made.

lib/ui/addon/components/toolbar-divider.js
nummi: The plan in my head was to made a pass at components like I did in this PR, then go over again and _Glimmerize_ in another PR.

rwwagner90: I would just handle this one instance now, so we can get rid of the JS file, which we are only keeping around for `tagName: ''`

nummi: Done and done.

Issue #1021, opened 2019-09-08
Component Tree Search


nummi:
We should match on Angle bracket style _and_ curly style.

<img width="278" alt="image" src="https://user-images.githubusercontent.com/3692/64492454-cc529600-d228-11e9-9f69-12c085f6db2a.png">

<img width="279" alt="image" src="https://user-images.githubusercontent.com/3692/64492460-d8d6ee80-d228-11e9-849a-18b7208c7864.png">

## `LinkTo` no results:

<img width="282" alt="image" src="https://user-images.githubusercontent.com/3692/64492463-e0969300-d228-11e9-9773-33325db114a3.png">

## `Link-to` finds components

<img width="280" alt="image" src="https://user-images.githubusercontent.com/3692/64492470-e8563780-d228-11e9-870b-dacce093abff.png">

nummi:
ComponentTree and a few other tabs (like RouteTree and Deprecations) use [search-match.js](https://github.com/emberjs/ember-inspector/blob/master/app/utils/search-match.js) — would there be any harm in ignoring dashes entered by the user?

PR #1023, opened 2019-09-08
Ignore dashes in search

nummi:
https://github.com/emberjs/ember-inspector/issues/1021


rwwagner90:
@nummi could we possibly add some quick tests for searching some terms with and without `-`?

nummi:
@rwwagner90 Added some unit tests — is that good or would you like to see an acceptance test?

rwwagner90:
> Added some unit tests — is that good or would you like to see an acceptance test?

These look good! Just trying to get some tests added whenever we change something, to ensure no regressions in the future 👍

PR #1024, opened 2019-09-10
Tidy up object inspector

nummi:
- `this, @, {{on}}, < />`, etc.
- namespace children components under ObjectInspector
- child components of ObjectInspector were passing a reference of themselves to their children, weird:


nummi:
I say 🤷‍♀ to the Code Climate issues.

app/components/object-inspector/property-field.js
```diff
@@ -0,0 +1,9 @@
+import TextField from '@ember/component/text-field';
+
+export default TextField.extend({
+  didInsertElement() {
+    this.element.select();
```
rwwagner90: Can you explain what this select is doing?

PR #1025, opened 2019-09-13
Tidy up render-tree tab

nummi:
- fix table striping
- remove observer
- fix search highlighting
- action, this, etc.

I simply removed the observer for now because the functionality it provided is not very helpful:
search

I believe this was meant to expand _every_ list item. In the future we could expand the tree where necessary.


lib/ui/addon/styles/_list.scss
```diff
@@ -58,6 +58,12 @@
   }
 }
 
+.list.list--css-striping {
```
nummi: This just seemed like the easiest way to get striping working again.

app/controllers/render-tree.js
```diff
@@ -79,15 +60,19 @@ export default Controller.extend({
...
-        return true;
-      }
-      let regExp = new RegExp(search);
+    if (isEmpty(this.escapedSearch)) {
+      return this.model;
+    }
+
+    return this.model.filter((item) => {
+      const regExp = new RegExp(this.escapedSearch);
       return !!recursiveMatch(item, regExp);
```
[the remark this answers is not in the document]
rwjblue: Ya, its not needed.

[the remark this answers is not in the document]
rwwagner90: @nummi if it's not needed, should we go ahead and remove?

tests/acceptance/render-tree-test.js
```diff
@@ -114,24 +114,6 @@ module('Render Tree Tab', function(hooks) {
     assert.dom(rows[0].querySelector('.js-render-profile-name')).hasText('First View Rendering');
     assert.dom(rows[1].querySelector('.js-render-profile-name')).hasText('Second View Rendering');
 
-    await fillIn('.js-render-profiles-search input', 'first');
```
rwwagner90: @nummi do we not need these tests anymore?

lib/ui/addon/styles/_list.scss
rwwagner90: Is there not an existing class for striping?

nummi: We do. It’s based on passing a boolean to a row. I think I was struggling with this since the dataset is nested.

tests/acceptance/render-tree-test.js
nummi: Search filtering is still tested in the lines below these — the code removed here was testing the (broken, see GIF) functionality of opening of children rows.

PR #1030, opened 2019-09-18
Tidy up deprecations tab

nummi:
- `deprecation-item` as template-only component
- remove observer from deprecations controller
- `vertical-collection` in place of List table
- `this, {{on}}, action, <Component/>` etc.
- UI more readable (IMHO)

### With source maps:

Screen Shot 2019-09-18 at 11 41 43 AM

### Without source maps:

image

### Old and busted:

image


app/styles/utils.scss
```diff
@@ -0,0 +1,71 @@
+$units: 0, 1, 2, 3, 4;
```
rwwagner90: I would prefer to add tailwind, rather than maintaining our own custom utils. What would your thoughts be on that?

app/controllers/deprecations.js
```diff
@@ -1,54 +1,42 @@
...
   search: null,
   searchValue: debounceComputed('search', 300),
+  toggleDeprecationWorkflow: false,
 
   filtered: computed('model.@each.message', 'search', function() {
     return this.model
       .filter((item) => searchMatch(get(item, 'message'), this.search));
   }),
 
-  optionsChanged: observer('options.toggleDeprecationWorkflow', function() {
```
rwwagner90: Always nice to see observers removed 😄

rwwagner90:
We can switch to tailwind later, going ahead and merging.

PR #1034, opened 2019-09-22
Tidy up data tab

nummi:


app/components/record-filter.js
```diff
@@ -1,8 +0,0 @@
-import { computed } from '@ember/object';
-import Component from '@ember/component';
-export default Component.extend({
```
nummi: This component is replaced with a simple comparison in the template: https://github.com/emberjs/ember-inspector/pull/1034/files?file-filters%5B%5D=.hbs&hide-deleted-files=true#diff-97c2c032aa308abb534f8894b734348eR18

app/helpers/is-empty.js
```diff
@@ -0,0 +1,8 @@
+import { isEmpty } from '@ember/utils';
```
rwwagner90: ember-truth-helpers already has an `is-empty` helper. Can we use that one instead?

app/controllers/records.js
```diff
@@ -51,22 +61,9 @@ export default Controller.extend({
...
-
-      // check search
-      if (!isEmpty(search)) {
-        let searchString = this.recordToString(item);
-        return !!searchString.match(new RegExp(`.*${escapeRegExp(search.toLowerCase())}.*`));
-      }
-      return true;
+  filteredRecords: computed('searchValue', 'model.@each.{columnValues,filterValues}', 'filterValue', function() {
+    return this.model.filter((record) => {
+      return this.passesFilter(record) && this.passesSearch(record);
```
rwwagner90: The logic for `passesFilter` and `passesSearch` seems fairly simple. Would you be okay with not pulling out into separate functions and leaving inside the computed?

PR #1035, opened 2019-09-22
Support tracked and other improvements

patricklx:
* Dont use observe for update properties
* Add icons for tracked, function, getter
* support computed that return undefined
* sort and group properties by type. 
Service -> Tracked -> Computed -> Getter -> Property -> Function
* support complex getters, can always be calculcated when needed
* include functions, (why not?)

### DIFF
--- a/app/components/object-inspector/property.js
+++ b/app/components/object-inspector/property.js
@@ -12,9 +12,7 @@ export default Component.extend({
   txtValue: null,
   dateValue: null,
 
-  isCalculated: computed('valueType', function () {
-    return this.valueType !== 'type-descriptor';
-  }),
+  isCalculated: alias('model.value.isCalculated'),
 
   valueType: alias('model.value.type'),
 
@@ -30,7 +28,7 @@ export default Component.extend({
 
   isInstance: equal('valueType', 'type-instance'),
 
-  isComputedProperty: alias('model.value.computed'),
+  isComputedProperty: alias('model.isComputed'),
 
   isFunction: equal('valueType', 'type-function'),
 
@@ -46,6 +44,42 @@ export default Component.extend({
 
   showDependentKeys: and('isDepsExpanded', 'hasDependentKeys'),
 
+  canCalculate: computed('model', 'isCalculated', 'isComputedProperty', function () {
+    if (this.get('isOverridden')) return false;
+    if (!this.get('isComputedProperty') && this.get('model.isGetter') && this.get('model.isExpensive')) {
+      return true;
+    }
+    return this.get('isComputedProperty') && !this.get('isCalculated');
+  }),
+
+  iconInfo: computed('isService', 'model.inspect.value', 'model.isTracked', 'model.isProperty', 'model.isGetter', 'isFunction', function () {
+    if (this.get('isService')) {
+      return { type: 'service', title: 'Service' };
+    }
+
+    if (this.get('isFunction')) {
+      return { type: 'function', title: 'Function' };
+    }
+
+    if (this.get('model.isTracked')) {
+      return { type: 'tracked', title: 'Tracked' };
+    }
+
+    if (this.get('model.isProperty')) {
+      return { type: 'property', title: 'Property' };
+    }
+
+    if (this.get('model.isComputed')) {
+      return { type: 'computed', title: 'Computed' };
+    }
+
+    if (this.get('model.isGetter')) {
+      return { type: 'getter', title: 'Getter' };
+    }
+
+    return { type: 'n/a', title: 'N/A' };
+  }),
+
   canDig() {
     return this.isInstance
       || this.isObject
@@ -54,11 +88,14 @@ export default Component.extend({
   },
 
   cannotEdit() {
+    if (!this.isCalculated) return true;
     return this.isFunction || this.isOverridden || this.readOnly;
   },
 
   toggleDeps: action(function () {
-    this.toggleProperty('isDepsExpanded');
+    if (this.hasDependentKeys) {
+      this.toggleProperty('isDepsExpanded');
+    }
   }),
 
   valueClick: action(function () {
--- a/app/components/object-inspector/sort-properties.js
+++ b/app/components/object-inspector/sort-properties.js
@@ -1,27 +1,55 @@
 import Component from '@ember/component';
-import { sort } from '@ember/object/computed';
+import { sort, map } from '@ember/object/computed';
+import { set, computed } from '@ember/object';
+import { A } from '@ember/array';
 
 export default Component.extend({
   tagName: '',
 
+  isArray: computed('properties', function () {
+    const props = A(this.get('properties') || []);
+    return props.findBy('name', 'length') && props.findBy('name', 0);
+  }),
+
   /**
-   * Sort the properties by name to make them easier to find in the object inspector.
+   * Sort the properties by name and group them by property type to make them easier to find in the object inspector.
    *
    * @property sortedProperties
    * @type {Array<Object>}
    */
-  sortedProperties: sort('properties', 'sortProperties'),
+  sortedProperties: sort('props', 'sortProperties'),
 
-  init() {
-    this._super(...arguments);
+  props: map('properties', function (p) {
+    set(p, 'isFunction', p.value.type === 'type-function');
+    if (p.name == parseInt(p.name)) {
+      set(p, 'name', parseInt(p.name));
+    }
+    return p;
+  }),
 
-    /**
-     * Used by the `sort` computed macro.
-     *
-     * @property sortProperties
-     * @type {Array<String>}
-     */
-    this.sortProperties = ['name'];
-  },
+  /**
+   * Used by the `sort` computed macro.
+   *
+   * @property sortProperties
+   * @type {Array<String>}
+   */
+  sortProperties: computed('isArray', function () {
+    const order = [
+      'isFunction',
+      'isService:desc',
+      'isProperty:desc',
+      'isTracked:desc',
+      'isComputed:desc',
+      'isGetter:desc',
+      'name'
+    ];
+    // change order for arrays, if the array doesnt have items, then the order does not need to be changed
+    if (this.get('isArray')) {
+      const i = order.indexOf('isProperty:desc');
+      order.splice(i, 1);
+      order.splice(order.length - 1, 0, 'isProperty:desc');
+    }
+    return order;
+  })
 });
 
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -66,9 +66,15 @@ export default Route.extend({
   },
 
   updateProperty(options) {
-    const detail = this.get('controller.mixinDetails.mixins').objectAt(options.mixinIndex);
-    const property = get(detail, 'properties').findBy('name', options.property);
-    set(property, 'value', options.value);
+    if (this.get('controller.mixinDetails.mixins')) {
+      const detail = this.get('controller.mixinDetails.mixins').objectAt(options.mixinIndex);
+      let property = get(detail, 'properties').findBy('name', options.property);
+      if (!property) return;
+      set(property, 'value', options.value);
+      if (options.dependentKeys) {
+        set(property, 'dependentKeys', options.dependentKeys);
+      }
+    }
   },
 
   updateErrors(options) {
--- a/app/styles/object_inspector.scss
+++ b/app/styles/object_inspector.scss
@@ -126,3 +126,27 @@
     content: "P";
   }
 }
+
+.mixin__property-icon--getter {
+  background-color: var(--spec07);
+
+  &:before {
+    content: "G";
+  }
+}
+
+.mixin__property-icon--function {
+  background-color: var(--base08);
+
+  &:before {
+    content: "ƒ";
+  }
+}
+
+.mixin__property-icon--tracked {
+  background-color: var(--tracked-icon-bg);
+
+  &:before {
+    content: "T";
+  }
+}
--- a/app/templates/components/object-inspector.hbs
+++ b/app/templates/components/object-inspector.hbs
@@ -17,7 +17,7 @@
       <button
         class="send-to-console js-send-object-to-console-btn"
         title="Send to Console"
-        {{on "click" (fn this.sendObjectToConsole @model.firstObject)}}
+        {{on "click" (fn this.sendObjectToConsole @model.lastObject)}}
       >
         {{svg-jar "send-with-text" width="20px" height="10px"}}
       </button>
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -5,10 +5,20 @@ import { isComputed, isDescriptor, getDescriptorFor } from 'ember-debug/utils/ty
 const Ember = window.Ember;
 const {
   Object: EmberObject, inspect: emberInspect, meta: emberMeta, typeOf,
-  computed, get, set, guidFor, isNone, removeObserver,
-  Mixin, addObserver, cacheFor, VERSION
+  computed, get, set, guidFor, isNone,
+  Mixin, cacheFor, VERSION
 } = Ember;
 const { oneWay } = computed;
+
+let glimmer;
+let metal;
+try {
+  glimmer = Ember.__loader.require('@glimmer/reference');
+  metal = Ember.__loader.require('@ember/-internals/metal');
+} catch (e) {
+  glimmer = null;
+  metal = null;
+}
 
 const keys = Object.keys || Ember.keys;
 
@@ -21,10 +31,14 @@ const keys = Object.keys || Ember.keys;
  */
 function inspectValue(object, key, computedValue) {
   let string;
-  const value = computedValue ? computedValue : object[key];
+  const value = computedValue;
+
+  if (arguments.length === 3 && computedValue === undefined) {
+    return { type: `type-undefined`, inspect: 'undefined' };
+  }
 
   // TODO: this is not very clean. We should refactor calculateCP, etc, rather than passing computedValue
-  if (computedValue) {
+  if (computedValue !== undefined) {
     return { type: `type-${typeOf(value)}`, inspect: inspect(value) };
   }
 
@@ -32,9 +46,9 @@ function inspectValue(object, key, computedValue) {
     return { type: 'type-ember-object', inspect: value.toString() };
   } else if (isComputed(object, key)) {
     string = '<computed>';
-    return { type: 'type-descriptor', inspect: string, computed: true };
+    return { type: 'type-descriptor', inspect: string };
   } else if (isDescriptor(value)) {
-    return { type: 'type-descriptor', inspect: value.toString(), computed: true };
+    return { type: 'type-descriptor', inspect: value.toString() };
   } else {
     return { type: `type-${typeOf(value)}`, inspect: inspect(value) };
   }
@@ -100,6 +114,51 @@ function inspect(value) {
   }
 }
 
+function isMandatorySetter(descriptor) {
+  if (descriptor.set && descriptor.set === Ember.MANDATORY_SETTER_FUNCTION) {
+    return true;
+  }
+  if (descriptor.set && Function.prototype.toString.call(descriptor.set).includes('You attempted to update')) {
+    return true;
+  }
+  return false;
+}
+
+function getTagTrackedProps(tag, ownTag, level=0) {
+  const props = [];
+  if (!tag || level > 1) {
+    return props;
+  }
+  if (tag.subtag) {
+    if (tag.subtag._propertyKey) props.push(tag.subtag._propertyKey);
+    props.push(...getTagTrackedProps(tag.subtag, level + 1));
+  }
+  if (tag.subtags) {
+    tag.subtags.forEach((t) => {
+      if (t === ownTag) return;
+      if (t._propertyKey) props.push(t._propertyKey);
+      props.push(...getTagTrackedProps(t, level + 1));
+    });
+  }
+  return props;
+}
+
+function getTrackedDependencies(object, property, tag) {
+  const proto = Object.getPrototypeOf(object);
+  if (!proto) return [];
+  const cpDesc = Ember.meta(object).peekDescriptors(property);
+  const dependentKeys = [];
+  if (cpDesc) {
+    dependentKeys.push(...(cpDesc._dependentKeys || []));
+  }
+  if (metal) {
+    const ownTag = metal.tagForProperty(object, property);
+    dependentKeys.push(...getTagTrackedProps(tag, ownTag));
+  }
+
+  return [...new Set([...dependentKeys])];
+}
+
 export default EmberObject.extend(PortMixin, {
   namespace: null,
 
@@ -107,10 +166,72 @@ export default EmberObject.extend(PortMixin, {
 
   port: oneWay('namespace.port'),
 
+  currentObject: null,
+
+  updateCurrentObject() {
+    if (this.currentObject) {
+      const {object, mixinDetails, objectId} = this.currentObject;
+      mixinDetails.forEach((mixin, mixinIndex) => {
+        mixin.properties.forEach(item => {
+          if (item.overridden) {
+            return true;
+          }
+          try {
+            let cache = cacheFor(object, item.name);
+            if (item.isExpensive && !cache) return true;
+            if (item.value.type === 'type-function') return true;
+
+            let value = null;
+            let changed = false;
+            const values = this.objectPropertyValues[objectId] = this.objectPropertyValues[objectId] || {};
+            const tracked = this.trackedTags[objectId] = this.trackedTags[objectId]  || {};
+
+            const desc = Object.getOwnPropertyDescriptor(object, item.name);
+            const isSetter = desc && isMandatorySetter(desc);
+
+            if (glimmer && glimmer.validate && metal.track && item.canTrack && !isSetter) {
+              let tagInfo = tracked[item.name] || { tag: metal.tagForProperty(object, item.name), revision: 0 };
+              if (!tagInfo.tag) return;
+
+              changed = !glimmer.validate(tagInfo.tag, tagInfo.revision);
+              if (changed) {
+                tagInfo.tag = metal.track(() => {
+                  value = get(object, item.name);
+                });
+                tagInfo.revision = glimmer.value(object, item.name);
+              }
+              tracked[item.name] = tagInfo;
+            } else {
+              value = calculateCP(object, item.name, {});
+              if (values[item.name] !== value) {
+                changed = true;
+                values[item.name] = value;
+              }
+            }
+
+            if (changed) {
+              value = inspectValue(object, item.name, value);
+              value.isCalculated = true;
+              let dependentKeys = null;
+              if (tracked[item.name]) {
+                dependentKeys = getTrackedDependencies(object, item.name, tracked[item.name].tag);
+              }
+              this.sendMessage('updateProperty', { objectId, property: item.name, value, mixinIndex, dependentKeys });
+            }
+          } catch (e) {
+            // dont do anything
+          }
+        });
+      });
+    }
+    // workaround for tests, since calling any runloop inside runloop will prevent any `settled` to be called
+    setTimeout(() => Ember.run.next(this, this.updateCurrentObject), 300);
+  },
+
   init() {
     this._super();
     this.set('sentObjects', {});
-    this.set('boundObservers', {});
+    Ember.run.next(this, this.updateCurrentObject);
   },
 
   willDestroy() {
@@ -122,7 +243,11 @@ export default EmberObject.extend(PortMixin, {
 
   sentObjects: {},
 
-  boundObservers: {},
+  parentObjects: {},
+
+  objectPropertyValues: {},
+
+  trackedTags: {},
 
   _errorsFor: computed(function() { return {}; }),
 
@@ -140,8 +265,7 @@ export default EmberObject.extend(PortMixin, {
       value = this.valueForObjectProperty(message.objectId, message.property, message.mixinIndex);
       if (value) {
         this.sendMessage('updateProperty', value);
-        message.computed = true;
-        this.bindPropertyToDebugger(message);
+        message.isCalculated = true;
       }
       this.sendMessage('updateErrors', {
         objectId: message.objectId,
@@ -235,7 +359,7 @@ export default EmberObject.extend(PortMixin, {
     if (isNone(prop)) {
       value = this.sentObjects[objectId];
     } else {
-      value = get(object, prop);
+      value = calculateCP(object, prop, {});
     }
 
     this.sendValueToConsole(value);
@@ -255,15 +379,17 @@ export default EmberObject.extend(PortMixin, {
 
   digIntoObject(objectId, property) {
     let parentObject = this.sentObjects[objectId];
-    let object = get(parentObject, property);
+    let object = calculateCP(parentObject, property, {});
 
     if (this.canSend(object)) {
+      const currentObject = this.currentObject;
       let details = this.mixinsForObject(object);
+      this.parentObjects[details.objectId] = currentObject;
       this.sendMessage('updateObject', {
         parentObject: objectId,
         property,
         objectId: details.objectId,
-        name: object.toString(),
+        name: details.objectName,
         details: details.mixins,
         errors: details.errors
       });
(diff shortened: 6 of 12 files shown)


rwwagner90:
Hi @patricklx, thanks so much for the PR! Supporting tracked would be 💯

rwwagner90:
@patricklx I apologize, but I just merged a PR that moved around some things, so there are some conflicts you will need to resolve. Sorry about that!

nummi:
@patricklx, is this still a work in progress? When I try to run this locally I get this error: `Uncaught TypeError: desc.hasOwnProperty is not a function`

patricklx:
@nummi mmm, its working fine for me...
when does it happen? Do you select a specific object, component, service, mixin... ?

I have a project where I still have both types, EmberObject.extend or es6 class.
It's working everywhere for routes, controller, components and services...

patricklx:
Working on fixing the tests...

nummi:
https://deprecations.emberjs.com

image

```
VM77:3331 Uncaught TypeError: desc.hasOwnProperty is not a function
```

patricklx:
@rwwagner90 @nummi this should be quite ready now. Many changes...
I hope you agree with them  :)

app/styles/object_inspector.scss
```diff
@@ -126,3 +126,27 @@
...
+  &:before {
+    content: "G";
+  }
+}
+
+.mixin__property-icon--function {
+  background-color: var(--base08);
+
+  &:before {
+    content: "F";
```
nummi: How about "ƒ" ? :)

rwwagner90:
Thanks so much for this PR @patricklx! This is super amazing and we really appreciate your help! We will review this in the coming days and get it in ASAP 😃

patricklx:
@rwwagner90 it's ready now :)

rwwagner90:
@patricklx it looks like we have a couple test failures. I know you already fixed a lot of failures, but do you mind taking a look?

app/controllers/application.js
```diff
@@ -78,6 +78,9 @@ export default Controller.extend({
    * Called when inspecting an object from outside of the ObjectInspector
   */
   activateMixinDetails(name, objectId, details, errors) {
+    if (this.mixinStack.objectAt(0) && this.mixinStack.objectAt(0).objectId === objectId) {
```
nummi: What is this check preventing?

patricklx: If there are many renders, it will remove and add the same object in the inspector everytime, therefore making it impossible to go into properties and it also decreases the performance of the inspector. (If by some mistake you have 100ths of renders / s ...)

rwwagner90:
@pzuraq wanted to get your opinions on these changes, as they are around `tracked` and I don't know much about that yet.

patricklx:
parts for tracked: 
https://github.com/emberjs/ember-inspector/pull/1035/files#diff-7c2c7ea1dfc0ef384d1ec6e685711d15R163
and
https://github.com/emberjs/ember-inspector/pull/1035/files#diff-7c2c7ea1dfc0ef384d1ec6e685711d15R720

rwwagner90:
@patricklx have you been testing this against an app? If so, is it something we could also use for testing?

app/controllers/application.js
nummi: If you end up adding another commit to this PR would you mind putting in a comment?

patricklx:
Hi, @rwwagner90
Yes, I've been testing with an app.
Got the dev build up:
http://116.203.88.77:9020

patricklx:
seems it will take some time to review the tracked part?

otherwise, here are some hints for tracked if you want to review yourself:
https://github.com/emberjs/ember.js/blob/016f8066d0b484f3f69ce8c337be3b8b8aa5fce4/packages/%40ember/-internals/metal/lib/computed.ts#L528
https://github.com/emberjs/ember.js/blob/016f8066d0b484f3f69ce8c337be3b8b8aa5fce4/packages/%40ember/-internals/metal/lib/computed.ts#L532

also note  that tagForProperty will setup mandatory setters, which we want to prevent while using ember-inspector
https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/metal/lib/tags.ts#L36
https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/metal/lib/tracked.ts#L193

rwwagner90:
@wycats @chancancode @pzuraq @krisselden I requested reviews from all of you, but I think the main thing is @pzuraq would like at least one sanity check on these things. If you guys would like to have a meeting to briefly discuss these changes, we could do that as well, but figured we could try adhoc review and discussion first.

pzuraq:
One main concern I have at the moment is exposing the `Revision` for each tracked property. While that might be useful information in some sense, if you know how the system works, it definitely feels like we're leaking some implementation details that I'm not sure we're ready to leak just yet.

patricklx:
mmm, any alternative then? How can I track changes manually. Can I obtain some reference somhhow or create it myself?
Though I do not know enough about that part.
Something like this?
```js
const cachedRef = new glimmer.CachedReference();
cachedRef.tag = tagForProperty(...);
cachedRef.compute = () => null;
const ref = new glimmer.ReferenceCache(cachedRef);
// cache ref somewhere
const changed = glimmer.isModified(ref.revalidate());
```
would work i guess, but no idea :)

pzuraq:
Ah I see, I misread it initially, thought we were displaying the Revision for tracked properties directly to users. We're using it to detect when _getters_ have changed, which I wasn't expecting, but I guess makes sense/matches current behavior.

One thing I would like to make sure is that native getters are marked _distinctly_ from tracked properties. It's very useful to know if something is a derived value or a tracked property. Does it currently do that? Can you post a screenshot of what the new UI looks like also, just to get a sense of what it is?

patricklx:
Yes, getters and tracked are shown differently.
I also show dependentKeys for getters, but it can only show some properties without any object relation

Annotation 2019-10-09 162526
Annotation 2019-10-09 162515

app/components/object-inspector/accordion.js
```diff
@@ -4,12 +4,25 @@ import Component from '@ember/component';
 export default Component.extend({
   tagName: '',
 
-  isExpanded: computed('mixin.expand', 'mixin.properties.length', function () {
-    return this.get('mixin.expand') && this.get('mixin.properties.length') > 0;
+  _mixin: null,
+
+  isExpanded: undefined,
```
rwwagner90: Should this be `false` if not expanded is the default?

app/components/object-inspector/accordion.js
```diff
@@ -4,12 +4,25 @@ import Component from '@ember/component';
...
+
+  mixin: computed({
```
rwwagner90: I'm curious what this `mixin` computed is for? Aren't we passing `mixin` in here? https://github.com/emberjs/ember-inspector/pull/1035/files#diff-8ce1165ef4c2dc2a8fe46fd33be76df5R2

app/components/object-inspector/properties-all.hbs
```diff
@@ -2,7 +2,7 @@
   <ObjectInspector::SortProperties
     @properties={{this.flatPropertyList}} as |sortedProperties|
   >
-    {{#each sortedProperties as |prop|}}
+    {{#each sortedProperties key="name" as |prop|}}
```
rwwagner90: Do you mind explaining the need for the key here?

app/components/object-inspector/property.js
```diff
@@ -30,7 +28,7 @@ export default Component.extend({
 
   isInstance: equal('valueType', 'type-instance'),
 
-  isComputedProperty: alias('model.value.computed'),
+  isComputedProperty: alias('model.isComputed'),
```
rwwagner90: Does this not need to be `model.value.isComputed`?

app/components/object-inspector/property.js
```diff
@@ -46,6 +44,42 @@ export default Component.extend({
...
 
+  canCalculate: computed('model', 'isCalculated', 'isComputedProperty', function () {
+    if (this.get('isOverridden')) return false;
+    if (!this.get('isComputedProperty') && this.get('model.isGetter') && this.get('model.isExpensive')) {
+      return true;
+    }
+    return this.get('isComputedProperty') && !this.get('isCalculated');
+  }),
+
+  icon: computed('isService', 'model.inspect.value', 'model.isTracked', 'model.isProperty', 'model.isGetter', 'isFunction', function () {
```
rwwagner90: `icon` seems to be used for displaying strings. Perhaps it would be better named something like `itemType` since it is not always for an icon?

app/components/object-inspector/property.js
```diff
@@ -54,10 +88,13 @@ export default Component.extend({
   },
 
   cannotEdit() {
+    if (this.model.name === '...') return true;
+    if (!this.isCalculated) return true;
```
rwwagner90: Could we combine these lines?
```suggestion
   if (this.model.name === '...' || !this.isCalculated) return true;
```

app/components/object-inspector/property.js
```diff
@@ -54,10 +88,13 @@ export default Component.extend({
...
     return this.isFunction || this.isOverridden || this.readOnly;
   },
 
   toggleDeps: action(function () {
+    if (!this.hasDependentKeys) return;
```
rwwagner90: Rather than returning for the false case, could we wrap in the truthy case before calling toggleProperty?

app/components/object-inspector/sort-properties.js
```diff
@@ -21,7 +46,15 @@ export default Component.extend({
      * @property sortProperties
      * @type {Array<String>}
      */
-    this.sortProperties = ['name'];
+    this.sortProperties = [
+      'isFunction',
```
rwwagner90: I'm curious, how did you decide the ordering for sorting?

app/controllers/application.js
rwwagner90: Would it perhaps be a good idea to wrap in some sort of throttle or debounce? That way if we render hundreds of times, we could make it only actually render a few of those times.

app/routes/application.js
```diff
@@ -66,9 +66,14 @@ export default Route.extend({
   },
 
   updateProperty(options) {
+    if (!this.get('controller.mixinDetails.mixins')) return;
```
rwwagner90: Rather than this early return, could we wrap the rest of the function in  `if (this.get('controller.mixinDetails.mixins'))`?

app/templates/components/object-inspector.hbs
```diff
@@ -17,7 +17,7 @@
       <button
         class="send-to-console js-send-object-to-console-btn"
         title="Send to Console"
-        {{on "click" (fn this.sendObjectToConsole @model.firstObject)}}
+        {{on "click" (fn this.sendObjectToConsole @model.lastObject)}}
```
rwwagner90: I'm curious why this changed from first to last?

ember_debug/utils/type-check.js
```diff
@@ -9,8 +9,16 @@ const { ComputedProperty } = Ember;
...
-    return Ember.Debug.isComputed(object, key) || getDescriptorFor(object, key);
+  if (Ember.Debug.isComputed && Ember.Debug.isComputed(object, key)) {
+    return true;
+  }
+
+  if (Ember.meta(object) && Ember.meta(object).peekDescriptors(key)) {
+    return !!Ember.meta(object).peekDescriptors(key)._getter;
+  }
+
+  if (getDescriptorFor(object, key) instanceof ComputedProperty) {
```
rwwagner90: @pzuraq will this always be `instanceof ComputedProperty`? I thought there was something where things were not instances of that anymore, but maybe I am remembering wrong.

ember_debug/view-debug.js
```diff
@@ -264,18 +264,20 @@ export default EmberObject.extend(PortMixin, {
...
-      if (this.isDestroying) {
-        return;
-      }
-      this.releaseCurrentObjects();
-      let tree = this.viewTree();
-      if (tree) {
-        this.sendMessage('viewTree', { tree });
-      }
-    }, 50);
+    throttle(this, this.send, 500, false);
```
rwwagner90: Is the idea of this throttle to cut down on the number of renders?

tests/ember_debug/object-inspector-test.js
```diff
@@ -435,14 +456,11 @@ module('Ember Debug - Object Inspector', function(hooks) {
...
-    assert.equal(message.details[3].name, 'TestObject');
-    assert.equal(message.details[3].properties.length, 2, 'Does not duplicate properties');
-    assert.equal(message.details[3].properties[0].name, 'hasChildren');
-    assert.equal(message.details[3].properties[1].value.type, 'type-descriptor', 'Does not calculate expensive properties');
+    assert.equal(message.details[2].properties.length, 2, 'Correctly merges properties');
+    assert.equal(message.details[2].properties[0].value.isCalculated, undefined, 'Does not calculate expensive properties');
+    assert.equal(message.details[2].properties[1].name, 'hasChildren');
 
-    assert.ok(message.details[4].name !== 'MixinToSkip', 'Correctly skips properties');
+    assert.ok(message.details[3].name !== 'MixinToSkip', 'Correctly skips mixins');
```
rwwagner90: I'm curious why so many indices changed in the tests?

app/components/object-inspector/accordion.js
patricklx: mmm, it might not be needed anymore.
It was an early attempt to fix issues with rerenders from here:
https://github.com/emberjs/ember-inspector/pull/1035/files/96efd0d7e7e98103fcd44e6a696b238d57d6e3df#diff-a9a19ee37ba8619bc68db992b637dd05R87
where the view was reset on every time, with this workaround i was able to keep the current accordion open, but I think it's really fixed by the above linked fix. I test again and revert if this is working

app/components/object-inspector/sort-properties.js
patricklx: mmm, a mix between how it looks like (most times) in the code and whats useful:
1. services are always at the top
2. then comes properties, tracked
3. later computed and getters,   -> but I think computed/getters and tracked are more interesting than normal properties, so there are more at the top
4. finally functions at the bottom (sometimes useful if one wants to debug the function, then `send to console` and click on the function and chrome will send you directly to the source code) (i would even prefer if there was a `goto` for functions instead of `send to console` :) )

So, final order is:
Service -> Tracked -> Computed -> Getter -> Property -> Function

the `sortProperties` is more a short trial & error :)
had to put `isFunction` at the top, since its also a `isProperty`

app/controllers/application.js
patricklx: actually, I remember the real issue now. 
When a rerender happens, and i'm currently digging into an object, it will reset the object inspector. Throwing away my current inspection.
I will update the comment

app/templates/components/object-inspector.hbs
patricklx: I think it makes more sense to to send the current object to the console (the one I digged into).
But instead it sent the initial parent object.
Can we send the current object instead? :)

ember_debug/view-debug.js
patricklx: yes, though I want to reduce the lag to 300, which feels a bit better

app/components/object-inspector/properties-all.hbs
rwwagner90: It doesn't hurt anything to keep it, just curious 👍

app/templates/components/object-inspector.hbs
rwwagner90: I'm not sure I understand. Clicking `$E` in the current inspector gives you the correct object, right? If so, I am curious why this changed?

tests/ember_debug/object-inspector-test.js
rwwagner90: Gotcha, thanks for the clarification! Can you explain what you mean by `parent`? To my knowledge, when you inspect an object, it should give you the info for that object, not parents and children, right?

app/components/object-inspector/sort-properties.js
patricklx: mmm, i think this is what I had initially too.
But for arrays i wanted to see the computed properties like `lastObject` & `firstObject` at the beginning and not scroll down 100 items...
But this is only in case of arrays and 2 CPs :) . 
So, if you prefer the order similar to eslint-plugin-ember, i go ahead and change it

app/controllers/application.js
rwwagner90: I'm not sure I have ever seen this behavior. Is there something in your app causing a lot of rerenders? Generally, once the page loads, I am able to use the object inspector without any issues.

rwwagner90: My concern here is something could be outdated if we just stop the updates.

app/components/object-inspector/accordion.js
rwwagner90: This comment appears to be a copied/pasted response for the one about the `key`? This is a different issue.

app/templates/components/object-inspector.hbs
patricklx: So,
I select an object to see in the inspector, then dig into a property.
In the object inspector title it will show something like `MainObject \n > property`.
If i then decide that i want to further inspect it in the console, i want to click on `$E` in the title.
But If i then click on  `$E` it sends the MainObject, but expect to get the `property` value in the console.
So, i would need to go back one stack and find the property to send it to console

rwwagner90: @patricklx ahhhh, this is the top level one. I thought this was for the `$E` on the individual properties. This makes sense then 👍

app/components/object-inspector/sort-properties.js
rwwagner90: I think maybe we should follow eslint-plugin-emebr, but you make a good point, so perhaps for arrays we could have a special case that changes the order?

app/controllers/application.js
rwwagner90: What is your app doing that is making it constantly refresh? I think that sounds like an issue with the app, not inspector.

rwwagner90: If we want to not refresh, we should check for deep equality of the properties in the object you are inspecting, if nothing changes, there is no need to refresh, but if it does we need to refresh or you are looking at the wrong info.

ember_debug/view-debug.js
patricklx: mmm, right, this is because i set it to call the function at the end of 500ms. to send it out of band.
but I can  actually do later(() => throttle(this, this.send, 300), 50);
so, we get a fast new render, but if there are too many rerendes it will delay it by 300ms to not kill ember-inspector

rwwagner90: Could we try `throttle(this, this.send, 100, true);`? To allow it to be called immediately, then throttle to 100ms?

app/components/object-inspector/properties-all.hbs
```diff
@@ -2,7 +2,7 @@
   <ObjectInspector::SortProperties
     @properties={{this.flatPropertyList}} as |sortedProperties|
   >
-    {{#each sortedProperties key="name" as |prop|}}
```
rwwagner90: If adding the key increases performance, totally fine to keep it.

tests/ember_debug/object-inspector-test.js
```diff
@@ -296,12 +301,14 @@ module('Ember Debug - Object Inspector', function(hooks) {
 
     inspected.set('name', 'Alex');
 
+    await new Promise(res => setTimeout(res, 400));
```
rwwagner90: Should we try something like `await settled()` here?

app/controllers/application.js
patricklx: it is an issue with the app, but i think we should be able to use ember-inspector during app development with app issues :)

but I just tested without this change, and its working fine now...

tests/ember_debug/object-inspector-test.js
patricklx: There are now 2 types of inheritance, the one from Ember and the one that comes from ES6.
Parent for me is now either a upper level mixin or prototype.
So, when we inspect an ES6 object I initially split it up into mixins, (where mixins are actually the prototypes of the object).
When we define a component, all properties, CPs, tracked are in the prototype of its instance. (We get the actual instance in the inspector, which doesn't have any properties set on its own (by `Object.hasOwnProperty`)). 
But it can also be that someone sets properties in the constructor. in that case it would have properties in the instance.

I merge the instance properties, with the one from its first prototype. So that both info is displayed in `Own Properties`
and more upper level prototypes are displayed as mixins.

ember_debug/view-debug.js
rwwagner90: Sure, we can try 300 then. Perhaps we should build in some kind of assertion here that if the app is going crazy with a ton of renders, that we warn you to look into the app?

tests/ember_debug/object-inspector-test.js
rwwagner90: Gotcha, that makes sense. I think the instance and prototype are all part of the item though. I would not consider it a parent. I suppose perhaps base class might be what you mean?

app/controllers/application.js
rwwagner90: I think we either should remove this check entirely or it should be a check that basically nothing on the instance changed. If all the properties remain the same, skip refreshing, etc. Might be easier to just remove for now.

ember_debug/utils/type-check.js
pzuraq: Newer code should never reach here, above it uses the 3.10+ method for checking and anything that's more recent should hit that first.

NullVoxPopuli:
with https://emberclear.io (running locally in development mode), I get these errors when I click on any of my services:
image

ember_debug/object-inspector.js
```diff
@@ -22,6 +22,18 @@ try {
 
 const keys = Object.keys || Ember.keys;
 
+/**
+ * workaround to support detection of `[object AsyncFunction]` as a function
+ * @param value
+ * @returns {string}
+ */
+function typeOf(value) {
```
rwwagner90: @patricklx what did we need this change for?

patricklx: Ember.typeOf will not detect async function as function and will tell its an object instead.
I could also make a pr against emberjs for that...

ember_debug/object-inspector.js
```diff
@@ -500,7 +500,7 @@ export default EmberObject.extend(PortMixin, {
       if (proto.hasOwnProperty('toString') && !proto.hasOwnProperty('hasOwnProperty')) {
         return proto.toString();
       }
-      if (name === 'Class' || name.startsWith('_')) {
+      if (name === 'Class' || name.startsWith('_') || name.length === 1) {
```
rwwagner90: What is the use case for `name.length === 1`?

ember_debug/object-inspector.js
```diff
@@ -842,6 +845,12 @@ function calculateCPs(object, mixinDetails, errorsForObject, expensiveProperties
             tagInfo.tag = metal.track(() => {
               value = calculateCP(object, item.name, errorsForObject);
             });
+            if (!tagInfo.tag.subtag && !tagInfo.tag.subtags && tagInfo.tag === metal.tagForProperty(object, item.name)) {
```
rwwagner90: I'm not sure what these ifs are for, but it seems like maybe it should be a util function to check for whatever this is checking?

ember_debug/object-inspector.js
patricklx: in prod build the class names are uglyfied to 1 character

ember_debug/object-inspector.js
patricklx: if only a tracked is in a tracking context `metal.track(...)`, then the resulting tag will be the same as the `metal.tagForProperty` and will not have any subtag or subtags.
I think emberjs should provide a way to check if a property on an object (or descriptor) is a tracked. Then we can remove this in the feature

ember_debug/object-inspector.js
rwwagner90: @patricklx wouldn't that not always be the case though? What if you have more classes than the 26 letters in the alphabet? Would it make it 2 characters? @pzuraq @rwjblue is there a more reliable check we can do here?

ember_debug/object-inspector.js
rwwagner90: @chancancode @pzuraq thoughts on this?

ember_debug/object-inspector.js
patricklx: mmm, right, it could happen. but I think it must be more than 26 classes in a module or in same scope?
But I guess we could set it to `length <= 3`? should be 17,576‬ different classes.

rwwagner90: I would like to hear from @pzuraq, @chancancode, or @rwjblue on if there is something less brittle we could do for this.

ember_debug/object-inspector.js
```diff
@@ -391,12 +549,31 @@ export default EmberObject.extend(PortMixin, {
...
+          objectMixins[0].properties = objectMixins[0].properties.filter(o => !props.includes(o.name));
+        });
+      }
+      // the base mixins appear again later in the mixins...
+      mixins.splice(0, 1);
     }
+    mixinDetails.push(...objectMixins.slice(0, index));
+    const objectName = mixinDetails[0].name;
+
     mixins.forEach(mixin => {
```
pzuraq: So, the new strategy appears to be:

1. Walk the entire prototype tree, gathering all information about the object/classes and putting it onto some fake-mixins
2. Put these on top, and then add mixins, and then add `EmberObject` at the very bottom

I think this is a bit problematic, because then the mixins are going to be out of order with the class prototypes they actually belong to. They'll also consist entirely of properties/descriptors that already show up on one of the prototype-mixins (unless they're overridden, which I guess technically they would all be?), e.g.:


```js
class Foo extends Bar.extend(MyMixin) {}

class Bar extends Baz {}

class Baz extends EmberObject {}
```
```
// inspector shows
- Foo 
- Bar
- Baz
- MyMixin
- EmberObject
```

I think ideally what we _should_ do here is show only the classes themselves by default, and then have a sort of subview where you can examine the mixins of a particular class:

```
// inspector shows
- Foo
- Bar
  - mixins: MyMixin
- Baz
- EmberObject
```

I think we could add the subview in a followup, but I'm not sure if we can show the out of order view in the meantime, that seems like it'd be really confusing. Maybe we could try to setup the order the correct way at least?

```
// inspector shows
- Foo
- Bar
- MyMixin
- Baz
- EmberObject
```

tests/ember_debug/object-inspector-test.js
pzuraq: At the least abstract this to be a separate helper I think

pzuraq:
Overall this is looking pretty good, there are just a couple things that need to change I think to be consistent. I should be able to work on this tomorrow morning (I think I have push access to the branch as a maintainer), if that works, would definitely like to get this merged asap 😄

ember_debug/object-inspector.js
```diff
@@ -772,7 +980,15 @@ function customizeProperties(mixinDetails, propertyInfo) {
...
 function getDebugInfo(object) {
   let debugInfo = null;
-  const objectDebugInfo = get(object, '_debugInfo');
+  let objectDebugInfo = get(object, '_debugInfo');
+  if (object instanceof Ember.ObjectProxy) {
+    if (!object.content) {
+      objectDebugInfo = null;
+    }
+    if (object.content) {
+      object = object.content;
```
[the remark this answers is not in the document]
patricklx: mmm, but for ember data it had issues with displaying data.
the `_debugInfo` is accessed with `get`, which passes through ember proxy.
It's also a bit annoying to click 2 times whenever I want to inspect e.g. `belongsTo` data...
Can we make a special case for ember data models?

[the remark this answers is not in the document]
pzuraq: So, loaded up an Ember app to take a look, `emberclear`, and it looks like currently at least models are already special cased. Whenever I click on a model, and then on a relationship within that model, it shows the model itself, not the proxy first. I think keeping that behavior would be fine since it's already established.

FWIW, I don't necessarily think this would be a bad idea either, but it seems out of scope for this PR. I think we should fix the existing behavior for sure, but any additional proxy changes should be followups.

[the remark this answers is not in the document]
patricklx: strange, for me it keeps failing in ember-data `_debugInfo`...  version `3.11.4`, but sure, i can extract this parts into another PR

ember_debug/object-inspector.js
patricklx: but if we only use prototype lookup, it would show to the `toString` from mixins...
And only default class name will not show the `toString` for mixins

ember_debug/object-inspector.js
patricklx: I changed to logic for this part and added a test to check the correct order of prototypes and mixins

ember_debug/object-inspector.js
pzuraq: From what I can see, `toString` from mixins should be applied to the class, and shouldn't belong to the mixin directly. I actually can't find a way to name mixins at all, which is a bit unfortunate, but that seems to be by design 😕

pzuraq:
@patricklx Made a few changes to clean some things up. It's really important that we show an accurate representation of the object's inheritance, so really everything in the prototype chain should accurately only show its own properties. We're definitely lacking a couple of the APIs that would make this ideal, but I found a couple of ways to hack around it for the time being.

I think with these changes this should be ready to merge, assuming tests pass 🎉

patricklx:
mmm, okay.
But now I only see some strange properties in the `Own Properties`, which might be correct, but not useful, since those are just set my ember itself, e.g. `renderer` and parentView`` etc...

Now, I always need to close the `Own Properties` tab and open some weird named tab :) .
This worked with Ember.Object.extend, since it would set CPs and properties on the instance itself, but its not happening with es6 classes anymore, so `Own Properties` is not so useful for debugging/development. Can we have a `Useful Info` tab then.... where we merge Own Properties and the one from the instance class prototype?

And now the names of the my components are always <@ember/component:ember123> instead of showing the Classname, and if I didn't set a class name with `export default class extends`,
it shows me `<_class2:ember123>` 


I cannot dig into any `belongsTo` in the models...
and I cannot get the content for PromiseManyArray, it just gets me the list of internal models when I click on `.content`. Both were fixed by including the proxy content. But I open another PR for that.

patricklx:
@pzuraq I added one more commit to show the first class protoype be default, since it has all the properties

rwwagner90:
@pzuraq @patricklx we should try to preserve how the object inspector worked before. If it merged own properties and prototype properties, perhaps we should leave that?

patricklx:
@rwwagner90 before it was showing own properties and the first class prototype in separately.  It was also expanding both.
I wanted to merge those, since those 2 have all the useful information. Now its back to 2 separated and expanded

rwwagner90:
@patricklx we should keep it how it was before 👍

rwwagner90:
@patricklx I know you've already put a lot of work in on this, but for some of the things you mentioned like the class names being different, not being able to expand belongsTo, proxies, etc it would be 💯 if we added some tests for some of these things, so we can ensure it keeps working.

pzuraq:
> I cannot dig into any belongsTo in the models...

This is not ideal, but this does appear to match the existing behavior. If I visit https://www.emberobserver.com/addons/ember-cli-babel and open the inspector, both before and after the change, relationships within the data store seem to work the same. We should definitely fix this, I just think it needs to be in a separate followup, with a lot of tests to make sure it doesn't get into this state again 😄 

> And now the names of the my components are always <@ember/component:ember123> instead of showing the Classname

Hmm, I was thinking when I rewrote it that this would be solved by us removing classic components and EmberObject as a whole, or updating the `toString()` behavior upstream. I also didn't want to remove the GUID stuff, in case some users were using it. However, it would be very nice to get some progressive enhancement here just by switching to native classes. Let me think about it for a minute.. 

> Can we have a Useful Info tab then.... where we merge Own Properties and the one from the instance class prototype?

FWIW, I do think this would be useful. Like, normal JS objects in the console inspector do show all inherited values, but slightly greyed out. I think we can do something better here, but that's a somewhat orthogonal change to the main issues here, which are tracked support and getters, and better support for native classes.

It's generally a good idea to try to break large changes like these up into smaller PRs, so we can review them more thoroughly. I actually think there will still need to be quite a few refactors (as I noted in the comment I left) so there's lots of opportunity to get a better DX

nummi:
Thank you @patricklx 🎉

rwwagner90:
@patricklx we definitely owe you a 🍺for your help! Thanks so much!

patricklx:
👍, thanks. And sorry that this PR got so big... 😅

PR #1037, opened 2019-09-23
Add initial octane tree support

rwwagner90:


ember_debug/vendor/startup-wrapper.js
```diff
@@ -23,6 +23,11 @@ if (typeof env !== 'undefined') {
 var EMBER_VERSIONS_SUPPORTED = {{EMBER_VERSIONS_SUPPORTED}};
 // @formatter:on
 
+// Enable debug render tree
+
+window.EmberENV = window.EmberENV || {};
+window.EmberENV._DEBUG_RENDER_TREE = true;
```
chancancode: Do you know what is the timing of this script's injection?

ember_debug/libs/octane-tree.js
```diff
@@ -0,0 +1,14 @@
+export function makeRenderNodeCloneable(retainObject, renderNode) {
```
chancancode: nit: the file _maybe_ misnamed, since it would work on 3.14+ "non-Octane" apps too, not sure if it matters

ember_debug/libs/octane-tree.js
```diff
@@ -0,0 +1,14 @@
...
+    // TODO add args and bounds
+    value: {
+      isComponent: renderNode.type === 'component',
+      name: renderNode.name,
+      objectId: renderNode.instance ? retainObject(renderNode.instance) : null,
+      viewClass: renderNode.name
+    },
+    children
+  };
+}
```
chancancode: To filter out the outlets:

```ts
function filterNodes(nodes: CapturedRenderNode[]): CapturedRenderNode[] {
  return nodes.map(
}

function filterNode(node: CapturedRenderNode): CapturedRenderNode {
  if (node.type === 'outlet') {
    assert('outlet node should have exactly one child', node.children.length === 1);
    return filterNode(node.children[0]);
  } else {
    return { ...node, children: filterNodes(node.children) };
  }
}
```

ember_debug/libs/octane-tree.js
rwwagner90: Fair enough. I was just thinking of a file like the current `glimmer-tree` one. What would be a good name for it? I suppose we could put it all in the existing `glimmer-tree` and do the version checks there.

ember_debug/libs/octane-tree.js
rwwagner90: So this would be filtering out after the fact, not inside the current recursive call?

chancancode: You could work it into the existing one too. Just do the logic inside your make RenderNodeClonable function (and probably rename it).

ember_debug/libs/octane-tree.js
chancancode: It may be a fine name as long as we don’t be tempted into thinking we can do anything octane specific (such as assuming the template-only semantics)

ember_debug/libs/octane-tree.js
rwwagner90: @chancancode that's the issue though, is my brain cannot grok how to do it in the existing `.map` for children.

ember_debug/vendor/startup-wrapper.js
rwwagner90: I do not

chancancode: Hm we will have to figure that out. It will not work correctly if it's set after Ember is loaded.

ember_debug/libs/octane-tree.js
chancancode: I would think this should work?

```
export function whateverYouCallThisNow(retainObject, renderNode) {
  if (node.type === 'outlet') {
    assert('outlet node should have exactly one child', node.children.length === 1);
    return whateverYouCallThisNow(retainObject, node.children[0]);
  } else {
    let children = renderNode.children.map(child => whateverYouCallThisNow(retainObject, child));

    return {
      // TODO add args and bounds
      value: {
        isComponent: renderNode.type === 'component',
        name: renderNode.name,
        objectId: renderNode.instance ? retainObject(renderNode.instance) : null,
        viewClass: renderNode.name
      },
      children
    };
  }
}
```

ember_debug/vendor/startup-wrapper.js
rwwagner90: We have a function under this that loads on Ember load, so I am guessing if we define stuff before it, it should get set before Ember is done, but not necessarily before Ember starts. Is there a way to log when Ember starts loading?

rwjblue:
@rwwagner90 - Should this be closed in favor of #1088?

rwwagner90:
@rwjblue yeah, now that I have ported over the tree args, there is no need for this PR.

PR #1038, opened 2019-09-23
Tidy up Promise tab

nummi:
`@, this, </>, action`,etc.

### DIFF
--- a/app/templates/components/promise-item.hbs
+++ b/app/templates/components/promise-item.hbs
@@ -1,8 +1,8 @@
-{{#@list.cell
-  class=(concat "list__cell_main " this.expandedClass)
-  style=this.labelStyle
-  on-click=(action @toggleExpand @model)
-}}
+<@list.cell
+  class={{concat "list__cell_main " this.expandedClass}}
+  style={{this.labelStyle}}
+  @on-click={{action @toggleExpand @model}}
+>
   <div class="list__cell-partial list__cell-partial_size_medium">
     <span
       title={{this.label}}
@@ -23,18 +23,18 @@
       </button>
     {{/if}}
   </div>
-{{/@list.cell}}
+</@list.cell>
 
-{{#@list.cell}}
+<@list.cell>
   <div
     class="pill pill--bold pill--text js-promise-state"
     style={{this.style}}
   >
     {{this.state}}
   </div>
-{{/@list.cell}}
+</@list.cell>
 
-{{#@list.cell class="js-promise-value"}}
+<@list.cell class="js-promise-value">
   {{#if this.hasValue}}
     <div
       class="list__cell-partial list__cell-partial_size_medium"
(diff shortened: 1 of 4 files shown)


app/templates/promise-tree-toolbar.hbs
```diff
@@ -23,26 +23,26 @@
   <Ui::ToolbarDivider />
 
   <button
-    class="{{if noFilter "active"}} toolbar__radio toolbar__radio--first js-filter"
-    {{action "setFilter" "all"}}
+    class="{{if (eq this.filter "all") "active"}} toolbar__radio toolbar__radio--first js-filter"
```
rwwagner90: 🔥

app/templates/components/promise-item.hbs
```diff
@@ -1,76 +1,76 @@
-{{#list.cell
-  class=(concat "list__cell_main " expandedClass)
-  style=labelStyle
-  on-click=(action toggleExpand model)
+{{#@list.cell
```
rwwagner90: Was this `@` on purpose? I suppose it is valid, but looks odd having `#@`

app/templates/promise-tree.hbs
```diff
@@ -9,18 +9,18 @@
...
-          tracePromise=(action "tracePromise")
-          inspectObject=(action "inspectObject")
-          sendValueToConsole=(action "sendValueToConsole")
-          list=list
-        }}
+        <PromiseItem
+          @model={{content}}
+          @filter={{this.filter}}
+          @effectiveSearch={{this.effectiveSearch}}
+          @toggleExpand={{fn this.toggleExpand content}}
```
rwwagner90: I'm curious why only two of the actions needed `content` passed?

app/templates/components/promise-item.hbs
nummi: It looks terrible :-P. But `list` is passed in that way, right? Maybe we should get a consult from an expert?

app/templates/promise-tree.hbs
nummi: I like to keep the components as dumb as possible but maybe consistency should win here.

app/templates/components/promise-item.hbs
rwwagner90: Does angle bracket syntax work here?

app/templates/promise-tree.hbs
rwwagner90: If only two need it, that's fine. I just wanted some clarity.

app/templates/components/promise-item.hbs
nummi: Yep :+1:

rwwagner90: Nice! That looks much better 😃

app/templates/promise-tree.hbs
```diff
@@ -9,18 +9,18 @@
...
-          model=content
-          filter=filter
-          effectiveSearch=effectiveSearch
-          toggleExpand=(action "toggleExpand")
-          tracePromise=(action "tracePromise")
-          inspectObject=(action "inspectObject")
-          sendValueToConsole=(action "sendValueToConsole")
-          list=list
-        }}
+        <PromiseItem
```
stefanpenner: Fancy!

nummi: Thanks. 😆

nummi:
code climate 🤷‍♂

PR #1044, opened 2019-09-29
Misc Component Cleanup

nummi:
## MainContent / MonitorContentHeight

See commits: https://github.com/emberjs/ember-inspector/pull/1044/commits/df45d75145abf5db5c08211d16d6b63835c97995 and https://github.com/emberjs/ember-inspector/pull/1044/commits/acf985e67b78a7127e72c968d4ea558ba4ebb8e2

The new component name describes what it actually does.

## X-App

See commit https://github.com/emberjs/ember-inspector/pull/1044/commits/eb52384b5e055bfe6664539aafdf93d2700dad3d

This component existed to apply two classes: `is-dragging` and `inactive`. This is now handled in the application template.

## Drag Handle

See commit https://github.com/emberjs/ember-inspector/pull/1044/commits/504e2fdbf32b10ee3dc7063df24fee221aec56d4 for changed code.

We were doing a lot of passing around the Application controller and `setIsDragging` action just to keep the column drag cursor from being glitchy. I removed all of that and just expand the width of the drag column so the hover style (column cursor) stays in place.

(The column cursor is applied on hover of the drag handle. If you drag your cursor quickly to the left or right you are briefly not hovering until the UI catches up — this displays the default cursor. To fix this we were tracking `isDragging` at the application level, applying a class to the root to keep the column cursor. There was a lot of plumbing involved and I replaced it simply by increasing the width of the hover area on `mousedown`.)

Default width of column:

default

On `mousedown`:

mousedown

It's a little difficult to see in an animated GIF but here is the glitch without the CSS fix:

glitch


rwwagner90:
This is 🔥 . Love removing all the controller injections!

PR #1045, opened 2019-10-02
<LinkTo>

nummi:


app/templates/nav.hbs
```diff
@@ -1,63 +0,0 @@
-<nav class="nav nav--main">
```
rwwagner90: Was this file just a duplicate? Never used?

rwjblue: Looks like it was left over from https://github.com/emberjs/ember-inspector/commit/412dc4bc544d2cbba8850ef0ca33684e1decfd3e. Before that commit we had `{{partial "nav"}}` which rendered this file, but I think after that commit this file was unused completely.

nummi: Correct. :+1:

PR #1051, opened 2019-10-14
Performance improvements

patricklx:
* Dont scroll when tree is refreshes and item is already selected
* Dont close opened object inspector groups on refresh
* Don't redraw same items (specific key in each )
* fix memory issue on profiles tab, limit to first 100 profiles
* Limit large objects or arrays to 100 properties

after tracked branch PR...
https://github.com/patricklx/ember-inspector/compare/support-refresh.same...patricklx:perf-improvements

### DIFF
--- a/app/components/object-inspector/properties-base.js
+++ b/app/components/object-inspector/properties-base.js
@@ -5,6 +5,12 @@ export default Component.extend({
   tagName: '',
 
   sendToConsole: action(function ({ name }) {
+    if (name === '...') {
+      this.port.send('objectInspector:sendToConsole', {
+        objectId: this.model.objectId
+      });
+      return;
+    }
     this.port.send('objectInspector:sendToConsole', {
       objectId: this.model.objectId,
       property: name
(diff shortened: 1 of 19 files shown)


NullVoxPopuli:
Services load things: 
image

image
I don't know what that e is down there.

`unknown` appears to be ember object things? get/set etc.


image
models load!

Component tree does not load. IS that a different PR?

patricklx:
@NullVoxPopuli fix for components is in master, just merged it into this PR as well

app/components/object-inspector/properties-base.js
```diff
@@ -5,6 +5,12 @@ export default Component.extend({
   tagName: '',
 
   sendToConsole: action(function ({ name }) {
+    if (name === '...') {
```
rwwagner90: What is the idea behind this `if`? Is it just to avoid sending the `property: name` from the code below? If so, perhaps we could conditionally add the `property` prop instead of duplicating the `this.port.send('objectInspector:sendToConsole',` call?

rwwagner90:
@patricklx would it be possible to add some tests for this functionality? Also, we have a build error.

PR #1052, opened 2019-10-17
Fix Ember.typeof replace with custom typeof

patricklx:
mainly to support detecting async function.
`type-instance` and `type-class` will be gone. (class probably never happend)


patricklx:
You already have:
https://github.com/emberjs/ember-inspector/blob/4f3822d34559a1bbe3cd1f3ef5a1534b98e644b6/tests/ember_debug/object-inspector-test.js#L96

Though i could add one more for async function

patricklx:
mmm, I would add the test for async func after I rebase once/if this is merged. since currently functions are ignored by the object-inspector

app/components/object-inspector/property.js
```diff
@@ -28,11 +28,11 @@ export default Component.extend({
 
   isObject: equal('valueType', 'type-object'),
 
-  isInstance: equal('valueType', 'type-instance'),
-
   isComputedProperty: alias('model.value.computed'),
 
-  isFunction: equal('valueType', 'type-function'),
+  isFunction: computed('valueType', function () {
+    return this.get('valueType') === 'type-function' || this.get('valueType') === 'type-asyncfunction';
```
btecu: Is the `.get` necessary?

ember_debug/adapters/web-extension.js
```diff
@@ -1,7 +1,8 @@
 import BasicAdapter from './basic';
+import { typeOf } from "../utils/type-check";
```
btecu: Single quotes

ember_debug/libs/glimmer-tree.js
```diff
@@ -24,15 +24,17 @@
  *
  * @class GlimmerTree
  */
+
 const Ember = window.Ember;
+import { typeOf } from "../utils/type-check";
```
btecu: Single quotes

app/components/object-inspector/property.js
patricklx: does not look like it. removed

app/components/object-inspector/property.js
```diff
@@ -28,11 +28,11 @@ export default Component.extend({
 
   isObject: equal('valueType', 'type-object'),
 
-  isInstance: equal('valueType', 'type-instance'),
```
rwwagner90: Is there a reason we are removing `instance`?

ember_debug/utils/type-check.js
```diff
@@ -41,3 +41,8 @@ export function getDescriptorFor(object, key) {
 
   return object[key];
 }
+
+
+export function typeOf(obj) {
+  return ({}).toString.call(obj).match(/\s([a-zA-Z]+)/)[1].toLowerCase();
```
rwwagner90: Won't this be brittle, since `toString` can be overridden?

app/components/object-inspector/property.js
patricklx: instance was specific to Ember.typeOf, its now also just an `object`. the only place where it was used is in the `canDig` function

rwwagner90: Okay, I was worried if we didn't have the instance, we would not be able to click further into the instances. You're saying it works still with this change?

patricklx: right, it still works

ember_debug/object-inspector.js
```diff
@@ -1,10 +1,11 @@
 import PortMixin from 'ember-debug/mixins/port-mixin';
 import { compareVersion } from 'ember-debug/utils/version';
 import { isComputed, isDescriptor, getDescriptorFor } from 'ember-debug/utils/type-check';
+import { typeOf } from "./utils/type-check";
```
btecu: Single quotes. First time I didn't want to reference all the instances but there's at least another one more that needs to be updated.

patricklx: :(  I though i catched them all...

rwwagner90:
@patricklx will this work back to Ember 3.4 without the `isInstance` check? We have to support 3.4 to latest.

patricklx:
> @patricklx will this work back to Ember 3.4 without the `isInstance` check? We have to support 3.4 to latest.

yes, since `type-instace` only came from Ember.typeOf, the new typeof just recognizes it as object.
the only place it was used, was in the canDig check. which also checks if something is an object

PR #1053, opened 2019-10-24
Improve proxies

patricklx:


ember_debug/object-inspector.js
```diff
@@ -528,6 +528,13 @@ export default EmberObject.extend(PortMixin, {
    */
   mixinDetailsForObject(object) {
     const mixins = [];
+    if (object instanceof Ember.ObjectProxy && object.content && !object._showProxyDetails) {
+      return this.mixinDetailsForObject(object.content);
+    }
+
+    if (object instanceof Ember.ArrayProxy && object.content && !object._showProxyDetails) {
+      return this.mixinDetailsForObject(object.slice(0, 101));
```
rwwagner90: What is this `slice` for?

ember_debug/object-inspector.js
```diff
@@ -1037,6 +1044,9 @@ function getDebugInfo(object) {
   let debugInfo = null;
   let objectDebugInfo = get(object, '_debugInfo');
   if (objectDebugInfo && typeof objectDebugInfo === 'function') {
+    if (object instanceof Ember.ObjectProxy && object.content) {
+      object = object.content;
```
rwwagner90: I wonder if maybe we should use `Ember.get`, wouldn't that resolve proxy values as well as normal objects?

ember_debug/object-inspector.js
patricklx: it does the same as `toArray()`, so it obtains the correct items from e.g. ember data many-array but only for the first 101 items, (though this should actually be part of the perf PR). otherwise it would call `objectAt` for all items in many-array

ember_debug/object-inspector.js
patricklx: but the new ember data debugInfo  does not do it...
always fails here:
https://github.com/emberjs/data/blob/master/packages/model/addon/-private/model.js#L1186

rwwagner90: @patricklx can you elaborate on that? Perhaps @runspired can help?

runspired: > but the new ember data debugInfo 

I wouldn't call this "new", its been around a long time and it's something we'd love to delete :P 

Reaching through the proxy for the content seems wise here.

patricklx: mmm, right...
but I think it did not happen before...  maybe because im using class extends Model {...} instead of Model.extend(...)

patricklx: no, same issue, so might have never worked with belongTo relations :)

rwwagner90:
@patricklx I fixed the conflicts, but seems like we have some test failures now.

PR #1055, opened 2019-10-28
Name Known Ember Mixins

patricklx:
set the names of known mixins
remove properties set by mixins unless different value
causes them to hide the class with only merged mixins from Object.extend(mix1, mix2)

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -17,11 +17,11 @@ let HAS_GLIMMER_TRACKING = false;
 try {
   glimmer = Ember.__loader.require('@glimmer/reference');
   metal = Ember.__loader.require('@ember/-internals/metal');
-  HAS_GLIMMER_TRACKING = glimmer && 
-                         glimmer.value && 
-                         glimmer.validate && 
-                         metal && 
-                         metal.track && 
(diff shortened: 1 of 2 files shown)


ember_debug/object-inspector.js
```diff
@@ -22,6 +22,39 @@ try {
 
 const keys = Object.keys || Ember.keys;
 
+/**
+ * Add Known Ember Mixins and Classes so we can label them correctly in the inspector
+ */
+const emberNames = new Map();
```
efx: NOTE: this is a nitpick but you can instantiate a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) directly from an array. I find it clearer:
```js
const emberNames = new Map([
  [Ember.Evented, 'Evented Mixin']
]);
```

ember_debug/object-inspector.js
```diff
@@ -30,6 +30,51 @@ try {
...
+  // do nothing
+}
+
+
+/**
+ * workaround to support detection of `[object AsyncFunction]` as a function
+ * @param value
+ * @returns {string}
+ */
+function typeOf(value) {
```
rwwagner90: @patricklx shouldn't this not be here, since we merged the `typeOf` stuff?

PR #1058, opened 2019-11-05
More dependencies upgrade

chancancode:


lib/ui/package.json
```diff
@@ -4,7 +4,7 @@
     "ember-addon"
   ],
   "dependencies": {
-    "ember-cli-babel": "^7.1.2",
```
rwwagner90: Do we even need this `package.json`? If we omit it, would it use the deps from the root `package.json`?

chancancode: We need it for the `ember-addon` stuff but I switched the deps to `*` so it just uses whatever is at the root

package.json
```diff
@@ -96,6 +96,7 @@
     "js-string-escape": "^1.0.0",
     "loader.js": "^4.7.0",
     "mkdirp": "^0.5.1",
+    "pretender": "^3.0.4",
```
rwwagner90: Do we need to switch any imports with the switch from ember-cli-pretender to pretender?

rwjblue: No, ember-cli-pretender made `import Pretender from 'pretender'` work

chancancode: We are already importing from pretender: https://github.com/emberjs/ember-inspector/blob/master/tests/acceptance/whats-new-test.js#L4

PR #1060, opened 2019-11-07
Refactor `ComponentViewItem` to use native class

chancancode:


app/controllers/component-tree.js
```diff
@@ -137,7 +136,7 @@ export default Controller.extend({
     viewArray.forEach(viewItem => {
       let cachedExpansion = expandedStateCache[getIdFromObj(viewItem)];
       if (cachedExpansion !== undefined) {
-        viewItem.set('expanded', cachedExpansion);
```
rwwagner90: We not longer need `set` here?

app/controllers/component-tree.js
```diff
@@ -210,7 +209,7 @@ export default Controller.extend({
     this.filteredArray.forEach((item) => {
       const id = getIdFromObj(item);
       if (objects.includes(id)) {
-        item.set('expanded', state);
```
rwwagner90: Or here?

app/controllers/component-tree.js
nummi: The class is using @tracked.

rwwagner90:
@chancancode this seems fine to me, we just need to fix the tests.

chancancode:
Yeah! I looked into it and I know why it's not working. It needs a bigger refactor, I'll get back to this next week.

chancancode:
I’ll resubmit this as part of a bigger refactor to use the new debug render tree api. I was hoping that this can be separated out relatively cleanly, but it needed more work that start to eat into the other refactor so probably doesn’t make sense to do it in isolation anymore.

PR #1061, opened 2019-11-08
Remove unused code

chancancode:
The inspector used to support filtering out components in the view tab, and there are a lot of supporting code for that. Since the view tab was removed, there is no longer any way to set this option, and it is always true in practice.


nummi:
image

🎉

PR #1069, opened 2019-11-09
Migrate CI to GitHub Actions

chancancode:
This is the first step of having CI automatically publish the extensions to the chrome/firefox stores when pushing a release tag. The first pull request ports the Travis config to the GitHub format pretty much as-is, but we will be able to build on top of this pipeline to add the other publishing rather easily.


.github/workflows/build.yml
```diff
@@ -0,0 +1,141 @@
...
+      - name: Set up Volta
+        uses: rwjblue/setup-volta@v1
+      - name: Install dependencies (yarn)
+        run: yarn install
+      - name: Setup ember-try scenario
+        run: yarn ember try:one ember-${{ matrix.scenario }} --skip-cleanup --- cat package.json
+      - name: Build
+        run: yarn ember build --environment test
+      - name: Run test
+        # Due to a bug in ember-cli, running `ember test` with `--path` doesn't set `EMBER_ENV=test`
```
rwjblue: Thanks for helping track this down, would you mind filing an issue?

package.json
```diff
@@ -110,5 +110,9 @@
     "paths": [
       "lib/ui"
     ]
+  },
+  "volta": {
+    "node": "8.16.2",
```
rwjblue: Should probably use 12? AFAIK, there is no consumer facing impact to changing, right?

PR #1072, opened 2019-11-09
Fix glimmer.value

### LINKED ISSUE #1068: Using Inspector (3.12.2) to view a particular Component or Data Model record fails with the following error:
Ember                 : 3.11.1
Ember Data        : 3.11.4
Inspector            : 3.12.2
Chrome Version 80.0.3963.0 (Official Build) canary (64-bit)

Inspecting a Component, Data Model Record or Container results in:

Ember Inspector has errored.
This is likely a bug in the inspector itself.
You can report bugs at https://github.com/emberjs/ember-inspector.
Error message: glimmer.value is not a function
Stack trace: TypeError: glimmer.value is not a function
    at <anonymous>:3551:42
    at Array.forEach (<anonymous>)
    at <anonymous>:3531:24
    at Array.forEach (<anonymous>)
    at calculateCPs (<anonymous>:3530:18)
    at Class.mixinsForObject (<anonymous>:3254:7)
    at Class.sendObject (<anonymous>:3068:26)
    at Class.inspectById (<anonymous>:2990:16)
    at sendEvent (http://localhost:4200/assets/vendor.js:30041:14)
    at Class.trigger (http://localhost:4200/assets/vendor.js:44727:28)

patricklx:
Fixes #1068

Older ember version did not have glimmer.value...
Any other way to do this?

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -15,11 +15,14 @@ let metal;
 try {
   glimmer = Ember.__loader.require('@glimmer/reference');
   metal = Ember.__loader.require('@ember/-internals/metal');
-  if (!glimmer || !glimmer.value || !glimmer.validate) {
+  if (!glimmer || !glimmer.validate) {
     glimmer = null;
-  }
-  if (!metal || !metal.track) {
+  } 
+  if (!metal || !metal.track || !metal.tagForProperty) {
     metal = null;
+  }
+  if (glimmer && !glimmer.value) {
+    glimmer.value = (obj, name) => const tag = metal.tagForProperty(obj, name), tag && tag.value && tag.value();
   }
 } catch (e) {
   glimmer = null;
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -12,18 +12,16 @@ const { oneWay } = computed;
 
 let glimmer;
 let metal;
+let HAS_GLIMMER_TRACKING = false;
 try {
   glimmer = Ember.__loader.require('@glimmer/reference');
   metal = Ember.__loader.require('@ember/-internals/metal');
-  if (!glimmer || !glimmer.validate) {
-    glimmer = null;
-  } 
-  if (!metal || !metal.track || !metal.tagForProperty) {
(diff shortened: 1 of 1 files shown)


rwwagner90:
@patricklx @pzuraq if we know the specific Ember versions that need each check here, we could instead of checking all the things, wrap checks in an Ember version check.

patricklx:
right, but I think it would be better to find a way to get the current tag revision without using glimmer.value()
Tracked has been introduced in ember 3.9?

pzuraq:
@patricklx no there is not, these are the base APIs. The public APIs do not currently expose them, though that may change in the future.

patricklx:
Mmm, but how does emberjs does it to detect changes?

patricklx:
I found this:
https://github.com/emberjs/ember.js/blob/v3.11.1/packages/%40ember/-internals/metal/lib/computed.ts

So tag.value is used

rwwagner90:
@pzuraq would it be possible to get a list of Ember versions that support each API, so we can do a version check, rather than checking for all the APIs?

pzuraq:
@patricklx right, but tags were not really used the same way before 3.13. My point being, the entire autotracking infrastructure and method of detecting changes does not exist at all before 3.13, so we should be using the old methods (observers, etc) for then.

@rwwagner90 `track`, `value`, and `validate` exist in 3.13+ only. Tags existed before 3.13, but we can't really reliably use them in the same way.

pzuraq:
I believe we discussed cutting a new version of Ember Inspector and dropping support for the older methods pre-3.13, this is probably good time to do that

pzuraq:
Hmm, reading through the code a bit more again, I think we can continue to support older versions for a bit longer. Remembering how we do everything.

@patricklx we should check for _all_ of the Glimmer APIs at the beginning of the module, and not attempt to polyfill them. If they don't all exist, we should fallback to the previous methods:

```js
const HAS_GLIMMER_TRACKING = 
  glimmer && 
  glimmer.value && 
  glimmer.validate && 
  metal && 
  metal.track && 
  metal.tagForProperty;
```

PR #1075, opened 2019-11-12
Build with Node 12

chancancode:


rwjblue:
can you update the package.json volta key too?

PR #1080, opened 2019-11-13
Auto publish

chancancode:
* [x] Merge #1077 
* [x] Merge #1078 
* [x] Rebase this
* [ ] Merge this
* [ ] Backport to stable branch
* [ ] 🤞 Try publishing `v3.12.3`?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,8 +103,43 @@ jobs:
         run: node scripts/download-panes.js
         env:
           EMBER_ENV: production
+      - name: Update manifest.json
+        if: github.event_name == 'schedule'
+        run: |
+          DATE="$(date +"%Y.%m.%d")"
+          SHA="${GITHUB_SHA:0:7}"
+          VERSION="$(jq -r '.version' package.json)"
+          CONTENT="$(cat "$FILE")"
+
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -106,8 +103,43 @@ jobs:
...
+          echo "$CONTENT" | jq --arg date "$DATE" --arg sha "$SHA" "$FILTER" > $FILE
+          cat "$FILE"
+        env:
+          FILE: package.json
+          FILTER: .version += "-alpha.\($date)+\($sha)"
       - name: Build
         run: yarn ember build --environment production
+      - name: Pack
+        run: |
+          VERSION="$(jq -r '.version | split("+") | .[0]' package.json)"
```
rwwagner90: Is this taking the version from package.json? If so, would this be the wrong version since it's always one minor version higher?

kategengler: This seems correct. If it's a nightly, the package.json version has already been amended in the previous step with "-alpha.\($date)+\($sha)", if it is a tag, so long as the directions to not bump the version in package.json were followed, the version should be correct.

chancancode: @rwwagner90 yeah, it will be the version on master but only for nightly builds and as @kategengler said it will have the alpha and date stamp on it, so today’s would be v3.13.0-alpha.2019.11.14+abcdef. Anyway; this particular step just have to match what `yarn pack` decides to name the gzip file, which is taken from package.json (but excluding the +...).

PR #1086, opened 2019-11-17
Fix more items indicator sometimes not appearing

patricklx:
Also, put the length property at the top

### DIFF
--- a/app/components/object-inspector/sort-properties.js
+++ b/app/components/object-inspector/sort-properties.js
@@ -1,13 +1,13 @@
 import { map, sort } from '@ember/object/computed';
 import Component from '@glimmer/component';
-import { get, set, computed } from '@ember/object';
+import { set, computed } from '@ember/object';
 import { A } from '@ember/array';
 
 export default class SortProperties extends Component {
   @computed('args.properties')
   get isArray() {
     const props = A(this.args.properties || []);
-    return props.findBy('name', 'length') && props.findBy('name', 0);
+    return props.findBy('name', 'length') && props.findBy('name', '0');
   }
 
   /**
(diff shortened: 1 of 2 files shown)


app/components/object-inspector/sort-properties.js
```diff
@@ -6,10 +6,10 @@ import { A } from '@ember/array';
 export default Component.extend({
   tagName: '',
 
-  isArray: computed('properties', function () {
```
rwwagner90: Should we leave this computed and switch it to `'properties.[]` to ensure it updates?

app/components/object-inspector/sort-properties.js
```diff
@@ -19,14 +19,20 @@ export default Component.extend({
    */
   sortedProperties: computed('sorted.length', function () {
     // limit arrays
-    if (this.get('isArray') && this.get('sorted.length') > 100) {
+    let props = A(this.get('sorted'));
+    if (this.isArray()) {
+      const item = props.findBy('name', 'length');
```
rwwagner90: What is this doing?

rwwagner90:
@patricklx any updates here? Would like to get this in if it's still needed.

app/components/object-inspector/sort-properties.js
patricklx: In this case I don't think it matters since it's only called once.

app/components/object-inspector/sort-properties.js
patricklx: This is to put the length property at the top, since otherwise it will be cutoff. Or it will look wierd at the end

rwwagner90:
@patricklx do we still need these changes? If so, could you please rebase?

rwwagner90:
@patricklx do we still need this PR?

patricklx:
Yes, it's still needed

rwwagner90 (approved):

PR #1087, opened 2019-11-18
Fix tracked detection

patricklx:
call getter directly
Ember.get auto tracks properties is seems, breaking the tracked detection.
So, it marks Getters as tracked, if the getter does not depend on any tracked property,

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -244,7 +244,7 @@ export default EmberObject.extend(PortMixin, {
               }
               tracked[item.name] = tagInfo;
             } else {
-              value = calculateCP(object, item.name, {});
+              value = calculateCP(object, item, {});
               if (values[item.name] !== value) {
                 changed = true;
                 values[item.name] = value;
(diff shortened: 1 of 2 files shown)


rwwagner90:
@pzuraq does this seem logical to you? I don't know about the internals.

rwwagner90:
@patricklx if this fixes a bug in tracked detection, would it be possible to add a test covering that bug, that would fail before and passes now?

pzuraq:
`get` should never mark a getter as anything. It will install a mandatory getter/setter on plain properties, but it does nothing unusual to plain getters. Can you describe the failure you were seeing in a bit more detail?

patricklx:
its not that `get` marks it, but it adds the property tag into the tracking tags.
But we try to detect a tracked by checking if `metal.track` returns the same which metal.tagForProperty returns.
Which happens when accessing getters through `get`.  https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/metal/lib/property_get.ts#L110

pzuraq:
Ah, ok, makes sense. I think this is fine for now 👍

PR #1088, opened 2019-11-20
Revamp components inspection (Octane support and improved UX)

chancancode:
## Known Issues

* [x] Tests are failing
* [x] Should not show internal `-top-level` route template in "Components" tab
* [x] Right click on the page and selecting "Inspect Ember Component" shows the tooltip but does not select the corresponding row in the "Components" tab
  * [x] Scroll to the appropriate row

## Instructions for testing

1. Click on the [Checks](https://github.com/emberjs/ember-inspector/pull/1088/checks) tab for this pull request

2. Under "Build and Publish" > "Build extensions" > "Artifacts", download the "chrome" build
    Screen Shot 2019-11-22 at 2 39 26 PM

3. Extract the zip file you downloaded from the previous step, it should result in a folder called "chrome"

4. Go to [chrome://extensions/](chrome://extensions/), enable "Developer mode" on the top-right corner, disable the Ember Inspector extension using the toggle (if you have it installed), click "Load unpacked", select the folder you extracted from the previous step
    Screen Shot 2019-11-22 at 2 41 08 PM

5. Try out the new inspector
    inspect-5

6. When you are done testing, go to [chrome://extensions/](chrome://extensions/), remove the custom build and switch back to the normal inspector

The custom extension installed using this method does not automatically update itself, so you will have to repeat these steps whenever there is a new commit available for testing.


patricklx:
I just tried this out 👍 
one issue i saw
* when a component updates the view (e.g checkbox component), the selected component (any, does not need to be the same component) in the inspector will be unselected... even though the instance (ember id) is the same

chancancode:
@patricklx I just pushed a refactor with various improvements, can you test it again and see if it is still a problem?

joukevandermaas:
For other people who wanted to try it in Firefox instead of Chrome, follow these steps:

1. Click on the Checks tab for this pull request
2. Under "Build and Publish" > "Build extensions" > "Artifacts", download the "firefox" build.
3. Extract the zip file you downloaded from the previous step, it should result in a folder called "firefox"
3. Open the addons manager in Firefox (Ctrl+Shift+A on Windows, not sure about other OS)
4. Click the little cog icon and select "Debug addons"
5. In the new page that opens, click "Load temporary addon"
6. Navigate to the directory you extracted in step 3, and select the "manifest.json" file
7. When you are done testing, you can click the "remove" button on the temporary addon.

nummi:
Personal project running Ember Source 3.15.0 — I'm seeing a lot of this:

image

Three errors and `[object object]` for the component name.

```
Object Inspector error for bounds Error: You attempted to access the 'bounds' property on a Glimmer Component, but that property does not exist in Ember.js applications, it only exists in Glimmer.js apps
```

chancancode:
The last commit should fix the Glimmer Components debug assertion issue!

app/controllers/component-tree.js
```diff
@@ -1,314 +1,358 @@
...
-    if (!selectedItemIndex) {
+
+    return renderItems;
+  }
+
+  get visibleItems() {
+    return this.matchingItems.filter(item => item.isVisible);
+  }
+
+  @tracked _pinned = undefined;
```
rwjblue: Not a giant issue, but I would have expected all of the tracked properties to be at the beginning of the class (essentially together).

app/controllers/component-tree.js
```diff
@@ -1,314 +1,358 @@
...
-      item.toggleProperty('expanded');
-      this.expandedStateCache[getIdFromObj(item)] = item.get('expanded');
+      this.expand(event.altKey);
     }
-  }),
+  }
 
-  viewInElementsPanel: action(function(item, event) {
+  @action scrollIntoView() {
     event.stopPropagation();
```
rwjblue: Hmm, where does `event` come from? It used to be passed as an action argument is that still happening? I don’t see where `event` is defined at all (why isn’t linting failing?).

app/controllers/component-tree.js
```diff
@@ -1,314 +1,358 @@
...
 
-    const objectId = item.get('view.objectId');
-    const elementId = item.get('view.elementId');
+    this.send('view:scrollIntoView', { id: this.id });
+  }
+
+  @action inspectElement() {
+    event.stopPropagation();
```
rwjblue: Same here RE: `event`

ember_debug/libs/view-inspection.js
```diff
@@ -0,0 +1,656 @@
...
+    #${prefix}-highlight-${id} {
+      display: none;
+      pointer-events: none;
+      box-sizing: border-box;
+      position: absolute;
+      margin: 0px;
+      padding: 0px;
+      border: none;
+      z-index: 10000;
+      /* https://github.com/ChromeDevTools/devtools-frontend/blob/b336f0440a8fb539352ac223ef466c3475618cf1/front_end/common/Color.js#L904 */
```
rwjblue: These inline comments are great, thank you!

skeletons/web-extension/boot.js
```diff
@@ -0,0 +1,16 @@
+// This script runs at document_start, avoid adding more stuff here!
+
+// TODO: make this conditional, only do it when requested.
+
+let id = `ember-inspector-boot-${(Math.random() * 100000000).toFixed(0)}`;
+
+let script = document.createElement('script');
```
rwjblue: I'm curious if these `let`'s pollute "global space" for the consumer, have you tested?

tests/ember_debug/view-debug-test.js
```diff
@@ -1,44 +1,93 @@
...
+
 function setTemplate(name, factory) {
   if (typeof factory.meta === 'object') {
-    factory.meta.moduleName = `my-app/${name}.hbs`;
+    factory.meta.moduleName = `my-app/templates/${name}.hbs`;
   } else if (typeof factory.__meta === 'object') {
     // Ember 3.13+
-    factory.__meta.moduleName = `my-app/${name}.hbs`;
+    factory.__meta.moduleName = `my-app/templates/${name}.hbs`;
   }
```
rwjblue: It isn't immediately obvious to me why we actually care about setting `meta.moduleName`, what is it used for in the UI?

tests/ember_debug/view-debug-test.js
```diff
@@ -118,183 +316,167 @@ module('Ember Debug - View', function(hooks) {
...
-    let previewDiv = find('[data-label=preview-div]');
+    matchTree(tree, [TopLevel(
+      Route({ name: 'application' },
+        Route({ name: 'simple' },
+          Component({ name: 'test-foo', bounds: 'single' }),
+          Component({ name: 'test-bar', bounds: 'range' })
+        )
+      )
+    )]);
+  });
```
rwjblue: This abstraction is really nice :clap:

tests/ember_debug/view-debug-test.js
chancancode: We use it to show the template name/path in the tooltips

skeletons/web-extension/boot.js
chancancode: It does not, it's in its own special scope (which is why I can't just set `window.EmberENV = { _DEBUG_RENDER_TREE: true };` here in the first place)

PR #1093, opened 2019-12-11
Add args to tree

rwwagner90:
Screen Shot 2019-12-12 at 2 15 13 PM
Screen Shot 2019-12-12 at 2 15 27 PM


app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +21,26 @@
...
+    class="component-tree-item__tag 
+    {{if (and (not @item.args.positional) @item.isComponent) "component-tree-item__bracket"}} 
+    {{if (and (not @item.args.positional) (not @item.hasChildren)) "component-tree-item__bracket--self-closing"}}
+    {{if (and @item.args.positional @item.isComponent) "component-tree-item-classic__bracket"}} 
+    {{if (and @item.args.positional (not @item.hasChildren)) "component-tree-item-classic__bracket--self-closing"}}"
   >
     {{#if @item.isComponent}}
-      {{classify @item.name}}
+      {{#if @item.args.positional}}
+        {{dasherize @item.name}}
```
chancancode: I think we don't need to do anything special here, since the names already came dasherized

app/helpers/component-argument.js
```diff
@@ -0,0 +1,25 @@
...
+  } else if (argument === undefined) {
+    return 'undefined';
+  } else if(isArray(argument) || typeof argument === 'object'){
+    return '...';
+  }
+
+  return argument;
+}
+
+export default helper(componentArgumentDisplay);
```
chancancode: trailing new line

app/styles/component_tree.scss
```diff
@@ -77,6 +77,20 @@
...
+  content: '{{';
+}
+
+.component-tree-item-classic__bracket:after {
+  color: var(--base07);
+  content: '}}';
+}
+
+.component-tree-item-classic__bracket.component-tree-item-classic__bracket--self-closing:after {
+  content: '/}}';
```
chancancode: There is no self-closing for curlies

app/helpers/component-argument.js
```diff
@@ -0,0 +1,25 @@
...
+    return `"${argument}"`;
+  } else if (argument === null) {
+    return 'null';
+  } else if (argument === undefined) {
+    return 'undefined';
+  } else if(isArray(argument) || typeof argument === 'object'){
+    return '...';
+  }
+
+  return argument;
```
chancancode: There are three main cases we need to handle 1) when it's a string, we need to quote with `JSON.stringify` (to handle quotes inside the string), 2) when it's typeof object and not `null`, we need to `...` (make this a link to the object inspector later), 3) for everything else, `String(...)` should do the trick (`null`, `undefined`, `true`, `false`, numbers).

The serialization guarantees that it's either a primitive (`null`, `undefined`, `true`, `false`, numbers, strings) or an object like `{ id: <object inspector id> }`. See `_serializeItem` in `render-tree.js`.

app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +21,26 @@
...
+      {{else}}
+        {{classify @item.name}}
+      {{/if}}
+
+      {{#each-in @item.args.positional as |name value|}}
```
chancancode: `#each`

app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +21,26 @@
...
+        {{component-argument value}}
+      {{/each-in}}
+
+      {{#each-in @item.args.named as |name value|}}
+        @{{name}}={{component-argument value}}
```
chancancode: Need something like this:

```suggestion
        {{#if @item.isCurlyInvocation}}
          {{name}}={{component-argument value}}
        {{else if (is-string value)}}
          @{{name}}={{component-argument value}}
        {{else}}
          @{{name}}=\{{{{component-argument value}}}}
        {{/if}}
```

app/templates/components/component-tree-item.hbs
rwwagner90: Why `#each`? `#each-in` is working here, I assumed because it is an object and not an array? Perhaps I am wrong though.

app/templates/components/component-tree-item.hbs
rwwagner90: What does this accomplish? The arguments are handled via the `component-argument` helper.

app/helpers/component-argument.js
rwwagner90: Does this really matter? If it's important, it should be enforced by the ember lint rules.

app/styles/component_tree.scss
rwwagner90: Ah, you're right. Good catch!

app/helpers/component-argument.js
chancancode: It's in the `.editorconfig` in the repo. If you are using vscode, there is an extension for it.

app/templates/components/component-tree-item.hbs
chancancode: It is an array. You _can_ use `{{each-in}}` on an array, for the same reason you _can_ use `for(... in ...)` on an array, which is that the array index are "properties" on the array "object". But it's definitely strange and suboptimal, so should be avoided.

app/templates/components/component-tree-item.hbs
chancancode: If using curly braces, you don't need `@` in front of args or curlies around args. If using angle bracket, you need both except in the string case where you don't need the curlies.

```
{{foo-bar some="arg" another=123456}}
<FooBar @some="arg" @another={{123456}} />
```

.prettierrc.js
```diff
@@ -0,0 +1,14 @@
...
+  overrides: [
+    {
+      files: '**/*.hbs',
+      options: {
+        parser: 'glimmer',
+        singleQuote: false
+      }
+    }
+  ]
+};
```
chancancode: newline 😅

rwwagner90:
@chancancode I think I addressed most feedback. Mind taking another quick look?

app/helpers/component-argument.js
chancancode: This is still needed. It fixes the problem with the string serialization (when the string has quotes, etc), and remove the unneeded branches.

```js
if (typeof argument === 'string') {
  return JSON.stringify(argument);
} else if (typeof argument === 'object' && argument !== null) {
  return '...';
} else {
  return String(argument);
}
```

app/styles/component_tree.scss
```diff
@@ -77,6 +78,20 @@
...
```
chancancode: This is unused?

app/templates/components/component-tree-item.hbs
```diff
@@ -21,14 +20,46 @@
   {{/if}}
 
   <code
-    class="component-tree-item__tag {{if @item.isComponent "component-tree-item__bracket"}} {{if (not @item.hasChildren) "component-tree-item__bracket--self-closing"}}"
+    class="component-tree-item__tag
+      {{if
+        (and (not @item.args.positional) @item.isComponent)
```
chancancode: We use `@item.args.positional` a lot so let's make it a getter in the class, i.e. `isCurlyInvocation` (in that case we can fold `@item.isComponent` into it also. The added benefit is that if we have a better way to detect curly invocation later we can just change that one place.

app/templates/components/component-tree-item.hbs
```diff
@@ -21,14 +20,46 @@
...
+        "component-tree-item__bracket"
+      }}
+      {{if
+        (and (not @item.args.positional) (not @item.hasChildren))
```
chancancode: Still need `@item.isComponent`. Alternatively, we can make this `component-tree-item__no-children` and deal with the composition in CSS.

app/templates/components/component-tree-item.hbs
```diff
@@ -21,14 +20,46 @@
...
+          {{#if (is-string value)}}
+            @{{name}}={{component-argument value}}
+          {{else}}
+            @{{name}}=\{{ {{~component-argument value~}}}}
+          {{/if}}
+        {{/each-in}}
+      {{/if}}
     {{else if @item.isOutlet}}
-      \{{outlet "{{@item.name}}"}}
+      {{outlet "{{@item.name}}"}}
```
chancancode: We should probably use your `component-argument` helper here (after you fixed to use `JSON.stringify`) since it has the same problem. Also the `\` is needed? Otherwise it will invoke `{{outlet}}` instead of rendering the string.

app/templates/components/component-tree-item.hbs
```diff
@@ -21,14 +20,46 @@
...
     {{else if @item.isEngine}}
-      \{{mount "{{@item.name}}"}}
+      {{mount "{{@item.name}}"}}
```
chancancode: Don't remove the `\` (unless I'm missing something)

app/templates/components/component-tree-item.hbs
```diff
@@ -51,4 +82,4 @@
       {{svg-jar "code-line" width="20px" height="20px"}}
     </button>
   {{/if}}
-</div>
+</div>
```
chancancode: newline 😅

rwwagner90: Even in hbs? I don't fully understand why we need this.

app/templates/components/component-tree-item.hbs
rwwagner90: Glimmer prettier formatter seems to have removed this for some reason. Will add it back.

app/templates/components/component-tree-item.hbs
chancancode: You are right, hbs is excluded from the stock editorconfig so it's fine one way or the other, although it probably doesn't hurt (the reason it's excluded is sometimes you really care that it doesn't have extra whitespace in HTML, but most of the time it doesn't matter). And in this case, you (your editor, presumably?) removed the newline even though you are not editing that and it doesn't seem intentional. This [blog post](https://thoughtbot.com/blog/no-newline-at-end-of-file) may help? It's mostly a convention thing that where a lot of tooling (git, GitHub, unix tools etc) have built around, and comes with some minor benefits also.

app/helpers/dasherize.js
```diff
@@ -0,0 +1,8 @@
+import { dasherize } from '@ember/string';
+import { helper } from '@ember/component/helper';
+
+export function dasherizeString([str]) {
+  return dasherize(str).replace(/\//g, '::');
+}
+
+export default helper(dasherizeString);
```
chancancode: This helper is unused, do you intend to keep it around?

chancancode: (if so, the `.replace(/\//g, '::')` is incorrect since in curly invocation it _should_ be slashes)

rwwagner90: Deleted it

nummi:
Maybe people never use these actions but they can easily get pushed out of view:
image

rwwagner90:
@nummi I am definitely open to suggestions. We could also maybe make them part of a context menu? What do other dev tools do?

nummi:
> @nummi I am definitely open to suggestions. We could also maybe make them part of a context menu? What do other dev tools do?

React puts that functionality in the sidebar for a selected node.
image

nummi:
Have you seen this behavior?

wut

chancancode:
We can also make clicking on the row scroll the element into view if needed (in addition to having the buttons in more convenient locations)? Anyhow, we still have to fix the CSS, because the background highlight and the closing tag is not rendering properly once scrolled.

chancancode:
We can also make clicking on the row scroll the element into view if needed (in addition to having the buttons in more convenient locations)? Anyhow, we still have to fix the CSS, because the background highlight and the closing tag is not rendering properly once scrolled.

chancancode:
@nummi hmm that is not expected. What ember-source version are you using and is the app available publicly? Can you test the latest commit on my PR and see if it also has this problem?

nummi:
> What ember-source version are you using and is the app available publicly? Can you test the latest commit on my PR and see if it also has this problem?

`3.15.0`. It's a personal app I don't have deployed anywhere. I did notice there are warnings: `Object Inspector error for debugName Error: You attempted to access the 'debugName' property on a Glimmer Component, but that property does not exist in Ember.js applications, it only exists in Glimmer.js apps`

chancancode:
Can you file a separate issue for that? Thank you 🙏

rwwagner90:
Closing in favor of https://github.com/emberjs/ember-inspector/pull/1112

PR #1094, opened 2019-12-12
fix issue with ember tracking reentry assert

patricklx:
this causes:
Assertion Failed: You attempted to update `[]` on `[object Object]`, but it had already been used previously in the same computation


ember_debug/deprecation-debug.js
```diff
@@ -206,7 +206,7 @@ export default EmberObject.extend(PortMixin, {
       // For ember-debug testing we usually don't want
       // to catch deprecations
       if (!this.get('namespace').IGNORE_DEPRECATIONS) {
-        this.get('deprecationsToSend').pushObject(deprecation);
+        this.get('deprecationsToSend').push(deprecation);
```
pzuraq:
```suggestion
        this.deprecationsToSend.pushObject(deprecation);
```

Is probably a better fix, the issue is `get` is entangling the array, and then we're immediately updating it. We still want to use the KVO compliant method if it's an Ember array, but we don't want to entangle.

PR #1095, opened 2019-12-13
Refactor `StorageService`

chancancode:
Ensure we are always using the in-memory storage service backend in tests, so that we aren't leaking test states into local storage, or otherwise change test behaviors due to existing items there.

### DIFF
--- a/app/components/list.js
+++ b/app/components/list.js
@@ -2,7 +2,6 @@ import Component from '@ember/component';
 import { run, scheduleOnce } from '@ember/runloop';
 import { task, timeout } from 'ember-concurrency';
 import ResizableColumns from 'ember-inspector/libs/resizable-columns';
-import LocalStorageService from 'ember-inspector/services/storage/local';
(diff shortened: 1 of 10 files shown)


chancancode:
@rwwagner90 The primary difference is that it used to be that everywhere we want to inject the service, we need to be careful to inject the "right" backend, but now there is an extra indirection (the "storage" service) to select the "right" backend once, then everywhere else we can just use it as usual. I also fixed some minor bugs regarding how `null` and `undefined` are handled, and making sure we still roundtrip the objects in the in-memory store, to ensure we don't accidentally rely on them being the stored objects being the same instance or trying to store unserializable objects.

chancancode:
Not sure why the linter wasn't catching it before and suddenly start to care, but it flagged a few "no-new-mixins" violations 🤔

rwwagner90:
@chancancode looks like this is passing now. Should we merge?

chancancode:
Yes 😄

PR #1096, opened 2019-12-13
Bump dependencies

chancancode:


lib/ui/app/templates/components/ui/empty-message.js
```diff
@@ -0,0 +1 @@
+export { default } from 'ui/templates/components/empty-message';
```
rwwagner90: Why are we moving all these files?

chancancode: This is related to #1101, but now superseded by #1102. Once that is merged I'll rebase this to drop that commit.

chancancode:
Don't merge this yet!

package.json
```diff
@@ -106,7 +107,10 @@
     "yauzl": "^2.10.0"
   },
   "engines": {
-    "node": "8.* || >= 10.*"
```
rwwagner90: Is this actually going to require 12+?

chancancode: @rwwagner90 As far as consumers go, it doesn't affect them, since they only consume the built artifacts. The Volta config is pinned to Node 12 so if you have it installed (you should, it's great 😄), it will automatically pick it up. That said, nothing in the deps actually require Node 12 so I'll change it back to `>= 10.*`. However, CI is running Node 12 (via Volta) so it's an untested setup.

chancancode:
Merged without squashing so we can revert the commit that pins broccoli-funnel later

PR #1098, opened 2019-12-14
List installed packages in ember-try

chancancode:


chancancode:
Apparently `yarn list` doesn't work correctly without the lock file!

PR #1099, opened 2019-12-14
Fix no-new-mixins

chancancode:
A new version of the `eslint-plugin-ember` caught a few more cases in this file. Since it's intentional here, we should just disable the rule for the whole file.


tests/ember_debug/object-inspector-test.js
```diff
@@ -222,21 +222,23 @@ module('Ember Debug - Object Inspector', function(hooks) {
   });
 
   test('Correct mixin order with es6 class', function (assert) {
-
+    // eslint-disable-next-line ember/no-new-mixins
```
rwwagner90: Should we maybe disable this rule for the whole project or just this whole folder? I think creating mixins in tests in okay.

PR #1103, opened 2019-12-17
Refactor acceptance tests

chancancode:
The main goal is to get the tests to more accurately reflect what actually happen in real applications. Instead of mocking/reopening the application "port", we now directly replace the adapter with a `TestAdapter` that faithfully implements the required hooks the same way that a "real" adapter does.

This allows us to write acceptance tests by only sending/responding to messages in a way that is, as a far as the inspector app can is concerned, indistinguishable from the "real" ember debug behaves in the real world. The new API also allows us to be a bit more precise in terms of the expectations (messages sent/received the correct number of times, in the right order, etc).

Split from #1088. I'll rebase that PR and merge in these changes when this lands.

### DIFF
--- a/app/components/app-picker.js
+++ b/app/components/app-picker.js
@@ -1,16 +1,11 @@
 import Component from '@ember/component';
-import { observer } from '@ember/object';
 import { alias, reads } from '@ember/object/computed';
 
 export default Component.extend({
   classNames: ['app-picker'],
 
-  apps: alias('port.detectedApplications.[]'),
-  selectedApp: reads('port.applicationId'),
-
-  selectedDidChange: observer('selectedApp', function() {
-    this.port.set('applicationId', this.selectedApp);
-  }),
+  apps: alias('port.detectedApplications'),
+  selectedAppId: reads('port.applicationId'),
 
   init() {
     this._super(...arguments);
@@ -19,8 +14,7 @@ export default Component.extend({
 
   actions: {
     selectApp(applicationId) {
-      this.set('selectedApp', applicationId);
-      this.port.send('app-selected', { applicationId });
+      this.port.selectApplication(applicationId);
     }
   }
 });
--- a/app/controllers/component-tree.js
+++ b/app/controllers/component-tree.js
@@ -98,7 +98,6 @@ export default Controller.extend({
    */
   pinnedObjectId: null,
   inspectingViews: false,
-  viewTreeLoaded: false,
 
   /**
    * Bound to the search field to filter the component list.
--- a/app/initializers/setup.js
+++ b/app/initializers/setup.js
@@ -1,20 +1,13 @@
-import { typeOf } from '@ember/utils';
 import config from 'ember-inspector/config/environment';
 import PromiseAssembler from "ember-inspector/libs/promise-assembler";
 
 export default {
   name: 'setup',
   initialize(instance) {
-    // {{EMBER_DIST}} is replaced by the build process.
-    instance.adapter = '{{EMBER_DIST}}';
+    // {{EMBER_DIST}} is replaced by the build process (basic, chrome, etc)
+    let Adapter = instance.resolveRegistration(`adapter:{{EMBER_DIST}}`);
 
     // register and inject adapter
-    let Adapter;
-    if (typeOf(instance.adapter) === 'string') {
-      Adapter = instance.resolveRegistration(`adapter:${instance.adapter}`);
-    } else {
-      Adapter = instance.adapter;
-    }
     register(instance, 'adapter:main', Adapter);
     instance.inject('controller:deprecations', 'adapter', 'adapter:main');
     instance.inject('route:application', 'adapter', 'adapter:main');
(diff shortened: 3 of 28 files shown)


chancancode:
I recommend that you start the review with `tests/test-adapter.js`, and also enable "Hide whitespace changes"

app/services/port.js
```diff
@@ -9,53 +10,49 @@ export default Service.extend(Evented, {
...
-    this.adapter.onMessageReceived(message => {
-      const { applicationId, applicationName } = message;
-
-      if (message.type === 'app-list') {
-        const apps = JSON.parse(message.appList);
-        apps.forEach((app) => {
-          if (!this.detectedApplications.mapBy('applicationId').includes(app.applicationId)) {
-            this.detectedApplications.push(app);
-          }
+        message.apps.forEach(({ applicationId, applicationName }) => {
```
rwwagner90: Does this not need to do a `push` into `this.detectedApplications`? Is it not a list?

chancancode: It is not a list anymore, it is a dictionary/object of `{ applicationId: applicationName }` (e.g. `{ "ember123": "Ember Inspector", "ember456": "Ember Observer" }`), so it uses `Ember.set` to instead

app/components/app-picker.js
```diff
@@ -1,16 +1,11 @@
...
 
 export default Component.extend({
   classNames: ['app-picker'],
 
-  apps: alias('port.detectedApplications.[]'),
-  selectedApp: reads('port.applicationId'),
-
-  selectedDidChange: observer('selectedApp', function() {
-    this.port.set('applicationId', this.selectedApp);
-  }),
```
chancancode: This seems unnecessary, since we already have an action here, and on the port side we also have a callback to maintain the `applicationId`/`applicationName` when they change.

app/controllers/component-tree.js
```diff
@@ -98,7 +98,6 @@ export default Controller.extend({
    */
   pinnedObjectId: null,
   inspectingViews: false,
-  viewTreeLoaded: false,
```
chancancode: Now unneeded, since the model hook of the route now blocks until the tree is loaded.

app/initializers/setup.js
```diff
@@ -1,20 +1,13 @@
...
-    // {{EMBER_DIST}} is replaced by the build process.
-    instance.adapter = '{{EMBER_DIST}}';
+    // {{EMBER_DIST}} is replaced by the build process (basic, chrome, etc)
+    let Adapter = instance.resolveRegistration(`adapter:{{EMBER_DIST}}`);
 
     // register and inject adapter
-    let Adapter;
-    if (typeOf(instance.adapter) === 'string') {
-      Adapter = instance.resolveRegistration(`adapter:${instance.adapter}`);
-    } else {
```
chancancode: I don't think this branch is possible: on line 9 we unconditionally set it to a string, so seems impossible for it to fall through.

app/routes/app-detected.js
```diff
@@ -19,6 +19,7 @@ export default Route.extend({
       this.applicationBooted = ({ booted }) => {
         if (booted) {
           port.off('general:applicationBooted', this.applicationBooted);
+          this.applicationBooted = null;
```
chancancode: In `deactivate` it checks for this variable to avoid calling `off` twice, but we never unset it.

app/routes/component-tree.js
```diff
@@ -8,21 +9,27 @@ export default TabRoute.extend({
   },
 
   model() {
-    return [];
+    return new Promise(resolve => {
+      this.port.one('view:viewTree', resolve);
+      this.port.send('view:getTree');
+    });
   },
```
chancancode: Now blocks until we get a tree, so we don't have to worry about the `viewTreeLoaded = false` state anymore

app/routes/deprecations.js
```diff
@@ -1,32 +1,39 @@
...
-  setupController() {
-    let port = this.port;
-    port.on('deprecation:deprecationsAdded', this, this.deprecationsAdded);
-    port.send('deprecation:watch');
+  model() {
+    return new Promise(resolve => {
+      this.port.one('deprecation:deprecationsAdded', resolve);
+      this.port.send('deprecation:watch');
+    });
+  },
```
chancancode: This now blocks until we received the message from ember-debug, which is unconditional (it still sends it even when there are no deprecations).

app/services/port.js
```diff
@@ -9,53 +10,49 @@ export default Service.extend(Evented, {
     this._super(...arguments);
 
     /*
-     * An array of objects of the form:
-     * { applicationId, applicationName }
+     * A dictionary of the form:
+     * { applicationId: applicationName }
```
chancancode: Keep track of the names as well, in addition to the IDs. Changed from an array to a dictionary object.

ember_debug/main.js
```diff
@@ -132,7 +132,6 @@ const EmberDebug = EmberObject.extend({
       this.startModule('deprecationDebug', DeprecationDebug);
 
       this.generalDebug.sendBooted();
-      this.viewDebug.sendTree();
```
chancancode: This will be requested by the component tree route as needed.

app/services/port.js
rwwagner90: Ah, I see. So this is setting the application name under the key of application id. Makes sense 👍 .

PR #1106, opened 2019-12-18
Fix inspecting Glimmer components w/ obj inspector

chancancode:
Glimmer components instead dev mode assertion getters to warn about mismatches between the Ember.js vs legacy Glimmer.js APIs. Skip these properties when inspecting them to avoid tripping them.

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -13,21 +13,28 @@ const {
 const { oneWay } = computed;
 const { backburner, join } = run;
 
-let glimmer;
-let metal;
+const GlimmerComponent = (() => {
+  try {
+    return window.require('@glimmer/component').default;
+  } catch(e) {
+    // ignore, return undefined
+  }
+})();
+
+let GlimmerReference = null;
+let metal = null;
 let HAS_GLIMMER_TRACKING = false;
 try {
-  glimmer = Ember.__loader.require('@glimmer/reference');
+  GlimmerReference = Ember.__loader.require('@glimmer/reference');
   metal = Ember.__loader.require('@ember/-internals/metal');
-  HAS_GLIMMER_TRACKING = glimmer &&
-                         glimmer.value &&
-                         glimmer.validate &&
+  HAS_GLIMMER_TRACKING = GlimmerReference &&
+                         GlimmerReference.value &&
+                         GlimmerReference.validate &&
                          metal &&
                          metal.track &&
                          metal.tagForProperty;
 } catch (e) {
-  glimmer = null;
-  metal = null;
+  // ignore
 }
 
 const keys = Object.keys || Ember.keys;
@@ -237,12 +244,12 @@ export default EmberObject.extend(PortMixin, {
               let tagInfo = tracked[item.name] || { tag: metal.tagForProperty(object, item.name), revision: 0 };
               if (!tagInfo.tag) return;
 
-              changed = !glimmer.validate(tagInfo.tag, tagInfo.revision);
+              changed = !GlimmerReference.validate(tagInfo.tag, tagInfo.revision);
               if (changed) {
                 tagInfo.tag = metal.track(() => {
                   value = get(object, item.name);
                 });
-                tagInfo.revision = glimmer.value(object, item.name);
+                tagInfo.revision = GlimmerReference.value(object, item.name);
               }
               tracked[item.name] = tagInfo;
             } else {
@@ -972,7 +979,7 @@ function calculateCPs(object, mixinDetails, errorsForObject, expensiveProperties
                 item.isTracked = true;
               }
             }
-            tagInfo.revision = glimmer.value(object, item.name);
+            tagInfo.revision = GlimmerReference.value(object, item.name);
             item.dependentKeys = getTrackedDependencies(object, item.name, tagInfo.tag);
           } else {
             value = calculateCP(object, item, errorsForObject);
@@ -1106,8 +1113,7 @@ function getDebugInfo(object) {
     if (object instanceof Ember.ObjectProxy && object.content) {
       object = object.content;
     }
-    // We have to bind to object here to make sure the `this` context is correct inside _debugInfo when we call it
-    debugInfo = objectDebugInfo.bind(object)();
+    debugInfo = objectDebugInfo.call(object);
   }
   debugInfo = debugInfo || {};
   let propertyInfo = debugInfo.propertyInfo || (debugInfo.propertyInfo = {});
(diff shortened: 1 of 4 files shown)


ember_debug/object-inspector.js
```diff
@@ -11,21 +11,28 @@ const {
 } = Ember;
 const { oneWay } = computed;
 
-let glimmer;
-let metal;
+const GlimmerComponent = (() => {
+  try {
+    return window.require('@glimmer/component').default;
```
lifeart: will it work in embroider?

tests/ember_debug/object-inspector-test.js
```diff
@@ -1005,4 +1013,59 @@ module('Ember Debug - Object Inspector', function(hooks) {
       ]);
     });
   }
+
+  // @glimmer/component 1.0 doesn't seem to support 3.4, even though it has the manager API
```
rwwagner90: Does this mean there will be errors with people using less than 3.8? We have to support 3.4+ currently.

ember_debug/object-inspector.js
rwjblue: TBH, I'm unsure, but assume not.

tests/ember_debug/object-inspector-test.js
rwjblue: No, this comment is saying that it isn't actually possible to use `@glimmer/component` on Ember < 3.8 at all (nothing to do with Ember Inspector). 

If we can, we should fix that issue, but it's unrelated to the specific changes here.

tests/helpers/setup-destroy-ei-app.js
```diff
@@ -22,7 +22,10 @@ export async function setupEIApp(EmberDebug, routes) {
     Router.map(routes);
   }
 
-  let App = Application.create({ autoboot: false });
+  let App = Application.create({
+    autoboot: false,
+    rootElement: document.getElementById('ember-testing'),
```
rwjblue: Normally this is done in `config/environment.js`, help me grok why we don't want that here?

ember_debug/object-inspector.js
lifeart: @ef4 ^^

tests/helpers/setup-destroy-ei-app.js
chancancode: @rwjblue there might be a better way to do it but I’m not entirely sure how. This is a completely manual setup helper (essentially replicating what setupApplicationTest does) that is used in ember-debug tests. It imports directly from @ember/application and doesn’t use the app in either app.js or test-helper.js. It is to simulate a “user app” similar to the ember-source tests. Presumably if we put this in the config I will still have to thread it through here? (Previously it was rendering into document.body, I noticed that and that the test div is more appropriate, and it makes it work with the dom helpers)

ember_debug/object-inspector.js
ef4: We can make this work under embroider, it's enough to declare `@glimmer/component` as an implicit-module.

But I'm more concerned that the reason this is being used is apparently to run `instanceof` checks. That is not going to be reliable as soon as people have more than one version of `@glimmer/component`, which I think is a use case @rwjblue has already been thinking about.

chancancode:
This conflicts with #1108 (logically, at least), so whichever is merged first, the other one will require a manual rebase.

ember_debug/object-inspector.js
rwjblue: Aye, I think that we might be better served by a `isGlimmerComponent` static property or something 🤔 in the long run

tests/helpers/setup-destroy-ei-app.js
rwjblue: Gotcha, thank you!

ember_debug/object-inspector.js
chancancode: This is really just a temporary UX improvement rather than a correctness fix.

The problem is that whenever you inspect a Glimmer Component instance with the object inspector, you will get a screen that looks like this:

70824213-c3d1a480-1daf-11ea-92b5-fc98373947f5

While it may look scary, nothing has really gone "wrong" here. It is fully expected that sometimes computing a property (previously just CPs, now getters too) could result in errors for a variety of reasons. Therefore, the object inspect code already wraps the computation around try/catch blocks and this is just the way it report those errors in the UI. The app didn't crash, no errors were sent to the console or bugsnag. This is all "working as expected".

The part where it is not so expected is that these why computing these properties causes errors in the first place. These properties doesn't really exists on the Glimmer components API, in the Ember.js version at least. However, they used to (and still) exist on the Glimmer.js version of the API. For backwards compatibility (and the lack of official modifiers support) they were not removed from the 1.0 API, but they remain available _only_ in Glimmer.js apps and not Ember.js apps.

To ease the confusion, Glimmer components, when operating in Ember.js apps and in development mode, install getters for these properties that throws an error when these "non-existent" properties are accessed. The intention is that you shouldn't need to know about them, but if you happen to be a Glimmer.js user, and you happen to be aware of them, and you call them in Ember.js apps, hopefully you will realize what's going on.

The problem is that by installing these getters, the object inspector now "sees" that they exist, and try to compute them eagerly whenever the object is being inspected. This is why, when inspecting a Glimmer component (in development mode), you see these red errors in the UI every time. (I think Glimmer Components should have installed those getters as non-enumerable, but I don't actually think it will change anything here.)

While nothing has technically gone wrong, and nothing is actually broken, this is certainly not great UX, basically caused by the composition of things that, on their own, seems reasonable enough.

This patch tries to put in a temporary fix by suppressing the errors where we can, in a best-effort fashion. If we fail to do so for any reason (couldn't find the package, or the `instanceof` check didn't match), it will basically fallback to the old behavior where it will try to compute those forbidden properties and shows that there were errors while doing so. Not ideal, but also not catastrophic. If there are ways to improve the reliability of this fix, we should definitely do it, but I think this version catches the vast majority of cases right now that it's worth doing, even though it's not very correct.

In my opinion (and @rwjblue's I believe), the correct fix is to make getter computation lazy, and just render a "..." in their place until the user clicks on them, at which point, if an error occurs, you got what you asked for and hopefully it will be more clear what's going on. Getters will fail to compute for perfectly good reason (e.g. you are only expect to call them when some other condition is true), and they may even be side-effectful, so I think eagerly computing all the getters (and CPs, for that matter) initially is not very sound.

Until we can do that, this temporary patch just paves over that more general problem in a very common case.

rwjblue: Discussed this a bit with @chancancode in discord, I think it's probably fine to ship this for now to avoid _completely_ failing when used with `@glimmer/component`. 

I think we ultimately want to solve this differently, most likely with:

1. a "handshake" that ember-inspector uses in order to avoid displaying certain properties (e.g. `instance.__emberInspectorIgnore = ['array','of','properties']`)
2. as @chancancode mentioned just above, the ember-inspector should avoid eagerly computing all getters on an object (showing a clickable `...` instead, similar to what you see when you `console.log(object)`)

rwjblue: I created https://github.com/emberjs/ember-inspector/issues/1110 to track removing this.

PR #1107, opened 2019-12-18
Only deepClone when without prototype extensions

chancancode:
This adds a non-trivial amount of cost for each message (especially the complex ones like the view tree) that is not necessary in most apps.

### DIFF
--- a/ember_debug/adapters/web-extension.js
+++ b/ember_debug/adapters/web-extension.js
@@ -66,28 +66,36 @@ export default BasicAdapter.extend({
   }
 });
 
-/**
- * Recursively clones all arrays. Needed because Chrome
- * refuses to clone Ember Arrays when extend prototypes is disabled.
- *
- * If the item passed is an array, a clone of the array is returned.
- * If the item is an object or an array, or array properties/items are cloned.
- *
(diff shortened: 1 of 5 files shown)


ember_debug/adapters/web-extension.js
```diff
@@ -77,28 +77,36 @@ export default BasicAdapter.extend({
...
-    clone = {};
-    keys(item).forEach(key => {
-      clone[key] = deepClone(item[key]);
-    });
-  }
-  return clone;
+let deepClone;
+
+if (Ember.ENV.EXTEND_PROTOTYPES.Array) {
+  deepClone = function deepClone(item) {
```
rwjblue: Just wanted to confirm, this is so that we can send the cloned item through the postmessage right? I ask because I want to confirm that we don't actually need to worry about deepClone being used just to be able to mutate the value...

chancancode: Yes. As far as. I can tell this is related to structure cloning. It’s done as one of the last steps before passing to chrome’s postMessage.

chancancode:
Oops! Forgot to squash and merge!

PR #1108, opened 2019-12-18
Don't abuse timers

chancancode:
Previously, the object inspector would schedule an infinite timer that fires every 300ms to update the current object (it actually fires even where there isn't a current object to update). This is extremely inefficient, and it causes a lot of extra work in big apps, since the timer starts a runloop which results in a full top-down revalidation. The more appropriate thing to do is tap into the runloop callbacks, since as far as Ember is concerned, if there are no runloops, there are no observable changes.

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -1,15 +1,17 @@
 import PortMixin from 'ember-debug/mixins/port-mixin';
-import { compareVersion } from 'ember-debug/utils/version';
+import bound from 'ember-debug/utils/bound-method';
 import { isComputed, isDescriptor, getDescriptorFor } from 'ember-debug/utils/type-check';
+import { compareVersion } from 'ember-debug/utils/version';
 import { typeOf } from './utils/type-check';
 
 const Ember = window.Ember;
 const {
   Object: EmberObject, inspect: emberInspect, meta: emberMeta,
   computed, get, set, guidFor, isNone,
-  cacheFor, VERSION
+  cacheFor, VERSION, run,
 } = Ember;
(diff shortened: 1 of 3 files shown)


rwwagner90:
@chancancode to clarify, there was not a 300ms timer that fired infinitely. Updates were fired when the `render` event was fired and there was a 300ms debounce.

chancancode:
@rwwagner90 This is unrelated to how we update the render tree. This is just how we update the currently inspected object. When you call updateCurrentObject (which we always do in init as of now), at the end of the method, it will always unconditionally (even when there isn’t a current object) schedule a setTimeout that calls itself again, run-wrapped, which will then schedule another timer, etc. Worse, i think it’s actually possible to call this method multiple times and end up with a set of 300ms infinitely repeating timers.

chancancode:
This conflicts with #1106 (logically, at least), so whichever is merged first, the other one will require a manual rebase.

wycats:
Awesome! Really good work.

ember_debug/object-inspector.js
```diff
@@ -1,15 +1,17 @@
...
 import { isComputed, isDescriptor, getDescriptorFor } from 'ember-debug/utils/type-check';
+import { compareVersion } from 'ember-debug/utils/version';
 import { typeOf } from './utils/type-check';
 
 const Ember = window.Ember;
 const {
   Object: EmberObject, inspect: emberInspect, meta: emberMeta,
   computed, get, set, guidFor, isNone,
-  cacheFor, VERSION
+  cacheFor, VERSION, run,
```
rwjblue: Thank you for not nesting this!

PR #1109, opened 2019-12-18
Adjust component tree colors

nummi:
@rwwagner90, here's a PR for your PR that is a PR for another PR.

image

image


chancancode:
Having seen both, I think I actually like showing the @ sign as the same color as the identifier a lot more! IMO, just like instance variables in Ruby, the @ sign is essentially part of the variable name (since @foo and foo are talking about two different variables), the same color really helps drive that point home.

nummi:
@chancancode, @rwwagner90, can you take a look at the change I made here: https://github.com/emberjs/ember-inspector/pull/1109/files#diff-a9290a87bac9f17e7512cb19a835f822R12 ?

It looks like the attr value coming out of Debug has double quotes around it — the result is the quotes being highlighted with the actual value. I felt it makes more sense for the quotes to be grey like the other less important syntax:

image

chancancode:
@nummi seems good to me

app/helpers/component-argument.js
```diff
@@ -9,7 +9,7 @@ import { helper } from '@ember/component/helper';
  */
 export function componentArgumentDisplay([argument]) {
   if (typeof argument === 'string') {
-    return JSON.stringify(argument);
+    return JSON.stringify(argument).replace(/^\"|\"$/g, '');
```
chancancode: At this point, this should probably just be `argument.replace(/\"/g, '\\"');`. Should probably leave a comment saying we are intentionally not wrapping it with quotes (doing it in the template), just escaping any interior quotes.

app/templates/components/component-tree-item.hbs
```diff
@@ -69,10 +73,11 @@
...
                 {{name}}
               </span>
-              =
+              ="
               <span class="value-token">
                 {{component-argument value}}
               </span>
+              "
             {{else}}
               <span class="at-token">
```
chancancode: my personal opinion is that we probably don't need this 🤔

app/helpers/component-argument.js
rwwagner90: I think this was a side effect of adding `JSON.stringify`. Before, the quotes were not part of the value.

rwwagner90:
Closing in favor of https://github.com/emberjs/ember-inspector/pull/1112

PR #1112, opened 2019-12-19
Add args to tree

rwwagner90:
Rebasing seemed hard, so opening a new PR

### DIFF
--- a/app/controllers/component-tree.js
+++ b/app/controllers/component-tree.js
@@ -194,16 +194,8 @@ class RenderItem {
     this.renderNode = renderNode;
   }
 
-  get args() {
-    return this.renderNode.args;
-  }
-
   get id() {
     return this.renderNode.id;
-  }
-
-  get isCurlyInvocation() {
-    return this.renderNode && this.renderNode.args && this.renderNode.args.positional;
   }
 
   get isOutlet() {
@@ -224,6 +216,14 @@ class RenderItem {
 
   get name() {
     return this.renderNode.name;
+  }
+
+  get args() {
+    return this.renderNode.args;
+  }
+
+  get isCurlyInvocation() {
+    return this.renderNode && this.renderNode.args && this.renderNode.args.positional;
   }
 
   get hasInstance() {
--- a/app/controllers/component-tree.js
+++ b/app/controllers/component-tree.js
@@ -223,7 +223,7 @@ class RenderItem {
   }
 
   get isCurlyInvocation() {
-    return this.renderNode && this.renderNode.args && this.renderNode.args.positional;
+    return this.renderNode.args && this.renderNode.args.positional;
   }
 
   get hasInstance() {
--- a/app/controllers/component-tree.js
+++ b/app/controllers/component-tree.js
@@ -194,8 +194,16 @@ class RenderItem {
     this.renderNode = renderNode;
   }
 
+  get args() {
+    return this.renderNode.args;
+  }
+
   get id() {
     return this.renderNode.id;
+  }
+
+  get isCurlyInvocation() {
+    return this.renderNode && this.renderNode.args && this.renderNode.args.positional;
   }
 
   get isOutlet() {
--- a/app/helpers/component-argument.js
+++ b/app/helpers/component-argument.js
@@ -10,7 +10,7 @@ import { helper } from '@ember/component/helper';
 export function componentArgumentDisplay([argument]) {
   if (typeof argument === 'string') {
     // Escape any interior quotes – we will add the surrounding quotes in the template
-    return argument.replace(/\"/g, '\\"');
+    return argument.replace(/"/g, '\\"');
   } else if (typeof argument === 'object' && argument !== null) {
     return '...';
   }
--- a/app/helpers/component-argument.js
+++ b/app/helpers/component-argument.js
@@ -1,4 +1,5 @@
 import { helper } from '@ember/component/helper';
+import truncate from 'ember-inspector/utils/truncate';
 
 /**
  * Determine the type of the component argument for display
@@ -10,7 +11,7 @@ import { helper } from '@ember/component/helper';
 export function componentArgumentDisplay([argument]) {
   if (typeof argument === 'string') {
     // Escape any interior quotes – we will add the surrounding quotes in the template
-    return argument.replace(/"/g, '\\"');
+    return truncate(argument.replace(/"/g, '\\"'));
   } else if (typeof argument === 'object' && argument !== null) {
     return '...';
   }
--- a/app/helpers/component-argument.js
+++ b/app/helpers/component-argument.js
@@ -1,0 +1,20 @@
+import { helper } from '@ember/component/helper';
+
+/**
+ * Determine the type of the component argument for display
+ *
+ * @method componentArgumentDisplay
+ * @param {*} argument
+ * @return {*} The argument with the correct type for display
+ */
+export function componentArgumentDisplay([argument]) {
+  if (typeof argument === 'string') {
+    return JSON.stringify(argument);
+  } else if (typeof argument === 'object' && argument !== null) {
+    return '...';
+  }
+
+  return String(argument);
+}
+
+export default helper(componentArgumentDisplay);
--- a/app/helpers/component-argument.js
+++ b/app/helpers/component-argument.js
@@ -9,7 +9,8 @@ import { helper } from '@ember/component/helper';
  */
 export function componentArgumentDisplay([argument]) {
   if (typeof argument === 'string') {
-    return JSON.stringify(argument).replace(/^\"|\"$/g, '');
+    // Escape any interior quotes – we will add the surrounding quotes in the template
+    return argument.replace(/\"/g, '\\"');
   } else if (typeof argument === 'object' && argument !== null) {
     return '...';
   }
--- a/app/helpers/component-argument.js
+++ b/app/helpers/component-argument.js
@@ -9,7 +9,7 @@ import { helper } from '@ember/component/helper';
  */
 export function componentArgumentDisplay([argument]) {
   if (typeof argument === 'string') {
-    return JSON.stringify(argument);
+    return JSON.stringify(argument).replace(/^\"|\"$/g, '');
   } else if (typeof argument === 'object' && argument !== null) {
     return '...';
   }
--- a/app/styles/component_tree.scss
+++ b/app/styles/component_tree.scss
@@ -8,6 +8,32 @@
   min-height: 22px;
   position: relative;
 
+  &__action {
+    align-items: center;
+    background: transparent;
+    border: 0;
+    cursor: pointer;
+    display: inline-flex;
+    height: 100%;
+    margin-left: 10px;
+    opacity: 0;
+    padding: 0;
+
+    &:focus {
+      outline: none;
+    }
+
+    &.disabled {
+      cursor: not-allowed;
+    }
+
+    polygon,
+    rect,
+    path {
+      fill: var(--base15);
+    }
+  }
+
   &:hover {
     background-color: var(--base05);
 
@@ -28,32 +54,6 @@
   padding-right: 3px;
 }
 
-.component-tree-item__action {
-  align-items: center;
-  background: transparent;
-  border: 0;
-  cursor: pointer;
-  display: inline-flex;
-  height: 100%;
-  margin-left: 10px;
-  opacity: 0;
-  padding: 0;
-
-  &:focus {
-    outline: none;
-  }
-
-  &.disabled {
-    cursor: not-allowed;
-  }
-
-  polygon,
-  rect,
-  path {
-    fill: var(--base15);
-  }
-}
-
 .component-tree-item__tag {
   cursor: default;
   display: flex;
@@ -72,7 +72,6 @@
     color: var(--base09);
   }
 
-  .at-token,
   .key-token {
     color: var(--component-attr);
   }
@@ -88,7 +87,6 @@
 
 .component-tree-item--selected .component-tree-item__tag {
   .component-name,
-  .at-token,
   .bracket-token,
   .key-token,
   .value-token {
--- a/app/styles/component_tree.scss
+++ b/app/styles/component_tree.scss
@@ -103,6 +103,16 @@
   }
 }
 
+/**
+ * Modifier
+ * highlighted - the component currently being previewed
+ */
+
+.component-tree-item.component-tree-item--highlighted {
+  background-color: var(--component-highlighted-bg);
+  border-radius: 0;
+}
+
 .component-tree-item--selected .component-tree-item__tag {
   .component-name,
   .bracket-token,
--- a/app/styles/component_tree.scss
+++ b/app/styles/component_tree.scss
@@ -45,10 +45,32 @@
 .component-tree-item__tag {
   cursor: default;
   display: flex;
+  white-space: nowrap;
 }
 
 .component-tree-item--has-instance .component-tree-item__tag {
   cursor: pointer;
+}
+
+.component-tree-item__tag .arg-token {
+  display: flex;
+  margin-left: 0.5rem;
+}
+
+.component-tree-item__tag .at-token {
+  color: var(--spec04);
+}
+
+.component-tree-item__tag .bracket-token {
+  color: var(--base07);
+}
+
+.component-tree-item__tag .key-token {
+  color: var(--spec02);
+}
+
+.component-tree-item__tag .value-token {
+  color: var(--spec03);
 }
 
 .component-tree-item__bracket:before {
@@ -63,6 +85,16 @@
 
 .component-tree-item__bracket.component-tree-item__bracket--self-closing:after {
   content: '/>';
+}
+
+.component-tree-item-classic__bracket:before {
+  color: var(--base07);
+  content: '{{';
+}
+
+.component-tree-item-classic__bracket:after {
+  color: var(--base07);
+  content: '}}';
 }
 
 .component-tree-item:hover {
--- a/app/styles/component_tree.scss
+++ b/app/styles/component_tree.scss
@@ -7,6 +7,18 @@
   margin: 0 3px;
   min-height: 22px;
   position: relative;
+
+  &:hover {
+    background-color: var(--base05);
+
+    .component-tree-item__action {
+      opacity: 1;
+    }
+
+    .component-tree-item__action.disabled {
+      opacity: 0.65;
+    }
+  }
 }
 
 .component-tree-item__expand {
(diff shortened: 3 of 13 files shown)


rwwagner90:
@chancancode @nummi there were a lot of conflicts since the big PR was merged, so I cherry picked our commits into this new PR. Please take a look!

app/controllers/component-tree.js
```diff
@@ -194,10 +194,18 @@ class RenderItem {
...
+  }
+
   get id() {
     return this.renderNode.id;
   }
 
+  get isCurlyInvocation() {
+    return this.renderNode && this.renderNode.args && this.renderNode.args.positional;
+  }
+
```
chancancode: Minor but can you move both `get args` and `get isCurlyInvocation` to _after_ `get name`? (Trying to maintain roughly [this ordering](https://github.com/emberjs/ember-inspector/blob/master/ember_debug/libs/render-tree.js#L28-L38) in all places for consistency and make things slightly easier to find).

Also, don't need to check `this.renderNode` as it's guaranteed to be present.

chancancode: Minor but can you move both `get args` and `get isCurlyInvocation` to _after_ `get name`? (Trying to maintain roughly [this ordering](https://github.com/emberjs/ember-inspector/blob/master/ember_debug/libs/render-tree.js#L28-L38) in all places for consistency and make things slightly easier to find).

Also, don't need to check `this.renderNode` as it's guaranteed to be present.

app/helpers/component-argument.js
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Determine the type of the component argument for display
+ *
+ * @method componentArgumentDisplay
+ * @param {*} argument
+ * @return {*} The argument with the correct type for display
+ */
+export function componentArgumentDisplay([argument]) {
+  if (typeof argument === 'string') {
+    return JSON.stringify(argument).replace(/^\"|\"$/g, '');
```
chancancode:
```suggestion
    // Escape any interior quotes – we will add the surrounding quotes in the template
    return argument.replace(/\"/g, '\\"');
```

app/styles/component_tree.scss
```diff
@@ -7,6 +7,18 @@
...
+    background-color: var(--base05);
+
+    .component-tree-item__action {
+      opacity: 1;
+    }
+
+    .component-tree-item__action.disabled {
+      opacity: 0.65;
+    }
+  }
```
chancancode: Remove this hunk (it was moved down due to linter complaints)

app/styles/component_tree.scss
```diff
@@ -45,36 +57,70 @@
...
-  .component-tree-item__action {
-    opacity: 1;
-  }
+.component-tree-item-classic__bracket:before {
+  content: '{{';
+}
 
-  .component-tree-item__action.disabled {
-    opacity: 0.65;
-  }
```
chancancode: Add this hunk back (it was moved down from above due to linter complaints)

app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +20,81 @@
...
+        "component-tree-item__bracket"
+      }}
+      {{if
+        (and (not @item.isCurlyInvocation) (not @item.hasChildren) @item.isComponent)
+        "component-tree-item__bracket--self-closing"
+      }}
+      {{if
+        (and @item.isCurlyInvocation @item.isComponent)
+        "component-tree-item-classic__bracket"
+      }}"
```
chancancode:
```hbs
{{if @item.isComponent
  (if @item.isCurlyInvocation
    "component-tree-item-classic__bracket"
    (if @item.hasChildren
      "component-tree-item__bracket--self-closing"
      "component-tree-item__bracket"
    )
  )
}}
```

(you can indent it however the linter wants it, but we should group the conditions)

app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +20,81 @@
...
+
+        {{#each-in @item.args.named as |name value|}}
+          <div class="arg-token">
+            {{#if (is-string value)}}
+              <span class="at-token">
+                @
+              </span>
+              <span class="key-token">
+                {{name}}
+              </span>
```
chancancode:
```
              <span class="key-token">
                 @{{name}}
               </span>
```

IMO we should just treat `@` as part of the key here.

app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +20,81 @@
...
+              ="
+              <span class="value-token">
+                {{component-argument value}}
+              </span>
+              "
+            {{else}}
+              <span class="at-token">
+                @
+              </span>
```
chancancode: Same re: treating `@` as part of the key.

app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +20,81 @@
...
+      {{#if @item.args.positional}}
+        <span class="component-name">
+          {{@item.name}}
+        </span>
+
+        {{#each @item.args.positional as |value|}}
+          <div class="arg-token">
+            <span class="value-token">
+              {{component-argument value}}
+            </span>
```
chancancode: This needs to be quoted now

```hbs
{{#if (is-string value)}}"{{/if}}
<span class="value-token">
  {{component-argument value}}
</span>
{{#if (is-string value)}}"{{/if}}
```

app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +20,81 @@
...
+
+        {{#each-in @item.args.named as |name value|}}
+          <div class="arg-token">
+            <span class="key-token">
+              {{name}}
+            </span>
+            =
+            <span class="value-token">
+              {{component-argument value}}
+            </span>
```
chancancode:
```hbs
{{#if (is-string value)}}"{{/if}}
<span class="value-token">
  {{component-argument value}}
</span>
{{#if (is-string value)}}"{{/if}}
```

app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +20,81 @@
...
+          </div>
+        {{/each-in}}
+      {{else}}
+        <span class="component-name">
+          {{classify @item.name}}
+        </span>
+
+        {{#each-in @item.args.named as |name value|}}
+          <div class="arg-token">
+            {{#if (is-string value)}}
```
chancancode: Instead of breaking up and duplicating the whole block, I think you can just add something like...

```hbs
{{#if (is-string value)}}"{{else}}<span class="bracket-token">\{{</span>{{/if}}
<span class="value-token">
  {{component-argument value}}
</span>
{{#if (is-string value)}}"{{else}}<span class="bracket-token">}}</span>{{/if}}
```

...around the values, similar to what I showed above.

chancancode:
I left some review comments, but I think the main issue is still the horizontal scrolling CSS issues? (scrolling breaks the row highlight/closing tag/buttons)

app/helpers/component-argument.js
rwwagner90: @chancancode doesn't replacing all interior quotes negate what we were trying to do with `JSON.stringify` in the first place?

app/styles/component_tree.scss
rwwagner90: I removed the one that was moved down. This is now the only copy of this.

app/helpers/component-argument.js
chancancode: It is replacing the interior quotes with `\"` so it's escaping them. `JSON.stringify` does (basically) two things for us here: 1) escape the interior quotes, 2) add the surrounding quotes. We wanted both before. Now we only want (1) because we are just going to add the surrounding quotes in the template so that we can highlight them differently. So, now `JSON.stringify` is essentially doing too much for us and instead of running (1) + (2) and then reverting the effect of (2), we should just write the code to do (1) at this point, which is what my suggestion does.

app/styles/component_tree.scss
chancancode: Weren't the linter unhappy with it? (Something something descending specificity, seen in the `ember build` output)

app/controllers/component-tree.js
rwwagner90: @chancancode I would disagree with that ordering, in favor of alphabetizing I think. People are not going to be familiar with what the API returns, that I know of.

chancancode: It's a logical grouping... `isRoot` and `isRouteTemplate` may sound similar but they do pretty different things and doesn't make a lot of sense to sort them next to each other just because of the name happens to match (and then you would have to resort when you refactor and change the name?). A lot of the other files (here, capture-render-tree, render-tree, tests, etc) have this same logical structure/ordering now, roughly from "most important" to "less important" things (subjective, but the ordering is at least consistent everywhere) and I would like to keep it that way. I just liked to that documentation comment because it happens to have enumerated them, but the code structure in other places follows that order (I try to at least).

app/templates/components/component-tree-item.hbs
```diff
@@ -21,10 +20,75 @@
...
+    class="component-tree-item__tag
+      {{if
+        @item.isComponent
+        (if
+          @item.isCurlyInvocation
+          "component-tree-item-classic__bracket"
+          (if
+            @item.hasChildren
+            "component-tree-item__bracket--self-closing"
+            "component-tree-item__bracket"
```
chancancode: Oops. My bad, these last two are in the reverse order I think (if hasChildren -> *not* self-closing)

nummi:
> I left some review comments, but I think the main issue is still the horizontal scrolling CSS issues? 

I don't know if we'll fix this before the next release. It has been an issue for a while and I'm afraid the fix will require some big changes to the markup and/or the CSS.

chancancode:
@nummi I don't think the problem is specific to our setup here, you can reproduce it with a pretty minimal html/css file:

```html
<style>
.scrollable {
  overflow: scroll;
  width: 100%;
  height: 100%;
}

.row {
  background-color: lightgrey;
  height: 50px;
}

.row:nth-child(2n) {
  background-color: lightslategrey;
}
</style>

<div class="scrollable">
  <div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div>

  <div class="row" style="width: 5000px"></div>

  <div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div><div class="row"></div>
</div>
```

I think we may have to switch to flexbox, or perhaps there is some other CSS Trick™ that would make it work here, but I'm pretty sure we can make _something_ work. Since it's a pretty major new feature, it's okay for it to take a bit longer (the 3.13.0 release already went out yesterday, so it's not a blocker), but we should find a way to introduce it without introducing a regression.

rwwagner90:
@chancancode I was talking to @nummi earlier and I think we should move the action icons into the object inspector. That would solve the issue of them being off the screen.

rwwagner90:
@chancancode I believe I resolved all the things you requested. Mind taking another look? I think we should get this in, and then have @nummi help us move the icons to the object inspector, so people do not have to scroll to see them.

nummi:
image

rwwagner90:
Nice work @nummi!

nummi:
https://github.com/emberjs/ember-inspector/pull/1129

app/controllers/component-tree.js
```diff
@@ -218,6 +218,14 @@ class RenderItem {
     return this.renderNode.name;
   }
 
+  get args() {
+    return this.renderNode.args;
+  }
+
+  get isCurlyInvocation() {
+    return this.renderNode && this.renderNode.args && this.renderNode.args.positional;
```
chancancode:
```suggestion
    return this.renderNode.args && this.renderNode.args.positional;
```

`this.renderNode` is guaranteed

app/templates/components/component-tree-item.hbs
```diff
@@ -1,54 +1,111 @@
 <div
   style={{@item.style}}
   class="
-    component-tree-item
+    component-tree-item relative flex items-center mx-1 rounded
+    {{if @item.hasInstance "cursor-pointer" "cursor-default"}}
     {{if @item.isComponent "component-tree-item--component"}}
-    {{if @item.hasInstance "component-tree-item--has-instance"}}
-    {{if @item.isSelected "component-tree-item--selected"}}
-    {{if @item.isHighlighted "component-tree-item--highlighted"}}
```
chancancode: why is this removed?

app/templates/components/component-tree-item.hbs
```diff
@@ -1,54 +1,111 @@
...
+              "component-tree-item__bracket"
+              "component-tree-item__bracket component-tree-item__bracket--self-closing"
+            )
+          )
+        }}"
     >
-      {{svg-jar "eye" width="20px" height="20px"}}
-    </button>
+      {{#if @item.isComponent}}
+        {{#if @item.args.positional}}
```
chancancode:
```suggestion
        {{#if @item.isCurlyInvocation}}
```

app/templates/components/component-tree-item.hbs
```diff
@@ -1,54 +1,111 @@
...
-      {{on "click" @item.inspectElement}}
-    >
-      {{svg-jar "code-line" width="20px" height="20px"}}
-    </button>
+          {{#each @item.args.positional as |value|}}
+            <div class="arg-token flex ml-2">
+              <span class="value-token">
+                {{component-argument value}}
+              </span>
+            </div>
```
chancancode: This need to check for strings and add quotes.

app/templates/components/component-tree-item.hbs
```diff
@@ -1,54 +1,111 @@
...
+
+          {{#each-in @item.args.named as |name value|}}
+            <div class="arg-token flex ml-2">
+              <span class="key-token">
+                {{name}}
+              </span>
+              =
+              <span class="value-token">
+                {{component-argument value}}
+              </span>
```
chancancode: This need to check for strings and add quotes.

chancancode:
> args in Glimmer components and no args on classic

@nummi are you testing both on the same app? I think we expect args to show up for both, assuming you have a recent enough version of `ember-source`. So if it's not showing up for classic components, there may be a bug.

app/templates/components/component-tree-item.hbs
nummi: As far as I could tell that was being used for the hover state. See: https://github.com/emberjs/ember-inspector/pull/1129/files#diff-cbd2aa0d940eef1f44430e6ca030fbd9L8

nummi:
> @nummi are you testing both on the same app? I think we expect args to show up for both, assuming you have a recent enough version of ember-source. So if it's not showing up for classic components, there may be a bug.

I am not. The classic app I tested in is deprecations.emberjs.com and the Glimmer app is ember-source: 3.15.0

app/templates/components/component-tree-item.hbs
nummi: Here is the original PR: https://github.com/emberjs/ember-inspector/pull/984

chancancode:
@nummi 👌 that's fine then. if you have any classic components in the 3.15 app, they would/should also show up with args.

Another thing I noticed is sometimes the strings passed to the components can be *very* long. Do we want to come up with a reasonable limit and truncate the rest with `...`? (The `component-argument` can handle this pretty naturally.)

app/templates/components/component-tree-item.hbs
chancancode: I see. Yeah that changed in Octane. It is now used to highlight the "currently previewing component". IMO the previous behavior is not that useful, since you can already see the parent/child relationship from the nesting.

There are two ways to "preview" a component, by hovering the rows in the component tree or by using the "live inspection" feature and hover over the DOM elements on the page:

highlight

Using CSS hover state only solves the first part and I think we still need this for the latter.

nummi: Oooooh. I see. I'll have to revert that.

rwwagner90:
@chancancode coming up with a limit sounds good.

nummi:
https://github.com/emberjs/ember-inspector/pull/1135

chancancode:
> @chancancode coming up with a limit sounds good.

Feel free to steal this:

```js
function truncate(str, limit = 20, rest = "...") {
  if (str.length <= limit) {
    return str;
  }

  if (rest.length >= limit) {
    throw new Error("rest cannot be longer than limit");
  }

  let parts = str.split(/(\s+|\b)/); // split by whitespace or word boundaries
  let selected = []; // the parts to keep
  let targetLength = limit - rest.length; // leave room for the "..."
  let currentLength = 0;

  while (true) {
    let candidate = parts.shift();

    if ((currentLength + candidate.length) <= targetLength) {
      selected.push(candidate);
      currentLength += candidate.length;
    } else {
      if (currentLength === 0) {
        // We have no choice but to break in the middle of a long word
        selected.push(candidate.slice(0, targetLength));
      }

      break;
    }
  }

  selected.push(rest);

  return selected.join("")
}
```

It ensures the resulting string (including the "..." if added) is shorter than the given limit, but try to break at word boundaries or whitespaces when possible:

```
> truncate("It ensures the resulting string (including the \"...\" if added) is shorter than the given limit, but try to break at word boundaries or whitespaces when possible:", 999)
"It ensures the resulting string (including the "..." if added) is shorter than the given limit, but try to break at word boundaries or whitespaces when possible:"

> truncate("It ensures the resulting string (including the \"...\" if added) is shorter than the given limit, but try to break at word boundaries or whitespaces when possible:", 100)
"It ensures the resulting string (including the "..." if added) is shorter than the given limit, ..."

> truncate("It ensures the resulting string (including the \"...\" if added) is shorter than the given limit, but try to break at word boundaries or whitespaces when possible:", 50)
"It ensures the resulting string (including the ..."

> truncate("It ensures the resulting string (including the \"...\" if added) is shorter than the given limit, but try to break at word boundaries or whitespaces when possible:", 20)
"It ensures the ..."

> truncate("It ensures the resulting string (including the \"...\" if added) is shorter than the given limit, but try to break at word boundaries or whitespaces when possible:", 10)
"It ..."
```

With this style the truncation, I find that 20 seems to be a pretty reasonable default (it will often be much shorter than that depending the length of the word that happen to sit at the boundary).

nummi:
Any chance we could use an ellipsis in place of three periods? "…"

chancancode:
Sounds good to me. The algorithm shot is work just as well

rwwagner90:
@chancancode I just pushed up the truncate and the adding quotes you requested. I was unable to test out the positional params because it seems like the renderNode is coming back with named args only, not positional ones, even when I use `{{set-body-class this.theme}}`

PR #1116, opened 2019-12-20
Bring back Node 8 support (for now)

chancancode:
Revert some dependencies bump from #1096 because `ember-electron` (probably the only npm consumer?) still support Node 8. This is mostly a technicality, since they don't actually run the build and only consume the artifacts in `dist` (included in the npm package).

We should probably find a better way for them to consume the app, but in the mean time, this will allow them to consume the latest inspector until Node 8 is officially EOL in a few weeks. We can try this again when Ember CLI drops Node 8 support (likely 3.17).

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,10 +45,10 @@
     "basiccontext": "^3.5.1",
     "broccoli-asset-rev": "^3.0.0",
     "broccoli-babel-transpiler": "^7.3.0",
-    "broccoli-concat": "^4.2.0",
+    "broccoli-concat": "^3.7.5",
     "broccoli-file-creator": "^2.1.1",
-    "broccoli-funnel": "3.0.0",
-    "broccoli-merge-trees": "^4.0.0",
+    "broccoli-funnel": "^2.0.2",
+    "broccoli-merge-trees": "^3.0.0",
     "broccoli-stew": "^3.0.0",
     "broccoli-string-replace": "^0.1.2",
     "codeclimate-test-reporter": "^0.5.1",
@@ -107,7 +107,7 @@
     "yauzl": "^2.10.0"
   },
   "engines": {
-    "node": ">= 10.*"
+    "node": "8.* || >= 10.*"
   },
   "ember": {
     "edition": "octane"
@@ -118,7 +118,7 @@
     ]
   },
   "volta": {
-    "node": "12.13.0",
+    "node": "8.17.0",
     "yarn": "1.19.1"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3633,22 +3633,23 @@ broccoli-concat@^3.2.2, broccoli-concat@^3.7.1, broccoli-concat@^3.7.4:
     lodash.uniq "^4.2.0"
     walk-sync "^0.3.2"
 
-broccoli-concat@^4.2.0:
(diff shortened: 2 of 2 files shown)


chancancode:
cc @bendemboski

chancancode:
@bendemboski just to set the expectation – in my personal opinion (other maintainers may disagree?) I doubt we will do a major version bump *just* to drop Node 8 support in the inspector, since that would signal the "wrong thing" to most consumers who consume this as an auto-updating browser extension. Not that the version number means anything there, but it does show up pretty prominently in the "What's New" screen. I think we can do at least minor bump (like this one) though. Just bringing this up since you may want to pin to `~` instead, for your purpose.

We should probably find you a better way to consume the package. The NPM build is not very coherent right now in my opinion, and this is one way that it surfaces. We may be better off switching to something like an S3 build that you can pull in.

bendemboski:
@chancancode yeah, completely agreed. Since this project isn't really intended to be consumed as a node module or run on Node by end users, it doesn't make sense to major version bump just because of a change in (what should be only) build-time Node support.

I agree with you that really we should find a better way for `ember-electron` to consume `ember-inspector` and I'm happy to make some time in the near future to do whatever is needed on the `ember-electron` side.

FWIW, [here](https://github.com/adopted-ember-addons/ember-electron/blob/master/vendor/install-extensions.js#L29-L51) is the code that installs `ember-inspector`. So the Node support thing is kinda silly because I'm pretty sure none of that code relies on the Node version at all -- the only problem is the install-time engines checks. I already want to rework how it's installed a bit, and was already planning to stop making `ember-inspector` a dependency of `ember-electron`, and instead install it in the Electron project that `ember-electron` generates when running its blueprint. That would allow me to install `ember-inspector` on its own with an `--ignore-engines` flag that would at least get us over this hurdle.

But I know you'd like to stop publishing all the build artifacts to npm, so it's probably worth discussing a more full solution so we stop using the npm registry at all. Perhaps we should discuss on Discord, or maybe in a separate issue either in `ember-inspector` or `ember-electron`?

And again, thank you very much for making this short-term solution happen!

chancancode:
Yes we should discuss more! 😄 I think your diagnosis is exactly right, I think none of this _actually_ matters but it just unfortunately cogs up the engine (no pun intended) here, and since this is an important update and there are otherwise no reasons for us to drop Node 8 support, I think it's totally justified to revert this so electron users can benefit from the update.

As far as publishing to npm or not – as far as I am aware, `ember-electron` is the only consumer so it's really just whatever is the most convenient for both of us. I think publishing _only_ the build artifacts to npm actually makes some sense, it's like how jQuery, moment.js or other browser-based libraries uses NPM just as a delivery mechanism. It may also make some sense to publish the unbuilt app if there is a way to consume that.

What doesn't make sense is doing _both_ at the same time in the same package, especially that is definitely no way to consume the app as an npm package today without doing a lot of hacks 😄

bendemboski:
>  I think publishing only the build artifacts to npm actually makes some sense

Very much agreed! That allows me to not worry about building some other mechanism to download the artifacts and get updates and whatnot, and also allows me to enable users to manage the `ember-inspector` version/updates their self using the familiar `npm`/`yarn` semantics.

PR #1117, opened 2019-12-20
Fix prototype extensions detection

### LINKED ISSUE #1114: Inspector does not recognize app if Ember.ENV.EXTEND_PROTOTYPES does not exist

On opening Ember inspector on one of my Ember apps, I am getting the below stack trace.

raven.js:383 Uncaught TypeError: Cannot read property 'Array' of undefined
    at n.callback (<anonymous>:402:35)
    at n.exports (vendor-ff8d5821b66c1…b3ac1b81a4b902.js:3)
    at n._reify (vendor-ff8d5821b66c1…b3ac1b81a4b902.js:3)
    at n.reify (vendor-ff8d5821b66c1…b3ac1b81a4b902.js:3)
    at n.exports (vendor-ff8d5821b66c1…b3ac1b81a4b902.js:3)
    at requireModule (vendor-ff8d5821b66c1…b3ac1b81a4b902.js:3)
    at sendVersionMiss (<anonymous>:9612:19)
    at <anonymous>:9420:7
    at triggerOnce (<anonymous>:9493:7)
    at completed (<anonymous>:5962:7)
    at d (raven.js:379)

On debugging, I found that it hits the code mentioned in the PR [here](https://github.com/emberjs/ember-inspector/pull/1107). Specifically: `if (Ember.ENV.EXTEND_PROTOTYPES.Array)` is the line throwing the error. In my app, Ember.ENV.EXTEND_PROTOTYPES is undefined.

## Your environment

This issue is definitely present on the latest version of the Chrome Browser tool. On Firefox, the stack trace instead says `TypeError: Ember.ENV.EXTEND_PROTOTYPES is undefined`, but is essentially caused by the same issue. The Ember CLI version for the app is 2.5.0.

chancancode:
Fixes #1114

### DIFF
--- a/ember_debug/adapters/web-extension.js
+++ b/ember_debug/adapters/web-extension.js
@@ -93,9 +93,23 @@ export default BasicAdapter.extend({
   }
 });
 
+// On some older Ember version `Ember.ENV.EXTEND_PROTOTYPES` is not
+// guarenteed to be an object. While this code only support 3.4+ (all
+// of which normalizes `EXTEND_PROTOTYPES` for us), startup-wrapper.js
+// eagerly require/load ember-debug modules, which ultimately causes
+// this top-level code to run, even we are going to pick a different
+// adapter later. See GH #1114.
+const HAS_ARRAY_PROTOTYPE_EXTENSIONS = (() => {
+  try {
+    return Ember.ENV.EXTEND_PROTOTYPES.Array === true;
+  } catch(e) {
+    return false;
+  }
+})();
+
 let deepClone;
 
-if (Ember.ENV.EXTEND_PROTOTYPES.Array) {
+if (HAS_ARRAY_PROTOTYPE_EXTENSIONS) {
   deepClone = function deepClone(item) {
     return item;
   };


ember_debug/adapters/web-extension.js
```diff
@@ -93,9 +93,22 @@ export default BasicAdapter.extend({
   }
 });
 
+// On some older Ember version `Ember.ENV.EXTEND_PROTOTYPES` is
+// not guarenteed to be an object. I (GC) am not really sure why
+// this code is even being executed on old Ember versions? If
+// we run the current adapters on 1.13 apps, it seems like we
+// are going to have bigger problems than just this? See #1114.
```
chancancode: Something really doesn't seem right here. cc @rwwagner90 do you have thoughts?

chancancode:
I think we should merge this and cut the patch release just as a matter of precaution, but we need to work with @jdrobertso to figure out what's up, because if we are using the wrong adapter this is probably not going to be the only problem.

jdrobertso:
Screenshot from 2019-12-20 14-01-30

Applying this fix to my own local build of the inspector, these are the results. It does at least detect the app as an ember app, but there do definitely seem to be other issues just at first glance.

chancancode:
@jdrobertso if you go to the "Info" tab, what does it show?

jdrobertso:
Screenshot from 2019-12-20 14-31-57

chancancode:
@jdrobertso since you are able to edit the inspector code, can you also add a `debugger;` or `console.log()` in the `init()` and `sendMessage()` functions in the same file, and see if they are ever called?

jdrobertso:
Output: 
```init called
sendMessage called
sendMessage called
init called
sendMessage called
init called
3 sendMessage called
sendMessage called
sendMessage called
sendMessage called
sendMessage called
sendMessage called
2 sendMessage called
sendMessage called
2 sendMessage called
sendMessage called```

chancancode:
@jdrobertso Hm okay, so here is the puzzling thing.

Since you are using an Ember version not supported by the latest inspector code (the current version supports 3.4 and above), the way it works is that the inspector is supposed to fallback to the older version that does support your Ember version. This is why in the Info tab you see Ember Inspector version "1.10.0", and that's evidence of things working correctly.

However, what is not working correctly is that, the code that causes the problem is only added very recently (the `Ember.ENV.EXTEND_PROTOTYPES` check), and so it's on the "adapter" that should only run on new apps, so it's puzzling why it's causing problems for you.

I have found at least one problem, which is that in the code that "selects" the right adapter to use: 

https://github.com/emberjs/ember-inspector/blob/master/ember_debug/vendor/startup-wrapper.js#L27-L28

Here we are eagerly requiring the current-gen adapter, causing the problematic code you found to run against your old Ember version even though it is not supported. If that's the only problem, we can tweak how the startup wrapper works, or wrap the code around a try/catch like I did.

However, the other think I asked you to check – by adding the `console.log` statements, is to confirm that the current-gen adapter is indeed *not* being used, i.e. we expected it to be "loaded" (hence running the detection code I added recently), but once loaded, it's left unused and the startup-wrapper should re-inject the older adapter into your app. However, based on your findings, it seems like somehow we are running the current-gen adapter against the old client app and old inspector app, which will for sure cause much more problems.

I think ultimately that's the thing we need to identify and fix.

jdrobertso:
After talking with the team who uses the inspector here for that version of the app, it seems to be behaving as expected with just the fix on this pull request. However, if I can provide any further information or be of any more help, I'll be happy to do so.

chancancode:
@jdrobertso Sounds great, I'll try to land this and release it in the next patch release, but I think we still need to dig deeper on the original issue. Can you change the `console.log()`s in `init`/`send` to `console.error()` to capture some stack traces for me?

jdrobertso:
[576878985875.log](https://github.com/emberjs/ember-inspector/files/3990436/576878985875.log)
Not the prettiest output, but here you go!

chancancode:
@jdrobertso Thank you so much! I'll aim to get this out soon!

PR #1128, opened 2020-01-08
ember inspects hangs up the app

patricklx:
updateCurrentObject is called many many times per second...
not sure about the correct solution here.

I guess this can also happen if a getter or CP of the current object calls a Ember.set or Ember.run?


rwwagner90:
I think that sounds like a problem in the app where it's infinite looping perhaps?

patricklx:
mmm, i could not reproduce it anymore... looks good now

PR #1129, opened 2020-01-09
Fix component list overflow

nummi:
This PR addresses two issues:

1. Clipping of a component background color that extends past the list bounds
    - This is addressed by adding a child element that has the same background color as the item
1. Component actions being hidden when the component extends past the list bounds
    - This is addressed by using `position: sticky` on the actions to pin them to the right side.

---

ℹ️ All additions to `app/styles/utils.scss ` are based off of Tailwind.

### DIFF
--- a/app/styles/base.scss
+++ b/app/styles/base.scss
@@ -14,6 +14,7 @@
   --unit10: calc(var(--unit1) * 10);
   --unit11: calc(var(--unit1) * 11);
   --unit12: calc(var(--unit1) * 12);
+  --transparent: transparent;
 }
 
 h1,
--- a/app/styles/base.scss
+++ b/app/styles/base.scss
@@ -14,7 +14,6 @@
   --unit10: calc(var(--unit1) * 10);
   --unit11: calc(var(--unit1) * 11);
   --unit12: calc(var(--unit1) * 12);
-  --transparent: transparent;
 }
 
 h1,
--- a/app/styles/component_tree.scss
+++ b/app/styles/component_tree.scss
@@ -1,30 +1,36 @@
 .component-tree-item {
-  align-items: center;
-  border-radius: 4px;
   color: var(--base12);
-  display: flex;
   font-size: 12px;
-  margin: 0 3px;
   min-height: 22px;
-  position: relative;
+
+  &__actions {
+    min-height: 22px;
+    opacity: 0;
+    right: var(--unit1);
+
+    /**
+      * __actions is position:sticky so it is always visible.
+      * Due to margins and padding on various elements, it was
+      * difficult to keep the background of __actions on top
+      * of .tree-item text and align with the .tree-item background.
+      * This pseudo element is 100% of the width of __actions plus
+      * a little extra to guarantee alignment.
+    */
+    &:after {
+      border-bottom-right-radius: var(--unit1);
+      border-top-right-radius: var(--unit1);
+      bottom: 0;
+      content: '';
+      left: calc(var(--unit1) * -1);
+      position: absolute;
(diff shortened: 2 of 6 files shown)


app/styles/base.scss
```diff
@@ -14,6 +14,7 @@
   --unit10: calc(var(--unit1) * 10);
   --unit11: calc(var(--unit1) * 11);
   --unit12: calc(var(--unit1) * 12);
+  --transparent: transparent;
```
nummi: Linter was screaming at me :-|

app/styles/utils.scss
```diff
@@ -65,7 +84,12 @@ $units: 0, 1, 2, 3, 4;
 .text-left { text-align: left; }
 .text-center { text-align: center; }
 .text-right { text-align: right; }
+.whitespace-no-wrap { white-space: nowrap; }
 
+.bg-transparent { background-color: var(--transparent); }
```
nummi: Linter was screaming at me :-|

app/styles/component_tree.scss
```diff
@@ -159,10 +133,6 @@
       fill: var(--focus-text);
     }
   }
-
-  .component-tree-item__action.disabled {
-    opacity: 0.65;
-  }
```
nummi: This functionality went away at some point. 👋

app/templates/components/component-tree-item.hbs
```diff
@@ -1,106 +1,109 @@
 <div
   style={{@item.style}}
   class="
-    component-tree-item
+    component-tree-item relative flex items-center mx-1 rounded
+    {{if @item.hasInstance "cursor-pointer" "cursor-default"}}
     {{if @item.isComponent "component-tree-item--component"}}
-    {{if @item.hasInstance "component-tree-item--has-instance"}}
-    {{if @item.isSelected "component-tree-item--selected"}}
-    {{if @item.isHighlighted "component-tree-item--highlighted"}}"
```
nummi: This was originally used to subtly highlight children of a selected component. At some point that changed and was being used for the hover state. There was a visible lag.

app/styles/component_tree.scss
```diff
@@ -123,17 +101,7 @@
 
 /**
  * Modifier
- * highlighted - children of selected component
- */
-
-.component-tree-item.component-tree-item--highlighted {
-  background-color: var(--component-highlighted-bg);
-  border-radius: 0;
-}
```
nummi: This was originally used to subtly highlight children of a selected component. At some point that changed and was being used for the hover state. There was a visible lag.

app/templates/components/component-tree-item.hbs
rwwagner90: Do we still highlight the children?

nummi: Nope. Not sure if we should consider this a regression or pretend it never existed :-P

rwwagner90: When was the last time it worked? If it worked right before the Octane stuff, we should probably fix it.

nummi: I think it got lost in the flurry of Octane changes.

nummi:
Fixed odd alignment.


screenshot

nummi:
Adding a gradient so the clipping doesn't look odd:

image

rwwagner90:
@chancancode looks like we have some failures. Are all the ember-try runs expected to run and pass? It does not seem to be a required check anymore, and it probably should be.

chancancode:
Failure is due to #1130. It is required for merging things into _master_, but this PR is not targeting master so the branch protection rules did not apply.

PR #1130, opened 2020-01-11
Fix canary test failure

chancancode:
See https://github.com/emberjs/ember.js/issues/18673


rwwagner90:
@chancancode it seems like there is another canary failure

chancancode:
I’ll investigate 🧐

rwwagner90:
@chancancode are we able to specify allowed failures with github actions? I think we allowed canary to fail before.

rwwagner90:
I made the canary tests not required, so we can merge this in.

PR #1132, opened 2020-01-13
 Remove invalid and unecessary legacy {{#each}}

mixonic:
In Ember 3.14ish {{#each}} without an argument throws an exception. The usage is not valid, however the tests also don't require this route to exist. So remove it.

The upstream issue https://github.com/emberjs/ember.js/issues/18673 should still get resolved but there isn't a need to use the old usage here.

### DIFF
--- a/tests/ember_debug/view-debug-test.js
+++ b/tests/ember_debug/view-debug-test.js
@@ -5,7 +5,6 @@ import {
   visit
 } from '@ember/test-helpers';
 import hasEmberVersion from '@ember/test-helpers/has-ember-version';
-import { A } from '@ember/array';
 import { run } from '@ember/runloop';
 import EmberComponent from '@ember/component';
 import EmberRoute from '@ember/routing/route';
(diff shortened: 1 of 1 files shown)


mixonic:
Actually I think the only usage of this route in test was removed here: https://github.com/emberjs/ember-inspector/commit/08a3cf58957dcf8bd35a6091a431db57693eb6ae#diff-83b9d766958a413fef9ac2c64e8c7f86L134 Going to try and simply remove it.

mixonic:
I'm getting an additional failure from `Ember Debug - Object Inspector: Tracked properties work`. It isn't clear if that is a new canary failure or related to this patch, but it isn't immediate understandable how it would be related :thinking:

mixonic:
It seems like that failure may be intermittent. I definitely don't get it locally every time.

mixonic:
It seems like that failure may be intermittent. I definitely don't get it locally every time.

mixonic:
Hm, yeah getting that "tracked properties don't work" error from CI: https://github.com/emberjs/ember-inspector/pull/1132/checks?check_run_id=387537014#step:8:81

chancancode:
See #1130 😅 I pinged @pzuraq to look at the other failure, which seems to be an issue with tracking

mixonic:
Yeah it solves the same issue but it seems like in a better way. The code doesn't need to be updated, it simply isn't used.

Imo we should merge one so nobody else fixes the same thing :-) seems fine, they incrementally are resolving the red state on master

rwwagner90:
I would merge if I could, but we have merging disabled when the build is failing. I'm going to close this as a duplicate of #1130 for now.

PR #1135, opened 2020-01-17
Bring back hover highlight in component tree

nummi:


app/styles/component_tree.scss
```diff
@@ -103,6 +103,16 @@
   }
 }
 
+/**
+ * Modifier
+ * highlighted - children of selected component
```
chancancode:
```suggestion
 * highlighted - the component currently being previewed
```

PR #1137, opened 2020-01-22
only boot inspector on HTML pages

### LINKED ISSUE #1136: injected script breaks browser default XML presentation
As a developer, I use web-browsers to visually inspect XML. Ember inspector injects a script tag that breaks the default browser presentation of the XML.

  - navigate to [https://blog.emberjs.com/feed.xml](https://blog.emberjs.com/feed.xml) in Firefox
  - Observe how the page appears as plaintext instead of formatted XML

I believe this occurs because we indiscriminately append the boot `<script>`:
https://github.com/emberjs/ember-inspector/blob/47138779d4f6c559bda90a0e53df5838ac6da446/skeletons/web-extension/boot.js#L3

### unexpected
Screen Shot 2020-01-20 at 16 22 06

### expected
Screen Shot 2020-01-20 at 16 21 18

efx:
Fix #1136.
I chose to whitelist the [HTML mime
type](https://www.iana.org/assignments/media-types/media-types.xhtml#text) only since it is simpler.
Per Godfrey's comment, we could declare [matching in the manifest](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Match_patterns) but it does so by URL or protocol. Because of that I think checking the mime type in the document is simpler and more readable.

### DIFF
--- a/skeletons/web-extension/boot.js
+++ b/skeletons/web-extension/boot.js
@@ -1,16 +1,18 @@
 // This script runs at document_start, avoid adding more stuff here!
 
 // TODO: make this conditional, only do it when requested.
+// only inject boot script when on an HTML page
+if (document.contentType === 'text/html') {
+  let id = `ember-inspector-boot-${(Math.random() * 100000000).toFixed(0)}`;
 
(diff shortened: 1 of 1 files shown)


chancancode:
@efx thank you!

PR #1147, opened 2020-02-21
`lib/ui` utility classes

nummi:
Moving more components to utility classes to help with the amount of CSS we maintain and help with possible move to Tailwind. Most utility classes added follow Tailwind CSS naming — there are unique classes marked at the bottom of `utils.scss`.

The easiest way to review is probably going through each commit one at a time.

### DIFF
--- a/app/styles/utils.scss
+++ b/app/styles/utils.scss
@@ -1,5 +1,8 @@
-$units: 0, 1, 2, 3, 4;
+$units: 0, 1, 2, 3, 4, 5, 6;
 
+.box-border { box-sizing: border-box; }
+
+.absolute { position: absolute; }
 .relative { position: relative; }
 
 .sticky {
@@ -7,9 +10,12 @@ $units: 0, 1, 2, 3, 4;
   position: sticky;
 }
 
-.right-0 {
-  right: 0;
-}
+.top-0 { top: 0; }
+.left-0 { left: 0; }
+.right-0 { right: 0; }
+
+.overflow-auto { overflow: auto; }
+.overflow-hidden { overflow: hidden; }
 
 .inline-block { display: inline-block; }
 .flex { display: flex; }
@@ -26,8 +32,6 @@ $units: 0, 1, 2, 3, 4;
 .self-stretch { align-self: stretch; }
 .justify-start { justify-content: flex-start; }
 
-.w-5 { width: var(--unit5); }
-.w-6 { width: var(--unit6); }
 .w-7 { width: var(--unit7); }
 .w-8 { width: var(--unit8); }
 .w-9 { width: var(--unit9); }
@@ -35,8 +39,10 @@ $units: 0, 1, 2, 3, 4;
 .w-11 { width: var(--unit11); }
 .w-12 { width: var(--unit12); }
 
+.w-full { width: 100%; }
 .h-full { height: 100%; }
 
+.m-auto { margin: auto; }
 .border-0 { border-width: 0; }
 
 @each $key in $units {
@@ -76,8 +82,6 @@ $units: 0, 1, 2, 3, 4;
   }
 }
 
-.border-0 { border-width: 0; }
-
 .rounded-none { border-radius: 0; }
 .rounded-sm { border-radius: calc(var(--unit1) / 2); }
 .rounded { border-radius: var(--unit1); }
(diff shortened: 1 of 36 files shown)


app/styles/utils.scss
```diff
@@ -80,18 +94,49 @@ $units: 0, 1, 2, 3, 4;
...
+
+.outline-none { outline: 0; }
+
+/*
+  Utilities not part of Tailwind:
+*/
+
+.text-inherit { color: var(--inherit); }
+
+.bg-base00 { background-color: var(--base00); }
```
rwwagner90: Tailwind supports setting your own colors to be used for both background-color and text color, so these could actually be part of tailwind.

app/styles/utils.scss
```diff
@@ -80,18 +94,49 @@ $units: 0, 1, 2, 3, 4;
...
+.bg-base01 { background-color: var(--base01); }
+.bg-base02 { background-color: var(--base02); }
+.bg-base05 { background-color: var(--base05); }
+.bg-spec00 { background-color: var(--spec00); }
+
+.text-base11 { color: var(--base11); }
+.text-base13 { color: var(--base13); }
+
+.text-12 { font-size: 12px; }
```
rwwagner90: I'm not sure how many text sizes we have, but maybe we should try to standardize on `text-xs` `text-sm` etc to align with tailwind?

lib/ui/addon/components/disclosure-triangle.hbs
```diff
@@ -1,6 +1,6 @@
 {{#if @toggle}}
   <button
-    class="disclosure-triangle {{if @expanded "expanded" "collapsed"}}"
+    class="disclosure-triangle {{if @expanded "expanded" "collapsed"}} appearance-none inline-block border-0 cursor-pointer bg-transparent"
```
rwwagner90: What is `appearance-none`? Is that setting `display: none`? If so, what toggles it being visible and removes the class?

lib/ui/addon/components/toolbar-search-field.hbs
```diff
@@ -1,15 +1,17 @@
...
     placeholder="Search"
+    class="box-border m-0 rounded pt-0 pr-4 pb-0 pl-1 text-12 leading-none outline-none bg-base00"
   />
 
   {{#if @value}}
     <button
-      class="toolbar__icon-button toolbar__search-clear-button js-search-field-clear-button"
+      class="toolbar__icon-button toolbar__search-clear-button absolute right-0 w-auto h-auto"
       title="clear"
+      data-test="search-field-clear-button"
```
rwwagner90: Could we use https://github.com/simplabs/ember-test-selectors for these selectors?

```suggestion
      data-test-search-field-clear-button
```

lib/ui/addon/components/disclosure-triangle.hbs
nummi: `appearance-none` removes platform specific styling on things like buttons and selects.

lib/ui/addon/components/toolbar-search-field.hbs
nummi: That's cool with me. I started down this road because it felt odd to have `js-test-xyz` along with a bunch of utility classes. If that seems like a good path I will add ember-test-selectors.

app/styles/utils.scss
nummi: I've been thinking about this. Since we have an app the spectrum of font sizes goes something like: `font-size: 10px – 15px`. Not sure it makes sense to have names like this:

```
```

lib/ui/addon/components/toolbar-search-field.hbs
rwwagner90: Yeah, I am a big fan of ember-test-selectors 👍

lib/ui/addon/components/disclosure-triangle.hbs
rwwagner90: Ah gotcha

app/styles/utils.scss
rwwagner90: I think your suggestion sounds fine:

```css
text-xs: 10px
text-sm: 11px
text-base: 12px
text-lg: 13px
text-xl: 14px
```

It is a little bit odd to have them all 1px different though. I think I would advocate more for setting a base font size and making all these `rem` based like tailwind does.

lib/ui/addon/components/toolbar-search-field.hbs
nummi: @rwwagner90, upgraded ember-test-selectors and moved everything in the app with `data-test="*"` to `data-test-*`

PR #1150, opened 2020-02-25
fix component attrs inspect

patricklx:
attrs is an object without prototype, so instanceof Object does not work but typeOf does


rwwagner90:
@chancancode @pzuraq does this look okay to you?

rwwagner90:
@chancancode @pzuraq should we merge this?

pzuraq:
Sure, seems fine for now

PR #1151, opened 2020-02-25
Update to ember 3.16

### LINKED ISSUE #1148: Upgrade to latest ember-{source,cli,data}
Our daily build just failed because the browser took too long to start. In recent ember-cli, it changed the default in testem.js to set a longer timeout, so it should have with that (on top of just generally good to be using the latest stable release).

Windvis:
Fixes #1148


config/optional-features.json
```diff
@@ -1,5 +1,6 @@
 {
   "application-template-wrapper": false,
+  "default-async-observers": true,
```
Windvis: Not sure if this was left out for a reason before? I just kept the blueprint default but I can flip the value if needed.

testem.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   launch_in_dev: [
     'Chrome'
   ],
-  browser_start_timeout: process.env.CI ? 60 : undefined,
+  browser_start_timeout: process.env.CI ? 120 : undefined,
```
Windvis: I kept the existing code but updated the value to match the blueprint.

Windvis:
I also bumped the pinned node and yarn versions in separate commits. I can drop / squash them if needed.

testem.js
rwwagner90: I think we could probably just set it to 120 all the time, like the new blueprint does.

Windvis:
Hmm I force pushed to my branch, but that doesn't seem to update the diff view in this MR?

testem.js
rwwagner90:
```suggestion
  browser_start_timeout: 120,
```

rwwagner90:
> Hmm I force pushed to my branch, but that doesn't seem to update the diff view in this MR?

I don't know why it didn't update, but I just pushed the change to `120` myself. I think we should be good to go now.

package.json
```diff
@@ -118,7 +119,7 @@
     ]
   },
   "volta": {
-    "node": "8.17.0",
-    "yarn": "1.19.1"
+    "node": "10.19.0",
+    "yarn": "1.22.0"
```
chancancode: cc @bendemboski do you need us to cut a new major version, or is a minor version bump sufficient for you?

bendemboski: I guess we need a major -- we depend on `^3.12.5` and haven't dropped node 8 support yet (and our 2.x/non-beta version also depends on `^3.12.5`), so a minor bump dropping node 8 support would break our node 8 support.

bendemboski: Thanks for checking!

Issue #1142, opened 2020-02-12
Implement Keyboard Navigation in the Component Tree


NullVoxPopuli:
## Issue and Steps to Reproduce

  - Describe your issue, stack trace, etc.

Left, Down, Up, Right arrows do not close, move down, move up, or open the component tree

  - Can you tell us how to reproduce it? Is there a site we can go to, [Ember Twiddle](https://ember-twiddle.com), [CodeSandbox](https://codesandbox.io)?

Point at component, try navigating via arrow keys

## Your environment

Ember version, browser, etc.

3.12.1 / chrome

## Screenshots

closed source :(

kangax:
Would love to have this as well. Coming from React devtools, this is much needed for quick navigation within a tree (including collapsing/expanding subtrees).

rwwagner90:
@NullVoxPopuli @kangax do either of you have time to give implementing this a try? I'm happy to help with guidance on how it could be implemented, but I won't be able to take a look at it myself for at least a couple weeks.

nummi:
I'll take it. :wave:

nummi:
@NullVoxPopuli https://github.com/emberjs/ember-inspector/pull/1153 — if you have time please review!

NullVoxPopuli:
@nummi I can provide feedback once released. My weekend computer isn't great at dev yet. (still trying to figure that out)

kangax:
Thanks for adding this! Would love a new release with this change, thank you ❤️

PR #1153, opened 2020-03-05
Component Tree Arrow Key Navigation

nummi:
Closes https://github.com/emberjs/ember-inspector/issues/1142

@NullVoxPopuli, here is a WIP. 

- [x] Prevent native scrolling when arrow keys are pressed
- [x] Should arrow keys navigate if the tree is visible or only if the tree has focus?
- [x] Holding down key is SLOW and glitchy


rwwagner90:
@nummi 
 
> Should arrow keys navigate if the tree is visible or only if the tree has focus?

I think only if the component tree window is focused, but we would need to make sure it is focusable with tab, etc.

> Holding down key is SLOW and glitchy

I would be fine with not allowing holding down personally.

nummi:
Changed to always use arrow keys when the component tree is visible _unless_ focus is in an input or textarea.

PR #1157, opened 2020-03-10
Base font size and utility classes

nummi:
- Tailwind uses normalize.css, we should too
- Set html font-size to `62.5%` so `1rem == ~10px`; makes rem sizes easier to grok
- I'll start using the new font-size utility classes in another PR

I've done a lot of clicking around and I don't see any visible changes to the text.

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -1,23 +1,3 @@
-/**
- APP
- ===
- The layout of the entire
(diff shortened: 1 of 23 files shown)


nummi:
> Looks like changing 213 to 212 should be sufficient in the test

I have no idea what is happening...

PR #1163, opened 2020-03-25
[FEAT] Add inspect store button to data pane in Ember Inspector

SYU15:
## Description
Add a button in data pane to enable user to inspect store and send it to console.

Would love some guidance as to what a good test would be since a lot of the end-to-end functionality seems to be stubbed out in other tests.

## Screenshots
Data_Pane_Store

### DIFF
--- a/app/controllers/model-types.js
+++ b/app/controllers/model-types.js
@@ -1,5 +1,5 @@
 import Controller from '@ember/controller';
-import { get, computed } from '@ember/object';
+import { action, get, computed } from '@ember/object';
 import { sort } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 
@@ -48,9 +48,14 @@ export default Controller.extend({
         return true;
       }
     });
-  })
-});
+  }),
 
+  getStore: action(function() {
+    this.port.send('objectInspector:inspectByContainerLookup', {
+      name: 'service:store'
+    });
+  }),
+});
 /**
  * Returns whether or not a given key has been set in storage.
  * @param {*} storage
--- a/app/templates/model-types-toolbar.hbs
+++ b/app/templates/model-types-toolbar.hbs
@@ -26,4 +26,7 @@
       Order Models By Record Count
     </label>
   </div>
+  <button class="toolbar__radio js-filter" {{on "click" this.getStore}} data-test-inspect-store>
+    Inspect Store
+  </button>
 </div>
--- a/tests/acceptance/data-test.js
+++ b/tests/acceptance/data-test.js
@@ -219,6 +219,22 @@ module('Data Tab', function(outer) {
 
       assert.dom('.js-model-type').exists({ count: 2 });
     });
+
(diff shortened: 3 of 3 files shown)


chancancode:
The code and functionality looks good to me. I'll leave it to @nummi see if there are any UI improvements to suggest. My initial thought is we should incorporate the `> $E` affordance into the button, maybe in addition to the text label.

Also, do you mind adding a test for this feature?

SYU15:
> Also, do you mind adding a test for this feature?

Happy to add a test @chancancode, however I didn't see an example of one that tested things end-to-end (I could have missed it) or if it is ok to stub out a lot of the functionality. Perhaps you could point me in the right direction or provide some guidance?

nummi:
How does this look?

image

image

SYU15:
@nummi @chancancode I updated the button to be the first suggested option. What do you think?
Screen Shot 2020-03-26 at 10 05 50 PM

chancancode:
IMO, the $E is important, since it tells you you can find it as that variable in the console, the > alone seems to generic to communicate that, so personally, I like the second option

SYU15:
@chancancode not sure if this is clear in the gif but right now clicking on the `inspect` button doesn't automatically send the console to the store, it opens up the right-hand panel where the user can click on the `$E`.
I was following the same pattern as the container pane. Is it problematic to add the `$E` to the button or is that ok? Or should I just have the button send the store directly to the console?

chancancode:
ohhhh sorry I got it all wrong! In that case I agree there is no need for the > $E. My bad!!

chancancode:
To be clear, I now think your original design, a button with just the label, is fine, since it doesn't actually send the store to the `$E` variable. We _may_ want a consistent affordance (icon, or whatever) for "inspecting X", but that would be a separate task since as you said. Sorry for the confusion!

chancancode:
For the test, I think you can add an acceptance test in `data-test.js`, that does something like:

```js
test('...', async function(assert) {
  await visit('/data/model-types');

  respondWith('objectInspector:inspectByContainerLookup', ({ name }) => {
    assert.equal(name, 'service:store');
    return false;
  });

  await click('[data-test-inspect-store]);
});
```

1. It visits the data tab.
2. It sets up the expectation that the `objectInspector:inspectByContainerLookup` will be sent, with the `name` argument `service:store`, and that we are not sending a response (`return false`).
3. It clicks the button (which is what triggers the message).
4. If the message was not sent, the test will fail.

Alternatively, in step 2, we _could_ set it up to send the proper response here (which would be sending the store service object from the "app" to the inspector). However, that would require quite a bit of set up, and then the inspector will send _more_ messages to inspect the properties on the store service object, which, imo, are not really the point of this test.

I think the functionality of the object inspector is already covered in other places, and in here we only really care about the button triggering the object inspector, so I think we can just stop there and trust that the object debug/inspector is going to do the right thing with that message (which is already covered elsewhere).

PR #1165, opened 2020-03-31
Introduce `setupTestAdapter` hook

chancancode:
While working on #1163, we encountered an issue with the test adapter, where unmet expectations (when you set up a `respondTo` for a test that was never triggered) did not cause the offending test to fail. Instead, they became unhandled promise rejections (i.e. "global errors") and fails the _next_ test that runs.

This is because we were using the `QUnit.{testStart,testDone}` hooks as if they were a global version of `{before,after}Each`. However, the global hooks are really intended for things like test reporters and runs too late to affect things like the pass or fail status of a test.

This commit introduces a `setupTestAdapter` function that uses the normal (not global) APIs so that the failures are attached to the correct test.

### DIFF
--- a/tests/acceptance/app-picker-test.js
+++ b/tests/acceptance/app-picker-test.js
@@ -5,9 +5,10 @@ import {
 } from '@ember/test-helpers';
 import { module, test } from 'qunit';
 import { setupApplicationTest } from 'ember-qunit';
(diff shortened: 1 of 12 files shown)


chancancode:
cc @SYU15

SYU15:
Thanks @chancancode!

chancancode:
canary failure is not related, @pzuraq is aware of it

PR #1166, opened 2020-03-31
Co-locate component templates

chancancode:
Don't remember if there was a reason we didn't do this already. Seems like it should just work, and IIRC #1102 should have already fixed any accidental problems.


rwwagner90:
@chancancode IIRC, I had several things colocated, and you moved them to not be. Do you recall why?

chancancode:
Don't think so! You may be remembering that we moved the in-repo addon in https://github.com/emberjs/ember-inspector/pull/1057, which caused some problems, and we talked about reverting it in https://github.com/emberjs/ember-inspector/pull/1101, but in the end we decided that was not necessary, and did https://github.com/emberjs/ember-inspector/pull/1102 instead. As far as I can tell, we just never did it in the app folder but I could be missing something.

Issue #1169, opened 2020-04-02
Make arguments in the component tree inspectable


SYU15:
Right now only primitives are supported, any objects render as `...`. This will require exploring a few UX options to make sure that it makes sense to the developer/isn't too cumbersome to inspect these arguments vs the entire row.

![Screen_Shot_2020-04-02_at_1 31 08_PM](https://user-images.githubusercontent.com/10149117/78298743-c33c5580-74e7-11ea-8c61-c0e4cbe89862.png)

rwwagner90:
Hi @SYU15! I think @chancancode and I had decided that anything more complex than primitives would require opening the object inspector for the whole row. Perhaps we could still do that, but go one level deeper into the property?

chancancode:
The idea is indeed to open the object inspector. However, opening the object inspector on the component instance isn't necessarily helpful, and in the case of template-only components, there is no component instance to inspect. We chatted on Discord a little, the idea is that when clicking on the `...` it will send that object to the object inspector, similar to what you said.

There are some potentially UX challenges here, as the whole row is currently a clickable area, and adding nested click areas could get uncanny, but @SYU15 will have to prototype it to see how much of an issue it is, in practice. If it turned out to be a problem, one idea would be to stop making the whole row clickable, and instead add a dedicated button to the right (where "scroll into view" and "inspect in elements panel" is currently).

Anyhow, we will still need @SYU15 to prototype the code needed to make inspecting component arguments possible before we can explore the UX problems and solutions.

SYU15:
Yeah @rwwagner90 I will prototype a few different flows and create a WIP PR and we can figure out which one provides the best user experience.

nummi:
<img width="496" alt="image" src="https://user-images.githubusercontent.com/3692/78463971-7d72bf00-7698-11ea-898e-68f3cb1fd462.png">

chancancode:
This is done!

Issue #1127, opened 2020-01-08
Cannot inspect args for certain components


afzal273:
## Issue and Steps to Reproduce
- Using Ember v3.15 on OS X, Chrome version 79.0.3945.88, Ember inspector v 3.13.1.
- Following ember guides super rental tutorial.
- Cannot inspect args for `Rental` component, parens are showing as empty in inspector and on clicking `$E` it shows ember proxy, so not sure where to find args in there. (screen capture attached)
- Also for certain components (`Rental::Image`) the details tab opens right away on clicking it but for others (like `Rental`), have to click on the button on top right of pane to see it, why is there that difference? Why not open the detail tab for every component when it is clicked? That will make inspecting easy. (can be seen in screen capture)

## Screenshots
![2020-01-07_18-38-07](https://user-images.githubusercontent.com/8137882/71945898-f25b5900-317c-11ea-95f1-e627852f4282.gif)

chancancode:
The component probably isn’t showing up because it is a template-only component and therefore doesn’t have an instance

afzal273:
I see, so if it is a template-only component then you have to click on the top right of the pane to see details? But that shouldn't be the case IMO, the user should be able to see details on clicking a component name regardless of what kind it is.

I can't inspect arguments for any component by clicking `$E` (if it's a template-only or not - in this example rental is a template only component and can't see args in the inspector, image and map are not template only and can only see args for map and only lat and lng but not the remaining ones). 

Also, can ember inspector allow clicking the caret to see details rather than having to click `$E` and checking console? The react inspector allows it and I think it will greatly improve productivity to be able to see details right in the ember inspector rather than having to go back and forth between console and inspector panes after clicking `$E`.

![2020-01-07_19-32-22](https://user-images.githubusercontent.com/8137882/71948408-c7750300-3184-11ea-81de-60b9b5ee484f.gif)
![2020-01-07_19-27-34](https://user-images.githubusercontent.com/8137882/71948414-c9d75d00-3184-11ea-9b50-458ea96a9163.png)

nummi:
> The component probably isn’t showing up because it is a template-only component and therefore doesn’t have an instance

Not opening the inspector or closing always feels like unexpected behavior. Perhaps would should keep the inspector open but have a message explaining that template-only components don't have an instance?

rwwagner90:
I would be on board with @nummi's suggestion unless you have any other ideas @chancancode?

nummi:
[...]

PR #1175, opened 2020-04-20
feat: allow object arguments in component tree to be inspected

SYU15:
## Description
Checks to see if an argument passed into component tree is an object, if so, allows the user to click on the argument directly to inspect the object properties.

Since the entire row is clickable in addition to the argument, I've created a few different options for UX which are shown below. I reused existing styles/color palettes but definitely open to suggestions.

**Currently working on tests.** Was curious to know why there weren't integration tests in this mp (unless I am missing something). Was thinking about adding a couple for `component-tree-arg` if that isn't an anti-pattern.

### **1. Most subtle - args is underlined when a user hovers over it**
underline_args

### **2. Less subtle - args is underlined and highlighted when user hovers over**
highlight_args2

### **3. Most apparent - args is a button that is clickable** 
Button_args

Addresses #1169 related to issue #1127


nummi:
Nice! So many options. We could even combine these and have a subtle version of 3 but only on hover.

SYU15:
> Nice! So many options. We could even combine these and have a subtle version of 3 but only on hover.

That is doable! I assume the suggestion is a button that highlights on hover @nummi?

rwwagner90:
Thanks for working on this @SYU15! I think I like the more subtle options, but I defer to @nummi on style decisions. Since we are making the clicks more explicit now, rather than just the whole row, perhaps we should do the same with the component itself? i.e. click the component name rather than anywhere in the row to inspect the component args.

I'm not sure if we want to change that or not, just throwing ideas out.

SYU15:
> i.e. click the component name rather than anywhere in the row to inspect the component args.

How would the user know the component name was the clickable area? I guess I'm slightly more in favor of having the whole row clickable because it seems more intuitive/obvious that way but open to suggestions.

rwwagner90:
@SYU15 yeah, it can stay as is. @nummi @chancancode any thoughts on which options you like best? 1,2 or 3? I like the more subtle ones, 1 or 2, but happy to go with everyone likes best.

chancancode:
Personally I am fine with any of these but wit a leaning towpaths not cluttering the screen when not hovering. So either 1, 2, what @nummi said, or showing all the click areas on the same row when the row itself is hovered.

When I raised the potential issues, I was not primarily worried about discoverability (although it certainly is an issue and these are good solutions). I was mostly worried about nesting clickable areas, which makes it easy to misclick something else by narrowly missing a pixel or two. Seems like it’s not as much of an issue as I worried about. I think the hover solutions here also helps (in addition to aiding discoverability) since you can physically see which of the two things will happen when you click.

We may also want to consider making the {{ and }} part of the click area, but exclude them from the visual highlight? (So you can hover on them to trigger the highlight, or click on them to inspect the arg, but it wouldn’t itself light up - I assume that doesn’t look good visually.)

Anyway i am pretty happy with any of these and we can also iterate a bit over time. Assuming it works, we can always start with the more basic/subtle options and add to it based on feedback after releasing.

rwwagner90:
@SYU15 I vote we go with solution number 2 for now. I think it's a good middle ground, and we can always tweak it later.

app/components/component-tree-arg.hbs
```diff
@@ -0,0 +1,13 @@
+{{#if this.valueIsObject}}
```
chancancode, line 1:
```suggestion
{{#if this.isObject}}
```

app/components/component-tree-arg.hbs
```diff
@@ -0,0 +1,13 @@
...
+  <span
+    class="value-token arg-value-token"
+    {{on "click" this.inspectArgumentValue}}
```
chancancode, line 4:
```suggestion
    {{on "click" this.inspect}}
```

app/components/component-tree-arg.hbs
```diff
@@ -0,0 +1,13 @@
+{{#if this.valueIsObject}}
+  <span
```
chancancode, line 2:
```suggestion
  <button type="button"
```

app/components/component-tree-arg.hbs
```diff
@@ -0,0 +1,13 @@
...
+    class="value-token arg-value-token"
+    {{on "click" this.inspectArgumentValue}}
+    data-test-arg-object
+  >
+    {{this.componentArgumentDisplay}}
```
chancancode, line 7:
```suggestion
    {{this.displayValue}}
```

app/components/component-tree-arg.js
```diff
@@ -0,0 +1,30 @@
+import Component from '@glimmer/component';
+import truncate from 'ember-inspector/utils/truncate';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class ComponentTreeArgs extends Component {
```
chancancode, line 6:
```suggestion
export default class ComponentTreeArg extends Component {
```

app/components/component-tree-arg.js
```diff
@@ -0,0 +1,30 @@
...
+  @service port;
+
+  get valueIsObject() {
```
chancancode, line 9:
```suggestion
  get isObject() {
```

app/components/component-tree-arg.js
```diff
@@ -0,0 +1,30 @@
...
+    return typeof this.args.componentArgValue === 'object' && this.args.componentArgValue !== null;
```
chancancode, line 10:
```suggestion
    return typeof this.args.value === 'object' && this.args.value !== null;
```

app/components/component-tree-arg.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  get componentArgumentDisplay() {
```
chancancode, line 13:
```suggestion
  get displayValue() {
```

app/components/component-tree-item.hbs
```diff
@@ -44,9 +44,7 @@
           {{#each @item.args.positional as |value|}}
             <div class="arg-token flex ml-2">
               {{if (is-string value) "\""}}
-              <span class="value-token">
-                {{component-argument value}}
-              </span>
+              <ComponentTreeArg @componentArgValue={{value}} />
```
chancancode, line 47:
```suggestion
              <ComponentTreeArg @value={{value}} />
```

SYU15:
Screen Shot 2020-04-27 at 7 17 43 PM

This is what it looks like now (imagine a pointer cursor as well, which the screenshot didn't capture). Have addressed the feedback from @chancancode.

nummi:
How does it feel with a border-radius on the highlight? Maybe the util class `rounded` would work (4px).

tests/acceptance/component-tree-test.js
```diff
@@ -354,4 +359,16 @@ module('Component Tab', function (hooks) {
     assert.equal(currentURL(), '/component-tree?pinned=render-node%3A3', 'It pins the element id as a query param');
     assert.dom('.component-tree-item--pinned').hasText('TodoList', 'It selects the item in the tree corresponding to the element');
   });
+
+  test('Can inspect component arguments that are objects in component tree', async function (assert) {
+    await visit('/component-tree');
+
```
chancancode, line 365:
```suggestion

```

(extra new line)

SYU15:
> How does it feel with a border-radius on the highlight? Maybe the util class `rounded` would work (4px).
> 
> Also, how does it look with the light theme?

Have added the border radius and I think it looks good, thanks for the suggestion!

The color scheme @nummi isn't perfect but it was kind of challenging to find a color that worked 
 decently under all conditions (which I have added below). Let me know if you can think of something more appropriate.

### Chrome Light:
Screen Shot 2020-04-27 at 11 43 26 PM
____________________________________________________________________
Screen Shot 2020-04-27 at 11 43 02 PM

### Chrome Dark:
Screen Shot 2020-04-27 at 11 43 49 PM
____________________________________________________________________
Screen Shot 2020-04-27 at 11 43 58 PM

### Firefox:
Screen Shot 2020-04-27 at 11 44 21 PM
____________________________________________________________________

Screen Shot 2020-04-27 at 11 45 04 PM

app/styles/component_tree.scss
```diff
@@ -101,6 +101,18 @@
...
+
+  .arg-value-token {
+    background: none;
+    border: none;
+    padding: 0;
+  }
+
+  .arg-value-token:hover {
+    background-color: var(--spec03);
+    color: var(--white);
```
chancancode, line 113: Instead of using a hardcoded color, try using one of the `--base` or `--spec` colors here (e.g. `--base00`), that way you will avoid the white-on-yellow low contrast problems when switching between light and dark themes. See https://github.com/emberjs/ember-inspector/blob/master/app/styles/colors.scss @nummi may have more concrete suggestion on which combination to pick.

app/components/component-tree-item.hbs
```diff
@@ -44,9 +44,7 @@
           {{#each @item.args.positional as |value|}}
             <div class="arg-token flex ml-2">
               {{if (is-string value) "\""}}
```
chancancode, line 46: We can save this refactor for another PR since this looks almost ready to go, but it had occurred to me that we should probably move the surrounding quotes and mustache (optional, via a `@mustache={{true}}` argument) into the component template, and turn `isString` into a getter instead of a global helper. Just wanted to note this somewhere so we don't forget.

app/styles/component_tree.scss
chancancode: The white-on-yellow contrast issue is probably a blocker, but should be an easy fix, it may or may not be the best looking option but switching to `--base00` should at least resolve the contrast problem.

chancancode:
Not nearly important enough the hold up the PR, but if you are making anymore style tweaks, but I am guessing the `...` case (and only the `...` case, not strings, etc) may look better with a tiny bit of horizontal padding (somewhere between 2-4px each side) so the highlight doesn't hug the text too tightly. Not a big deal if you don't get to it or if it doesn't end up working out though.

app/components/component-tree-item.hbs
SYU15: Sounds good @chancancode

SYU15:
Thanks for the tips @chancancode. Looks much better now:

Screen Shot 2020-04-28 at 5 25 42 PM
Screen Shot 2020-04-28 at 5 25 19 PM

PR #1180, opened 2020-04-30
Use brace expansion lint rule

ansmonjol:
Seems that eslint-plugin-ember support that now.

Test and linter are passing so I guess we're good with enabling this rule back.


ansmonjol:
@rwwagner90 do you have an idea about the failing builds ? 🤔

rwwagner90:
@ansmonjol try rebasing with master

rwwagner90:
@ansmonjol I don't think removing the comma to trigger the build is going to do it. Can you please rebase with master?

ansmonjol:
@rwwagner90 I was up to date with master 😕 
The failing job was pointing out a branch hash it cannot find. 
As I made a force push just before this failing test I hoped "just" retry CI jobs would work

ansmonjol:
Have different fail now.
Had the same issue in another project today, it was due to a `moment` sub-package version in yarn that was failing. Let me see if I can find it back.

See https://github.com/moment/moment/pull/5025

PR #1181, opened 2020-04-30
No observers lint rule

ansmonjol:
This MR aims to enable the lint rule `no-observer` back, in order to prevent anyone to use observer in the future.

I would prefer to have this rule enabled by default and disable linter when needed, so anyone is aware that observer usage is avoided when using them WDYT ?


rwwagner90:
I personally think people know not to use observers these days, but I suppose this would be fine.

PR #1184, opened 2020-05-06
Remove moment, bump deps

rwwagner90:
## Description

We did not seem to be using ember-cli-moment-shim for anything, and we also did not appear to be using js-string-escape, so I removed them. I also bumped various deps.

## Screenshots

N/A


rwwagner90:
@chancancode I think this fixes some build issues we have been having as well. Please let me know if it looks okay to you!

PR #1185, opened 2020-05-06
Start converting some things to octane

rwwagner90:
## Description

Converting a few things to glimmer components, native classes, and angle brackets


nummi:
Ran locally, I'm not seeing any issues when clicking around.

PR #1189, opened 2020-05-07
fix attrs/args - toString is undefined

patricklx:
getClassName already handles this case

### DIFF
--- a/tests/ember_debug/object-inspector-test.js
+++ b/tests/ember_debug/object-inspector-test.js
@@ -712,7 +712,7 @@ module('Ember Debug - Object Inspector', function(hooks) {
 
     let message = await inspectObject(proxy);
 
-    assert.ok(message.name.includes('(unknown)'));
+    assert.equal(message.name, 'TestObject');
 
     assert.equal(message.details[0].name, 'Basic Info');
     assert.equal(message.details[0].properties[0].name, 'name');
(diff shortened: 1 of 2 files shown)


rwwagner90:
@patricklx seems like we had a failing test. I restarted them in case it was a fluke, but may need to look into it.

tests/ember_debug/object-inspector-test.js
```diff
@@ -712,7 +712,7 @@ module('Ember Debug - Object Inspector', function(hooks) {
 
     let message = await inspectObject(proxy);
 
-    assert.ok(message.name.includes('(unknown)'));
+    assert.ok(message.name.includes('ObjectProxy'), 'object name should start with <ObjectProxy:');
```
rwwagner90, line 715: @patricklx Is this the intended output? I suppose it is better than `unknown`, but just want to make sure things are working as intended.

patricklx: yes, since getClassName replaces that part with the class name. But i think it only does it dev builds. in prod it will show unknown again

rwwagner90: Is there not a way to show the same things in both?

patricklx: no, since the the class name comes from constructor.name, which will get uglified in production

PR #1193, opened 2020-05-08
Fix string arguments in component tree to render only one pair of double quotes fixes #1190

### LINKED ISSUE #1190: Component tree: incorrect string component arguments serialization
image

`@col=""1""` should be `@col="1"` (without duplicated quotes)

testing https://ember-twiddle.com/a1a1448823aad56dfdd4615b0e5bf0b8?openFiles=templates.components.mega-input%5C.hbs%2C

SYU15:
This was a bad merge while experimenting with refactoring the surrounding string code for argument values.

Reverting the logic back to the way it was here: https://github.com/emberjs/ember-inspector/pull/1175/files#diff-a9290a87bac9f17e7512cb19a835f822L14 (in `app/helpers/component-argument.js`)

### After fix

Screen Shot 2020-05-07 at 4 01 19 PM


chancancode:
Unfortunately we don't have a good way to test this right now, the integration test harness is broken for some reason, but I'm looking into it.

rwwagner90:
I forget why we needed this stuff in the first place. Were we trying to handle if the strings included `"` in their output I think? I think the idea was make sure we remove all of them, then add some surrounding ourselves.

chancancode:
@rwwagner90 Yeah, the surrounding quotes is added in the template outside, so that the quotes are not part of the highlighting. This point is mostly moot now that we refactored it from being a helper to the component, and the component has its own template. This is the refactor that I was referring to. The quotes should still be added in template-land (instead of inside this JavaScript function), just that it should be moved from the tree-item component template into the tree-arg component template.

rwjblue:
> Unfortunately we don't have a good way to test this right now, the integration test harness is broken for some reason

Fixed by #1203. 

@SYU15 - If you rebase against master, you should be able to use standard `setupRenderingTest` style tests here.

SYU15:
Thanks @rwjblue @chancancode @rwwagner90!

PR #1198, opened 2020-05-09
Sort container instances alphabetically

lukemelia:
## Description

This commit uses `<EmberTable>`'s sorting capabilities to sort container
instances alphabetically be default. A sort indicator is shown in the
column header, and the header can be clicked to reverse the sort.

This PR adds test coverage for sorting and converts the template and
tests to use `data-test-` selectors instead of `js-` css classes.

## See It In Action

2020-05-09 at 1 37 PM

### DIFF
--- a/app/controllers/container-type.js
+++ b/app/controllers/container-type.js
@@ -1,5 +1,5 @@
-import { action, get, computed } from '@ember/object';
 import Controller, { inject as controller } from '@ember/controller';
+import { action, computed, get, set } from '@ember/object';
 import debounceComputed from 'ember-inspector/computed/debounce';
 import searchMatch from 'ember-inspector/utils/search-match';
 
@@ -10,17 +10,9 @@ export default Controller.extend({
 
   search: null,
 
-  filtered: computed('model.@each.name', 'search', function() {
+  rows: computed('model.@each.name', 'search', function() {
     return this.model
-      .filter((item) => searchMatch(get(item, 'name'), this.search));
-  }),
-
-  rows: computed('filtered.[]', function() {
-    return this.filtered.map(function(item) {
-      return {
-        name: item
-      };
-    });
+      .filter((instance) => searchMatch(get(instance, 'name'), this.search));
   }),
 
   init() {
@@ -31,7 +23,13 @@ export default Controller.extend({
       name: 'Name'
     }];
 
-    this.sortProperties = ['name'];
+    // By default, sort alphabetically
+    this.sorts = [
+      {
+        valuePath: 'name',
+        isAscending: true
+      }
+    ]
   },
 
   /**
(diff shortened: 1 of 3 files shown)


app/controllers/container-type.js
```diff
@@ -10,17 +10,9 @@ export default Controller.extend({
...
     return this.model
-      .filter((item) => searchMatch(get(item, 'name'), this.search));
-  }),
-
-  rows: computed('filtered.[]', function() {
-    return this.filtered.map(function(item) {
-      return {
-        name: item
-      };
-    });
```
chancancode, line 23: Good catch, thank you 🤦

app/controllers/container-type.js
```diff
@@ -31,26 +23,52 @@ export default Controller.extend({
...
     });
   }),
 
   sendContainerToConsole: action(function() {
     this.port.send('objectInspector:sendContainerToConsole');
   }),
+
+  @action
+  updateSorts(sorts) {
+    let hasExistingSort = this.sorts && this.sorts.length;
```
chancancode, line 57: Seems like this is always expected to be true right? Since we initialize it to non-empty and also normalizes it to be non-empty in this setter.

app/controllers/container-type.js
```diff
@@ -31,26 +23,52 @@ export default Controller.extend({
...
+    if (hasExistingSort && isDefaultSort) {
+      // override empty sorts with reversed previous sort
+      let newSorts = [
+        {
+          valuePath: this.sorts[0].valuePath,
+          isAscending: !this.sorts[0].isAscending,
+        },
+      ];
+      set(this, 'sorts', newSorts);
+      return;
```
chancancode, line 69:
```suggestion
      // override empty sorts with reversed previous sort
      sorts = [
        {
          valuePath: this.sorts[0].valuePath,
          isAscending: !this.sorts[0].isAscending,
        },
      ];
```

app/controllers/container-type.js
```diff
@@ -10,17 +10,9 @@ export default Controller.extend({
...
-      .filter((item) => searchMatch(get(item, 'name'), this.search));
-  }),
-
-  rows: computed('filtered.[]', function() {
-    return this.filtered.map(function(item) {
-      return {
-        name: item
-      };
-    });
+      .filter((instance) => searchMatch(get(instance, 'name'), this.search));
```
chancancode, line 15:
```suggestion
      .filter((instance) => searchMatch(instance.name), this.search));
```

app/controllers/container-type.js
```diff
@@ -31,26 +23,52 @@ export default Controller.extend({
...
    * Called whenever an item in the list is clicked.
    *
    * @method inspectInstance
-   * @param {Object} obj
+   * @param {Object} instance
    */
-  inspectInstance: action(function(obj) {
-    if (!get(obj, 'inspectable')) {
+  inspectInstance: action(function(instance) {
+    if (!get(instance, 'inspectable')) {
```
chancancode, line 43:
```suggestion
    if (!instance.inspectable) {
```

chancancode: Let's also flip this logic around and do `if (instance.inspectable) { ... }` instead of the early return

app/controllers/container-type.js
```diff
@@ -31,26 +23,52 @@ export default Controller.extend({
...
       return;
     }
     this.port.send('objectInspector:inspectByContainerLookup', {
-      name: get(obj, 'fullName')
+      name: get(instance, 'fullName')
```
chancancode, line 47:
```suggestion
      name: instance.fullName
```

app/templates/container-type.hbs
```diff
@@ -1,26 +1,34 @@
 <EmberTable as |t|>
   <t.head
     @columns={{columns}}
+    @sorts={{sorts}}
```
chancancode, line 4:
```suggestion
    @columns={{this.columns}}
    @sorts={{this.sorts}}
```

app/templates/container-type.hbs
```diff
@@ -1,26 +1,34 @@
...
+    @onUpdateSorts={{this.updateSorts}}
     @enableReorder={{false}}
   />
   <t.body
     @rows={{rows}}
```
chancancode, line 9:
```suggestion
    @rows={{this.rows}}
```

app/controllers/container-type.js
lukemelia: 👍  eliminating...

lukemelia:
Thanks for the review @chancancode. All makes sense -- pushed the changes. 👍

chancancode:
Thanks @lukemelia!

lukemelia:
🎉  Glad to be able to contribute to a great tool! Thanks @chancancode

PR #1202, opened 2020-05-12
Ensure renderDebug is properly unregistered.

rwjblue:
Adds a side effecting import to `tests/test-helper.js` to ensure that `ember-debug/main` can be imported/required eagerly in the test suite (and is not broken by a cycle).

Updates all lazy requires for `ember-debug` modules to ES module import statements.

### DIFF
--- a/ember_debug/render-debug.js
+++ b/ember_debug/render-debug.js
@@ -29,7 +29,7 @@ export default EmberObject.extend(PortMixin, {
     };
 
     this.profileManager.offProfilesAdded(this, this.sendAdded);
-    this.profileManager.offProfilesAdded(this, this._updateRenderPerformanceAndComponentTree);
+    this.profileManager.offProfilesAdded(this, this._updateComponentTree);
   },
 
   sendAdded(profiles) {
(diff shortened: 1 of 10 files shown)


ember_debug/render-debug.js
```diff
@@ -29,7 +29,7 @@ export default EmberObject.extend(PortMixin, {
     };
 
     this.profileManager.offProfilesAdded(this, this.sendAdded);
-    this.profileManager.offProfilesAdded(this, this._updateRenderPerformanceAndComponentTree);
```
rwjblue, line 32: This method doesn't exist on `ember-debug/render-debug` at all. It was removed in https://github.com/emberjs/ember-inspector/commit/47138779d4f6c559bda90a0e53df5838ac6da446#diff-04f0316e4c3ec9f79f8bf1ec48f6aafbL45, but we missed updating the cleanup code with the renamed method.

chancancode:
Looks good to me but is it possible to set up a CI scenario where we run the app test only, to avoid accidentally relying on ember debug existing?

chancancode:
(ember-try tests are ember debug only, so they are like the opposite of that, and this new scenario can go into the same “phase”)

PR #1206, opened 2020-05-12
Add Prettier to ESLint configuration.

rwjblue:
Prior to this change, we had a `.prettierrc.js` in the repo that some
contributors used with their editors' "format on save" ability, but
since it was not enforced by linting it was very inconsistent throughout
the codebase.

This change does a few things (best reviewed as separate commits):

* Update eslint to v7.0.0
* Add `prettier` related packages to `devDependencies`
* Configure prettier setup in `.eslintrc.js`
* Run `yarn lint:js --fix`


.eslintrc.js
```diff
@@ -62,13 +64,12 @@ module.exports = {
         node: true
       },
       plugins: ['node'],
-      rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
```
rwjblue, line 65: ESLint 6.8+ allows us to use `extends` directly in the overrides configuration, so we no longer need this manual shenanigans.

rwjblue:
I usually try to go with the defaults as much as possible (with single quotes being the one exception). Ember, Ember Data, Ember CLI, ember-test-helpers all have trailing commas on (set to `es5`).

Also, FWIW, in this PR I didn't actually change/modify the prettier configuration (it was introduced in https://github.com/emberjs/ember-inspector/pull/1112 / https://github.com/emberjs/ember-inspector/pull/1093).

rwwagner90:
@rwjblue yeah, trailing commas wasn't on by default until prettier 2.0 though. Now that it is on by default, I typically disable. I am fine going with what the group wants though, not going to die on this hill.

PR #1216, opened 2020-05-14
Use computed macro

ansmonjol:
## Description

This MR aims to fix the future linter error while [upgrading eslint-plugin-ember from 7.13.0 to 8.5.1](https://github.com/emberjs/ember-inspector/pull/1211)

See related issue: https://github.com/emberjs/ember-inspector/issues/1197

And the rule documentation: https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-computed-macros.md

Issue #1197, opened 2020-05-09
Upgrade to eslint-plugin-ember latest (v8)


chancancode:
I had to remove that part from #1196 because there were too many failures to triage, but we should get on the latest, even if it means we have to add a bunch of pending files to the config and move one bit at a time! Definitely help wanted!

rwwagner90:
@chancancode I plan to work on updating all the things to latest and greatest Octane and glimmer components, so tackling this would be a good first step. I'll do some of this in the coming week.

chancancode:
@rwwagner90 Nice! We can land any changes that you have incrementally too, so don't feel like you have to do it all at once.

PR #1217, opened 2020-05-14
Fix require-computed-property-dependencies lint rule

ansmonjol:
## Description

This MR aims to fix the future `ember/require-computed-property-dependencies` rule, while upgrading eslint-plugin-ember from 7.13.0 to 8.5.1.

See related issue: https://github.com/emberjs/ember-inspector/issues/1197

And the rule documentation: https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-computed-property-dependencies.md

Note that some error occurence from this build error list has been omitted: https://github.com/emberjs/ember-inspector/pull/1211/checks?check_run_id=672333609
They will be fixed by this PR: https://github.com/emberjs/ember-inspector/pull/1216

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,6 +32,7 @@ module.exports = {
 
     // Best practice
     'no-duplicate-imports': 'error',
+    'ember/require-computed-property-dependencies': 'error',
   },
   overrides: [
     // node files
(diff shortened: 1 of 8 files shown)


ansmonjol:
> Can you also enable the rule in the eslint config? 

Done, note that we still have 2 lint error related ot this change, until https://github.com/emberjs/ember-inspector/pull/1216 is merged.

PR #1219, opened 2020-05-16
Prevent future ember/no-mixins eslint error

ansmonjol:
## Description

This MR aims to fix the future `ember/no-mixins` error, while upgrading eslint-plugin-ember from 7.13.0 to 8.5.1.

Preferred to disable the lint on each occurence instead of disabling the rule globally, as we did for observers.

See related issue: https://github.com/emberjs/ember-inspector/issues/1197

And the rule documentation: https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-mixins.md

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,6 +32,7 @@ module.exports = {
 
     // Best practice
     'no-duplicate-imports': 'error',
+    'ember/no-mixins': 'error',
   },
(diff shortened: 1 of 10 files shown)


ansmonjol:
> Can you enable the rule

Of course ! Done

PR #1247, opened 2020-06-10
Update eslint-plugin-ember, fix lint

### LINKED ISSUE #1197: Upgrade to eslint-plugin-ember latest (v8)

rwwagner90:
This was basically just updating to latest eslint-plugin-ember and running `yarn lint:js --fix`. I did disable `no-get` for now because they didn't all fix, and I figured we could incrementally fix those later.

Closes #1197 
Closes #1216 
Closes #1217

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,6 +29,8 @@ module.exports = {
 
     // Temporarily turn these off
     'ember/avoid-leaking-state-in-ember-objects': 'off',
+    'ember/no-get': 'off',
+    'ember/no-test-import-export': 'off',
 
     // Best practice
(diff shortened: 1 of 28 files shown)


rwwagner90:
@nummi @chancancode mind reviewing this when you have a chance?

rwwagner90:
@chancancode I had not considered that, but I just did a quick scan myself and none of the `get` references changed seem to be something that would cause a problem. Definitely good to keep in mind for the future though!

PR #1256, opened 2020-06-18
Update ember-cli-version-checker

rwwagner90:

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -44,7 +44,7 @@ function resolveRelativeModulePath(name, child) {
 
 module.exports = function (defaults) {
   let checker = new VersionChecker(defaults);
-  let emberChecker = checker.forEmber();
+  let emberChecker = checker.for('ember');
 
   if (emberChecker.isAbove('3.0.0')) {
     options.vendorFiles = { 'jquery.js': null };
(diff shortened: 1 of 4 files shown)


ember-cli-build.js
```diff
@@ -44,7 +44,7 @@ function resolveRelativeModulePath(name, child) {
 
 module.exports = function (defaults) {
   let checker = new VersionChecker(defaults);
-  let emberChecker = checker.forEmber();
+  let emberChecker = checker.for('ember');
```
rwjblue, line 47: I think this needs to be `.for('ember-source')` right?

lib/ui/index.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
   // in the first place.
   treeForAddon() {
     let checker = new VersionChecker(this.project);
-    let emberChecker = checker.forEmber();
+    let emberChecker = checker.for('ember');
```
rwjblue, line 22: Same as above (`.for('ember-source')`)

ember-cli-build.js
rwwagner90: Ah, perhaps. I had seen `let dep = checker.for('ember', 'bower');` in the docs.

PR #1272, opened 2020-07-18
Move window resize to application template

nummi:
This is some cleanup work before my collapsible navigation PR. I moved the window resize event binding to the application template and controller. All objects that care about the window resizing can use the Layout Service:

``` javascript
this.layoutService.on('resize', this, this.doStuff)
```
----

# Questions
1. I debounced the window resize event at 250ms. Is this okay?
2. In other parts of the app, the code to trigger the Layout Service 'resize' event is scheduled to run in an `afterRender` job — is this necessary? Just to be consistent I did the same in the new `AppllcationController#windowDidResize` action.
3. Should we have a generic `MonitorContentHeight` ? Not sure we currently have a use case for it.


rwwagner90:
@nummi I believe the intent of the `afterRender` was too slightly debounce. Rather than infinitely calling, we make sure to only call if a render has occurred.

PR #1309, opened 2020-08-12
Fix some tests

rwwagner90:
## Description

It seems some of the ordering changed for properties in newer Ember versions. I made the tests not expect a specific order, but just expect the property to exist. There may also be other failures, but wanted to let CI run and see what happens.


rwwagner90:
Seems like yarn is having some trouble tonight

tests/ember_debug/object-inspector-test.js
```diff
@@ -659,15 +660,28 @@ module('Ember Debug - Object Inspector', function (hooks) {
...
+      'toString'
+    );
+    const hasChildren = A(message.details[3].properties).findBy(
+      'name',
+      'hasChildren'
+    );
+    const expensiveProperty = A(message.details[3].properties).findBy(
+      'name',
+      'expensiveProperty'
+    );
```
rwjblue, line 675: I think it would be easier to use `.find` directly.

```suggestion
    const toString = message.details[3].properties.find((p) => p.name === 'toString');
    const hasChildren = message.details[3].properties.find((p) => p.name === 'hasChildren');
    const expensiveProperty = message.details[3].properties.find((p) => p.name === 'expensiveProperty');
```

tests/ember_debug/object-inspector-test.js
```diff
@@ -659,15 +660,28 @@ module('Ember Debug - Object Inspector', function (hooks) {
...
+    assert.ok(toString);
+    assert.ok(hasChildren);
```
rwjblue, line 677: I know you didn't introduce this here (it was also wrong in the original version) but IMHO if you have more than 1 assertion in a test, all of the assertions **must** have a message (otherwise tracking down which one is failing is really annoying).

```suggestion
    assert.ok(toString, 'has toString');
    assert.ok(hasChildren, 'has hasChildren');
```

tests/ember_debug/object-inspector-test.js
```diff
@@ -757,14 +771,23 @@ module('Ember Debug - Object Inspector', function (hooks) {
...
-    assert.equal(message.details[3].properties[1].name, 'hasChildren');
+
+    const hasChildren = A(message.details[3].properties).findBy(
+      'name',
+      'hasChildren'
+    );
+    const expensiveProperty = A(message.details[3].properties).findBy(
+      'name',
+      'expensiveProperty'
+    );
```
rwjblue, line 782:
```suggestion
    const hasChildren = message.details[3].properties.find((p) => p.name === 'hasChildren');
    const expensiveProperty = message.details[3].properties.find((p) => p.name === 'expensiveProperty');
```

package.json
```diff
@@ -76,13 +76,13 @@
     "ember-did-resize-modifier": "^1.0.0",
     "ember-export-application-global": "^2.0.1",
     "ember-fetch": "^8.0.2",
-    "ember-flatpickr": "^2.16.3",
+    "ember-flatpickr": "^2.16.4",
```
rwjblue, line 79: This seems unrelated to the failures?

package.json
```diff
@@ -42,7 +42,7 @@
     "@glimmer/tracking": "^1.0.1",
     "@html-next/vertical-collection": "^1.0.0",
     "amd-name-resolver": "^1.3.1",
-    "aws-sdk": "^2.729.0",
+    "aws-sdk": "^2.731.0",
```
rwjblue, line 45: Seems unrelated?

package.json
rwwagner90: There was another failure from a patch version of flatpickr.

package.json
rwwagner90: Yes, this one is unrelated, but just bumped the couple deps that were not up to date.

tests/ember_debug/object-inspector-test.js
rwwagner90: @rwjblue is there a reason not to use `Ember.A`?

rwjblue: Ya, we don't need it 😃. Ember.A is about providing observeability, we aren't trying to observe and react to changes here. We just want to find a specific item. That's exactly what `.find` is for, and it results in the codebase being easier to follow (IMHO).

It is also quite likely that Ember.A itself will go away over time (in favor of specific mutation methods that themselves trigger the reactivity that we need), and in general we should prefer native methods for things whenever / where ever they exist.

rwwagner90:
@rwjblue changes made. Mind reviewing again?

PR #1375, opened 2020-10-11
WIP: in-element support

NullVoxPopuli:
## Description

Currently, as reported by  and reproduced in https://github.com/NullVoxPopuli/repro-repo-ember-inspector-1255,
when components are "portaled" to another part of the DOM, the inspector doesn't know how find them both with the pointer highlight toggle, and the right click menu -> Ember Inspect.

_This PR aims to fix that, *hopefully* meaning that *all* attempts to inspect a component "work"_

## Screenshots

TBD


rwwagner90:
Looks good so far! Thanks for picking this up.

PR #1400, opened 2020-11-14
Fix canary build

locks:
Due to the new deprecations introduced in the canary build, CI started failing.
This should address the problem.

This should unblock the open dependabot PRs.

### DIFF
--- a/ember_debug/libs/view-inspection.js
+++ b/ember_debug/libs/view-inspection.js
@@ -1,4 +1,4 @@
-import { classify } from '@ember/string';
+import classify from '../utils/classify';
 import bound from 'ember-debug/utils/bound-method';
 
 function makeHighlight(id) {
--- a/tests/ember_debug/deprecation-debug-test.js
+++ b/tests/ember_debug/deprecation-debug-test.js
@@ -38,13 +38,25 @@ module('Ember Debug - Deprecation', function (hooks) {
       Route.extend({
         setupController() {
           EmberDebug.IGNORE_DEPRECATIONS = false;
-          deprecate('Deprecation 1', false, { id: 'dep-1', until: '1.0.0' });
+          deprecate('Deprecation 1', false, {
+            id: 'dep-1',
+            for: 'ember-inspector',
+            since: '0.1.0',
+            until: '1.0.0',
+          });
           deprecate('Deprecation 2', false, {
             id: 'dep-2',
+            for: 'ember-inspector',
+            since: '0.1.0',
             until: '1.0.0',
             url: 'http://www.emberjs.com',
           });
-          deprecate('Deprecation 1', false, { id: 'dep-1', until: '1.0.0' });
+          deprecate('Deprecation 1', false, {
+            id: 'dep-1',
+            for: 'ember-inspector',
+            since: '0.1.0',
+            until: '1.0.0',
+          });
           EmberDebug.IGNORE_DEPRECATIONS = true;
         },
       })
@@ -93,8 +105,18 @@ module('Ember Debug - Deprecation', function (hooks) {
       Route.extend({
         setupController() {
           EmberDebug.IGNORE_DEPRECATIONS = false;
-          deprecate('Deprecation 1', false, { id: 'dep-1', until: '1.0.0' });
-          deprecate('Deprecation 2', false, { id: 'dep-2', until: '1.0.0' });
+          deprecate('Deprecation 1', false, {
+            id: 'dep-1',
+            for: 'ember-inspector',
+            since: '0.1.0',
+            until: '1.0.0',
+          });
+          deprecate('Deprecation 2', false, {
+            id: 'dep-2',
+            for: 'ember-inspector',
+            since: '0.1.0',
+            until: '1.0.0',
+          });
           EmberDebug.IGNORE_DEPRECATIONS = true;
         },
       })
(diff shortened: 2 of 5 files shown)


locks:
The extension doesn't like `@ember/string` 😬

ember_debug/libs/view-inspection.js
```diff
@@ -1,8 +1,6 @@
+import { classify } from '@ember/string';
 import bound from 'ember-debug/utils/bound-method';
 
-const Ember = window.Ember;
```
rwwagner90, line 3: I thought we had to use `window.Ember` so we can use the Ember from the app?

rwwagner90 (commented):
Seems okay, but I think we have to use `window.Ember` to use the Ember from the app. I don't think the manual imports from Ember tended to work before, but I could be wrong.

ember_debug/libs/view-inspection.js
rwjblue: This was only used for destructuring `classify`

package.json
```diff
@@ -30,6 +30,7 @@
     "@ember/edition-utils": "^1.2.0",
     "@ember/optional-features": "^2.0.0",
     "@ember/render-modifiers": "^1.0.2",
+    "@ember/string": "^1.0.0",
```
rwjblue, line 33: https://github.com/emberjs/ember-string/tree/v2.0.0 is tagged, is it not released to npm?

locks: Seems like I didn't publish. I don't remember if I had any specific reason not to 🤔 Maybe I was waiting for the toc deprecation?

rwwagner90 (approved):

rwwagner90:
LGTM! Feel free to merge when ready. Not sure if we wanted to get `@ember/string` 2.0.0 in as @rwjblue suggested.

locks:
I got the artifact from the "Build extensions" action and tried it out and I'm getting this error:

```
Uncaught Error: Module: '@ember/string' not found.
    at requireModule (<anonymous>:26:15)
    at requireModule (<anonymous>:38:24)
    at requireModule (<anonymous>:38:24)
    at requireModule (<anonymous>:38:24)
    at <anonymous>:9461:62
    at triggerOnce (<anonymous>:9521:7)
    at e.runLoadHooks (vendor-70026647264709751e5458c51d3ddb08.js:2033)
    at vendor-70026647264709751e5458c51d3ddb08.js:3491
    at o (vendor-70026647264709751e5458c51d3ddb08.js:102)
    at t (vendor-70026647264709751e5458c51d3ddb08.js:103)
```

Where did I goof? 😬

rwwagner90:
@locks I think we can't do imports. I don't fully understand how the build or the dependencies work, but since it runs in the context of the consuming app, I don't think our imports get shipped and added to that app. I believe that it why we have to use `window.Ember`.

tests/ember_debug/deprecation-debug-test.js
```diff
@@ -93,8 +105,18 @@ module('Ember Debug - Deprecation', function (hooks) {
       Route.extend({
         setupController() {
           EmberDebug.IGNORE_DEPRECATIONS = false;
-          deprecate('Deprecation 1', false, { id: 'dep-1', until: '1.0.0' });
-          deprecate('Deprecation 2', false, { id: 'dep-2', until: '1.0.0' });
+          deprecate('Deprecation 1', false, {
+            id: 'dep-1',
+            for: 'ember-inspector',
+            since: '0.1.0',
```
chancancode, line 111: I thought this has to be an object.. is that not the case? Other than that seems good to me

chancancode (approved):

chancancode:
> I got the artifact from the "Build extensions" action and tried it out and I'm getting this error

`ember_debug` doesn't automatically get any dependencies built into it, it's more or less just a concatenation of the files in the `ember_debug` folder. I think you can make it work, but you'll need to tweak the build to include any dependencies you need in here: https://github.com/emberjs/ember-inspector/blob/master/ember-cli-build.js#L102-L121

If we just need one utility function, it may end up being easier just vendoring our own version into `ember_deubg`.

> I think we can't do imports

It Just Works™ in `app/*`, like in any other ember apps. But for `ember_debug` it'll need more work.

rwwagner90:
@chancancode @locks is there a reason we can't use window.Ember anymore? Will vendoring the util mean we may be doing outdated things as the util evolves, since ours will be unchanging?

locks:
@rwwagner90 I am finishing the deprecation for `Ember.String` so I'm preemptively making sure the Inspector will not trigger said deprecation. The code has not changed in a couple years and is scheduled to be removed (the `@ember/string` package is available), so inlining should be ok.

rwwagner90:
> @rwwagner90 I am finishing the deprecation for `Ember.String` so I'm preemptively making sure the Inspector will not trigger said deprecation. The code has not changed in a couple years and is scheduled to be removed (the `@ember/string` package is available), so inlining should be ok.

Gotcha, feel free to inline then 👍

rwwagner90:
@locks should we go ahead and add the util to vendor, so we can get this in?

locks:
@rwwagner90 yeah, I'm going to inline the code later today :)

locks:
Build is green and I've tested the extension on Firefox and encountered no problems!

rwwagner90 (approved):

rwwagner90:
Thanks so much for doing this @locks 🎉

locks:
Glad to help!

PR #1405, opened 2020-11-17
update htmlSafe imports to the current package

locks:
## Description

Updates import paths for `htmlSafe` to use `@ember/template` instead of the deprecated `@ember/string` import.

Source: https://github.com/ember-cli/ember-rfc176-data/blob/master/mappings.json


rwjblue:
Currently failing due to https://github.com/emberjs/ember-inspector/issues/1406

rwwagner90:
Merged since the only failures were the expected ones.

PR #1445, opened 2021-01-03
Bypass unresolved route promises in route-debug

### LINKED ISSUE #1173: Error message: routeHandler.get is not a function
**Describe the bug**

Getting the following error while switching to the Routes tab of the inspector. No data is being shown on the Routes tab due to this error.

**To Reproduce**
1. Open Ember inspector 
2. Click on Routes
3. See the error `routeHandler.get is not a function` on the console.

**Screenshots**
image

image

**Environment**
Ember version: 3.4.8
Browser: Chrome 80.0.3987.163

steventsao:
## Description
This PR skips rendering a route when it is wrapped in an Engine that is not resolved. Since the route tree refreshes on route update, those unresolved engines will eventually be included in the pane, as shown in the screenshot.

[Another WIP](https://github.com/cyril-sf/ember-inspector/commit/ff5f26655130ba2eff61366f4bc39ce94185023f) adds Ember Engine to the inspector, but is incomplete and is nontrivial because the inspector has to know the implementation detail of an Engine.

My PR does not add the concept of an Engine but skips all unresolved Promises. This reenables the tab to show the routes in the right hierarchy and fresh data when the Promises are resolved. Let me know if you can consider merging this to unblock other users, ie https://github.com/emberjs/ember-inspector/issues/1173 since April.

## Screenshots
Screen Shot 2021-01-02 at 11 24 54 PM

Thanks!

Closes #1173


ember_debug/route-debug.js
```diff
@@ -168,6 +174,12 @@ function buildSubTree(routeTree, route) {
         // Ember < 3.9.0
         routeHandler = routerLib.getHandler(handler);
       }
+
+      // Skip when route is an unresolved promise
+      if (typeof routeHandler?.then === 'function') {
+        continue;
```
rwwagner90, line 180: @pzuraq @chancancode does this sound like a good fix or should we instead be waiting on the promises here?

steventsao: In my observation, `buildSubtree` keeps getting called until all the promises are settled anyway, so skipping or waiting for the engines in the initial calls go by faster than I can perceive the difference. I'm interested in your thoughts tho.

rwwagner90 (approved):

rwwagner90:
@pzuraq @chancancode does this seem like a good fix for now or what are your thoughts?

steventsao:
I spoke to @pzuraq, who is "not very familiar with the inspector," and so is not comfortable reviewing this.

Do you have bandwidth @chancancode ?

FYI @rwwagner90

rwwagner90:
@villander perhaps you have some insights here regarding engines?

steventsao:
Found a bug in changing engines. I'll fix it before opening another PR @rwwagner90

rwwagner90:
@steventsao was the bug you mentioned fixed?

steventsao:
@cyril-sf Can you let me know what visual change you were going with your [PR](https://github.com/cyril-sf/ember-inspector/commit/ff5f26655130ba2eff61366f4bc39ce94185023f)? Were you planning on adding an Engine type in a container that would show inside the route tab, or would routes inside an engine render the same way as those not contained in one?

Asking so I can either build on top of it or separately but with consistency as what you were going for. Thanks.

ember_debug/route-debug.js
rwjblue, line 180: > buildSubtree keeps getting called until all the promises are settled anyway

`buildSubTree` is definitely called many times (once for each route),

I think we should do something like:

```js
      // Skip when route is an unresolved promise
      if (typeof routeHandler?.then === 'function') {
        // ensure we rebuild the route tree when this route is resolved
        routeHandler.then(() => this.notifyPropertyChange('routeTree'));
        controllerName = '(unresolved)';
        controllerClassName = '(unresolved)';
        templateName = '(unresolved)';
      } else {
        controllerName =
          routeHandler.get('controllerName') || routeHandler.get('routeName');
        controllerFactory = owner.factoryFor
          ? owner.factoryFor(`controller:${controllerName}`)
          : owner._lookupFactory(`controller:${controllerName}`);
        controllerClassName = this.getClassName(controllerName, 'controller');
        templateName = this.getClassName(handler, 'template');
      }

      subTree[handler] = {
        value: {
          name: handler,
          routeHandler: {
            className: routeClassName,
            name: handler,
          },
          controller: {
            className: controllerClassName,
            name: controllerName,
            exists: !!controllerFactory,
          },
          template: {
            name: templateName,
          },
        },
      };
```

This ensures that the subtree structure itself is correct (and does indicate that _something_ is at this location / route name), it also ensures that a re-computation will happen when the promise resolves.

rwjblue (changes requested):
Generally speaking, this change is fine (and prevents the specific error that folks have reported). I do think that we should do _something_ more than `continue` though.

---

If you see here, `buildSubTree` is invoked from a computed property that only depends on `router`:

https://github.com/emberjs/ember-inspector/blob/d4f1fbb1ee30d178f81ce03bf8f037722bd4b166/ember_debug/route-debug.js#L60-L73

However, since `router` is also a CP that has no dependent keys that will force a re-computation (the `namespace.owner` dep key won't have an effect on re-running as routes are resolved) `buildSubTree` is not guaranteed to be called _again_ after the promise has resolved:

https://github.com/emberjs/ember-inspector/blob/d4f1fbb1ee30d178f81ce03bf8f037722bd4b166/ember_debug/route-debug.js#L23-L25

I've left an inline suggestion that _should_ ensure that `buildSubTree` is invoked again for this specific route.

ember_debug/route-debug.js
rwwagner90: Thanks for this detailed suggestion @rwjblue! @steventsao would you be interested in updating the PR accordingly?

steventsao: Done @rwwagner90

rwjblue (approved):

rwjblue:
Thank you!

steventsao:
Awesome, can't wait for the next release!

cc @rwwagner90

PR #1493, opened 2021-02-20
Change message type to switch to components route

rwwagner90:
@chancancode I have been unable to determine a way to programmatically open the devtools when clicking this context menu item, but this gets us one step closer by actually switching routes in inspector. Any ideas on how to open devtools to the Ember Inspector panel? I haven't been able to find anything about it.


rwwagner90:
@rwjblue @nummi  any thoughts on this? I don't recall if the `inspect ember component` context menu ever programmatically opened dev tools or if it just showed the component if you already had dev tools open?

pzuraq (approved):

PR #1509, opened 2021-03-05
Migrate to volta-cli/action

steventsao:
## Description

The Build and Publish action has been broken since last September. This PR fixes the [error](https://github.com/emberjs/ember-inspector/runs/2026942742?check_suite_focus=true
) in step `volta install chrome-webstore-upload-cli` by inserting `volta install node` before it.

## Screenshots
A sample build with the fix is able to move past `volta install chrome-webstore-upload-cli` to the upload step.
https://github.com/steventsao/ember-inspector/runs/2037371567?check_suite_focus=true

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,9 @@ jobs:
           echo "GIT_COMMIT_SHA=${GIT_COMMIT_SHA}" >> $GITHUB_ENV
           echo "GIT_BRANCH=${GIT_BRANCH}" >> $GITHUB_ENV
       - name: Set up Volta
-        uses: rwjblue/setup-volta@v1
+        uses: volta-cli/action@v1
+        with:
+          node-version: 12.x
       - name: Install dependencies (yarn)
         run: yarn install
       - name: Lint (hbs)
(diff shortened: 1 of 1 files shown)


steventsao:
@rwwagner90  @rwjblue 
Can you review this?

rwwagner90:
@steventsao I could be wrong, but I think the actual fix is to use `volta-cli/action` instead of `rwjblue/setup-volta` I *think* if we replace those with:

```
      - uses: volta-cli/action@v1
        with:
          node-version: 12.x
```

This might solve the problem, but @rwjblue please let me know if I am on the right track here.

rwwagner90:
@steventsao sorry to be picky, but could you replace all the old rwjblue ones in the file with the new one please?

steventsao:
> @steventsao I could be wrong, but I think the actual fix is to use `volta-cli/action` instead of `rwjblue/setup-volta` I _think_ if we replace those with:
> 
> ```
>       - uses: volta-cli/action@v1
>         with:
>           node-version: 12.x
> ```
> 
> This might solve the problem, but @rwjblue please let me know if I am on the right track here.

Thanks. I'm able to pass volta install with your recommendation too. https://github.com/steventsao/ember-inspector/runs/2041899466?check_suite_focus=true

steventsao:
> @steventsao sorry to be picky, but could you replace all the old rwjblue ones in the file with the new one please?

Done

rwwagner90 (approved):

rwwagner90:
Thanks so much for the PR @steventsao!

PR #1520, opened 2021-03-10
v4.3.5

steventsao:
## Description
Bump version to 4.3.5 so a release can be created with the tag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "compare-versions": "^3.5.0",
     "del": "^6.0.0",
     "ember-auto-import": "^1.10.1",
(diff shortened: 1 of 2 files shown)


steventsao:
@rwwagner90 Can you add a tag to this on merge? I hope it's automatic =p

rwwagner90:
@steventsao I think what you'll want to do here is run `release-it`, rather than manually updating the version. It will generate the changelog and tag as well.

steventsao:
> @steventsao I think what you'll want to do here is run `release-it`, rather than manually updating the version. It will generate the changelog and tag as well.

@rwwagner90 Thanks, done.

PR #1528, opened 2021-03-18
Run dependabot less often

sandstrom:
Running dependabot too often creates a lot of noise in the commit logs, making it nearly useless to track actual progress.

Running it less often would reduce this problem (security updates still running immediately).

*emberjs/ember.js has already made this change, though I think they configured it via the UI, which is obviously also an option (actually it's even better than doing it via a file IMO).*

See: https://github.com/emberjs/ember.js/pull/18979#issuecomment-631864842

image

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -10,4 +10,4 @@ update_configs:
   # other updates monthly, to cut down on repo noise.
   - package_manager: "javascript"
     directory: "/"
-    update_schedule: "monthly"
+    update_schedule: "weekly"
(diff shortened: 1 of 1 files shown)


rwwagner90:
I'm not sure I am in favor of this change. A lot happens in a month, and I would prefer not to wait a whole month to get new updates. I wonder if perhaps we should combine some of the dep updates into one PR instead of having so many? Not sure if something like that is possible or not.

rwjblue:
I think weekly would be better than "all the time", but I also think combining updates into a single commit would be good (this is something that renovate could do, but dependabot can't at the moment).

sandstrom:
@rwwagner90 @rwjblue updated to weekly

(if either of you have repo access I think you can set this directly on the org or repository)

Issue #1595, opened 2021-05-22
Uncaught TypeError: Cannot destructure property 'A' of '_ember.default' as it is undefined.


liberlanco:
**Describe the bug**
Exception in console: `Uncaught TypeError: Cannot destructure property 'A' of '_ember.default' as it is undefined.`

From time to time Ember also said that this is not an Ember application. I suppose because of this is error.

With more loaded application  (tailwind,  ember-cli-mirage, some components, some routes, some models and so on) it starts failing more often. When I did mirage mockups, and gives this error in 50-60% of refreshes.

**To Reproduce**
1. Install ember-cli 3.26.1,  and ember-inspector 4.4.0 in Google Chrome browser
2. Create new project and start it: `mkdir tmp-dir;  cd tmp-dir;  ember init;  npm start`
3. Open http://127.0.0.1:4200/  and manually add some non-existent url like http://127.0.0.1:4200/some/url
4. Open chrome inspector dev tools (F12)
5. Open console
6. Now spam Ctrl-R to refresh page. After time to time there will be exception:
```
VM4178:61 Uncaught TypeError: Cannot destructure property 'A' of '_ember.default' as it is undefined.
    at <anonymous>:61:5
    at requireModule (<anonymous>:42:28)
    at requireModule (<anonymous>:38:24)
    at <anonymous>:9574:62
    at triggerOnce (<anonymous>:9642:7)
    at runLoadHooks (lazy_load.js:75)
    at index.js:588
    at internalRequire (loader.js:54)
    at require (loader.js:87)
    at rsvp.js:2493
```

Research showed that it points to this line: https://github.com/emberjs/ember-inspector/blob/master/ember_debug/adapters/basic.js#L3   Ember object here could be undefined.

**Expected behavior**
No exception

**Environment**
ember 3.26.1
node-inspector 4.4.0
google chrome  90.0.4430.212

rwwagner90:
@pzuraq @chancancode is there a good way to ensure Ember is defined here? This seems to be related to switching to using `require` instead of the Ember global.

pzuraq:
@rwwagner90 I'm not exactly sure how that could be `undefined`, it's either loaded via the module or via `window.Ember`: https://github.com/emberjs/ember-inspector/blob/master/ember_debug/utils/ember.js

The only way this seems like it could cause this error is if `window.Ember` is `undefined`, so it's not an Ember application at all. In that case, we may need better logic to prevent the inspector from loading. 

[...]

PR #1601, opened 2021-05-25
Ensure vendored loader does not cache missing modules.

rwjblue:
Prior to this change, calling `requireModule('foo')` would **create** a `foo` module (with no exports) due to eager caching in the list of "seen" modules.

This means that code like this:

```js
let Ember;
try {
  Ember = requireModule('ember')['default'];
} catch {
  Ember = window.Ember;
}
return Ember;
```

Will properly fallback to `window.Ember` the first time, but all subsequent runs will end up finding a module and returning `{}`.

This addresses at least _part_ of https://github.com/emberjs/ember-inspector/issues/1595 (cannot confirm if it is fully resolved, due to raciness and testing timing).

### DIFF
--- a/ember_debug/vendor/loader.js
+++ b/ember_debug/vendor/loader.js
@@ -17,13 +17,13 @@ if (typeof define !== 'function' || typeof requireModule !== 'function') {
       if (seen[name]) {
         return seen[name];
       }
-      seen[name] = {};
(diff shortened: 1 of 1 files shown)


rwwagner90 (approved):
LGTM!

PR #1578, opened 2021-05-12
Fix bug that prevents new change logs from showing

hannakim91:

PR #1619, opened 2021-06-09
Move changelog heading to top

sandstrom:
Format also changed, so that each changelog entry sub-heading is no longer a link. But I think that makes sense actually, so I haven't modified that.

There is a separate PR that has updated the regex to detect this new sub-heading format.

Somewhat related: https://github.com/emberjs/ember-inspector/pull/1578

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# Changelog
+
 ## v4.4.1 (2021-05-25)
 
 #### :bug: Bug Fix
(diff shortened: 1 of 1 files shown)


rwwagner90 (approved):

rwwagner90:
Thanks @sandstrom! We had some issues getting tests to pass in #1578. Would you have some time to take a look?

sandstrom:
@rwwagner90 Thanks, I answered inline, in #1578.

PR #1683, opened 2021-08-30
Remove toolbar outlet, remove more global Ember

rwwagner90:
## Description

Ember 4 does not allow named outlets, so this refactors one of our named outlets to components instead.


rwwagner90:
@rwjblue @pzuraq I'm getting some errors on beta and canary saying Ember is undefined. I think I have removed all the global Ember now. Any ideas what else would be causing it?

app/routes/application.js
```diff
@@ -1,8 +1,7 @@
 import { inject as service } from '@ember/service';
 import { set, get, action } from '@ember/object';
 import Route from '@ember/routing/route';
-import Ember from 'ember';
-
+const Ember = require('ember')['default'];
```
sly7-7, line 4: I just wonder why this move ? Isn't the recommended way to `import Ember from 'ember` ?

rwwagner90: @pzuraq I thought `import Ember from 'ember'` counted as global?

PR #1685, opened 2021-09-03
add highlight render option to Render Performance tab

michaelbdai:
## Description
Added a checkbox in the render performance tab/route
When enabled, the component gets highlighted when rends.
If the component has no child component, highlight is red, otherwise it is blue.
See more details in screen recording.

## Screenshots
Here is the screen recording and unpacked build
https://drive.google.com/drive/folders/1OSOYbIj1jn7hrUs7EYfUqQFg-6HXpqiM?usp=sharing


nummi:
It's beautiful!

ember_debug/render-debug.js
```diff
@@ -35,9 +36,37 @@ export default EmberObject.extend(PortMixin, {
...
+      })
+    }
+  },
+
+  _createOutline(viewGuid, hasChildren) {
+    const element = document.getElementById(viewGuid);
+    if (element) {
+      const outline = element.style.outline;
+      element.style.outline  = `${hasChildren ? '0.5' : '1'}px solid ${hasChildren ? 'blue' : 'red'}`;
+      setTimeout(()=> {
```
lifeart, line 64: could we clear timeout in case of more than one rerender will happen in 1s?

ember_debug/render-debug.js
```diff
@@ -35,9 +36,37 @@ export default EmberObject.extend(PortMixin, {
...
+    }
+    if (hasChildren) {
+      children.forEach(childNode => {
+        this._hightLightNode(childNode);
+      })
+    }
+  },
+
+  _createOutline(viewGuid, hasChildren) {
+    const element = document.getElementById(viewGuid);
```
lifeart, line 60: as I understand it wll work only with ember components with element property?

ember_debug/render-debug.js
```diff
@@ -35,9 +36,37 @@ export default EmberObject.extend(PortMixin, {
...
+    if (element) {
+      const outline = element.style.outline;
+      element.style.outline  = `${hasChildren ? '0.5' : '1'}px solid ${hasChildren ? 'blue' : 'red'}`;
+      setTimeout(()=> {
+        element.style.outline  = outline && 'none';
```
lifeart, line 65: should it be `outline ?? 'none'`?

ember_debug/render-debug.js
michaelbdai: Good question and thank you for the feedback. 
It should work for all the ember component, which should have `id="ember101"`. We do not need highlight the other dom which is within the ember component.
However, can you tell me in which case ember component does not have element property?

ember_debug/render-debug.js
michaelbdai: I think it is ok and normal if the highlights of different components happen at the same time. React dev tool has the same behavior which seems to be pretty good.
However, you remind me that we can have some special highlight if the same component got re-rendered within short time, 300ms for example. I can add that feature in the future change.

ember_debug/render-debug.js
michaelbdai: updated.

ember_debug/render-debug.js
lifeart: it's about glimmer component itself, or ember component with element = '';

we could get roots of glimmer component using renderTree api (already used in ember-inspector), example of how we could find roots of glimmer component <https://github.com/microsoft/playwright/pull/8444/files#diff-758a9d9fc7482116b104ad9be2c48f82e39f039e600d571afb3c7f65f6ae213bR142>

michaelbdai: I looked into the renderTree API. From what I understand, it can match the node and highlight it.
However, render-debug use the Ember Instrumentation API. https://api.emberjs.com/ember/2.14/classes/Ember.Instrumentation
https://github.com/emberjs/ember-inspector/blob/master/ember_debug/render-debug.js#L74
Instrumentation provides a payload for the component which got rendered. The payload does not have the exact dom.
In the other word, we do not get the newly rendered node directly from the browser. I take a screenshot of the console log of the payload
Screen Shot 2021-09-18 at 12 10 21 AM
I do see the "element" key you are talking about. Do you think we can match the component in the renderTree using the payload? Does Ember has any util can convert to node?

ember_debug/render-debug.js
michaelbdai: Great point, I will make the change.

ember_debug/models/profile-manager.js
```diff
@@ -2,9 +2,84 @@ import ProfileNode from './profile-node';
...
+  if (start === null || end === null)
+    return [];
+
+
+  for (let i = start; i <= end; i++)
+    roots.push(closest.item(i));
+
+
+  return roots.filter((el) => {
+      if (el.nodeType === 3) {
```
lifeart, line 32: likely we should mention "higligtable" node types here

ember_debug/models/profile-manager.js
```diff
@@ -2,9 +2,84 @@ import ProfileNode from './profile-node';
...
+          if (el.nodeValue.trim() === '') {
+              return false;
+          }
+      }
+      return el;
+  })
+}
+
+function makeHighlight(id) {
```
lifeart, line 41: I don't think we need Id here..
we could use Weak map to match higlight node with one of component roots;

ember_debug/models/profile-manager.js
```diff
@@ -2,9 +2,84 @@ import ProfileNode from './profile-node';
...
+  return `<div id="ember-inspector-render-highlight-${id}" role="presentation" class="ember-inspector-render-highlight"></div>`;
```
lifeart, line 42:
```js
let node= document.createElement('div');
node.setAttribute('role', 'presentation');
node.classList = 'ember-inspector-render-higlight');
```

ember_debug/render-debug.js
michaelbdai: changed.

ember_debug/render-debug.js
michaelbdai: nice to meeting. I changed the highlight logic as you suggested.

ember_debug/models/profile-manager.js
michaelbdai: updated.

ember_debug/models/profile-manager.js
michaelbdai: thanks you for the suggest. updated.

ember_debug/models/profile-manager.js
```diff
@@ -5,6 +5,74 @@ const {
...
+  if (start === null || end === null) return [];
+
+  for (let i = start; i <= end; i++) roots.push(closest.item(i));
+
+  return roots;
+}
+
+function _findRoots({ first, last, parent }) {
+  const closest = parent.childNodes;
+  if (first.node === last.node) return [first.node];
```
lifeart, line 26: looks like here we also should check for `nodeType === 1` and return empty array if not matched

ember_debug/models/profile-manager.js
```diff
@@ -45,6 +118,16 @@ export default class ProfileManager {
     return callback.call(context);
   }
 
+  renderHighLight(view) {
+    const symbols = Object.getOwnPropertySymbols(view);
+    const bounds = view[symbols.find((sym) => sym.description === 'BOUNDS')];
```
lifeart, line 123: likely we should be shure that bounds is not undefined

ember_debug/models/profile-manager.js
michaelbdai: updated.

ember_debug/models/profile-manager.js
```diff
@@ -5,6 +5,74 @@ const {
...
+  const { top, left, width, height } = rect;
+  const { scrollX, scrollY } = window;
+  const { style } = highlight;
+  if (style) {
+    style.position = 'absolute';
+    style.top = `${top + scrollY}px`;
+    style.left = `${left + scrollX}px`;
+    style.width = `${width}px`;
+    style.height = `${height}px`;
+    style.zIndex = `1000000`;
```
lifeart, line 70: could we have `z-index` and `position` values belongs to `ember-inspector-render-highlight` class definition? (it will be same for every node we have)

ember_debug/models/profile-manager.js
```diff
@@ -5,6 +5,74 @@ const {
...
+  return node;
+}
+function _insertHTML() {
+  return document.body.appendChild(makeHighlight());
+}
+
+function _insertStylesheet() {
+  const content = `
+    .ember-inspector-render-highlight {
+      border: 1px solid red;
```
lifeart, line 46: for multiple rerender cases (to get it more visible), I assume we could have box-shadow with .5 opacity (to allow layers multiply)

ember_debug/models/profile-manager.js
```diff
@@ -80,6 +163,10 @@ export default class ProfileManager {
     }
   }
 
+  teardown() {
+    this.stylesheet.remove();
```
lifeart, line 167: likely we should remove all "higlight" nodes too here (if exists)

lifeart (changes requested):

ember_debug/models/profile-manager.js
michaelbdai: updated.

ember_debug/models/profile-manager.js
michaelbdai: updated.

ember_debug/models/profile-manager.js
michaelbdai: changed to
`border: 2px solid rgba(255,0,0,0.2);
      box-shadow: 0px 0px 1px rgba(255,0,0,0.2);`

ember_debug/models/profile-manager.js
michaelbdai: updated.

ember_debug/models/profile-manager.js
michaelbdai: Added highlightComponents to track the active highlights.

ember_debug/models/profile-manager.js
```diff
@@ -80,6 +162,47 @@ export default class ProfileManager {
...
+  _renderHighlight(node, guid) {
+    if (!node?.getBoundingClientRect) {
+      return;
+    }
+    let id = guid;
+    // if guid does not exist for payload.view
+    // or payload.view is correlated to more than one component
+    // create random id to avoid collision.
+    if (!guid || this.highlightComponents[guid]) {
+      id = (Math.random() * 100000000).toFixed(0);
```
lifeart, line 183: `Math.random().toString(36).slice(2)` return quite neat cache

ember_debug/models/profile-manager.js
```diff
@@ -80,6 +162,53 @@ export default class ProfileManager {
     }
   }
 
+  teardown() {
+    this.stylesheet.remove();
+    // remove all the active highlighted components
+    for (const key in this.highlightComponents) {
+      const elementId = getNodeId(key);
```
lifeart, line 169: seems like we should have ability to cancelTimeouts here, to avoid some side-effects

Looks like we could have array-based higlight manager:

```js
higlights = [];
removeHiglight(el: { timeout: number, el: DOMNode }) {
   higlights = higlights.filter(item => item !== el);
  cancelTimout(el.timeout);
  el.node.remove();
}
addHiglight(el) {
   this.higlights.push(el);
   el.timeout = setTimout(...);
  // ...
}
```

lifeart (changes requested):

ember_debug/models/profile-manager.js
michaelbdai: updated.

ember_debug/models/profile-manager.js
michaelbdai: updated.

ember_debug/models/profile-manager.js
```diff
@@ -5,6 +5,63 @@ const {
...
+}
+
+function makeHighlight() {
+  const node = document.createElement('div');
+  node.setAttribute('role', 'presentation');
+  node.setAttribute('class', 'ember-inspector-render-highlight');
+  return node;
+}
+function insertHTML() {
+  return document.body.appendChild(makeHighlight());
```
lifeart, line 47: let's get makeHighlight() result as function argument here (to store DOM reference in object for destroying)

ember_debug/models/profile-manager.js
```diff
@@ -80,6 +156,59 @@ export default class ProfileManager {
...
+    this.highlights = this.highlights.filter((item) => item !== el);
+    clearTimeout(el.timeout);
+    el.node.remove();
+  }
+
+  addHighlight(highlight) {
+    this.highlights.push(highlight);
+
+    highlight.timeout = setTimeout(() => {
+      highlight.el.remove();
```
lifeart, line 182: I think we could call this.removeHiglight() here (to not duplicate removal logic)

ember_debug/models/profile-manager.js
```diff
@@ -80,6 +156,59 @@ export default class ProfileManager {
...
+    }, 1000);
+  }
+
+  _constructHighlight(node) {
+    const rect = node.getBoundingClientRect();
+    const highlight = insertHTML();
```
lifeart, line 188: ref: https://github.com/emberjs/ember-inspector/pull/1685/files#r717414371

```js
const highlight  = makeHighlight();
```

ember_debug/models/profile-manager.js
```diff
@@ -80,6 +156,59 @@ export default class ProfileManager {
...
+
+  removeHighlight(el) {
+    this.highlights = this.highlights.filter((item) => item !== el);
+    clearTimeout(el.timeout);
+    el.node.remove();
+  }
+
+  addHighlight(highlight) {
+    this.highlights.push(highlight);
+
```
lifeart, line 180: and likely here, we should call `_insertHighlight` method, to inject item into DOM

ember_debug/models/profile-manager.js
michaelbdai: updated.

ember_debug/models/profile-manager.js
michaelbdai: updated.

ember_debug/models/profile-manager.js
michaelbdai: updated.

tests/ember_debug/profile-manager-test.js
```diff
@@ -1,8 +1,138 @@
...
+      first: {
+        node: simpleComponent,
+      },
+      last: {
+        node: simpleComponent,
+      },
+    };
+    const payload = {
+      view: {
+        [Symbol('BOUNDS')]: bounds,
```
lifeart, line 113: could we get real payload here (to be sure we match exsting implemetation once changed)

tests/ember_debug/profile-manager-test.js
```diff
@@ -1,8 +1,203 @@
...
+    assert.equal(getRounded(simpleStyle.height), getRounded(simpleExpected.height), 'same height as component');
+
+    const anotherStyle = anotherHighlight.style;
+
+    const anotherExpected = anotherComponent.getBoundingClientRect();
+    assert.ok(anotherHighlight, 'anotherHighlight is visible');
+    assert.equal(getRounded(anotherStyle.left), getRounded(anotherExpected.x), 'same x as component');
+    assert.equal(getRounded(anotherStyle.top), getRounded(anotherExpected.y), 'same y as component');
+    assert.equal(getRounded(anotherStyle.width), getRounded(anotherExpected.width), 'same width as component');
+    assert.equal(getRounded(anotherStyle.height), getRounded(anotherExpected.height), 'same height as component');
```
lifeart, line 201: could we add await for 1s to ensure DOM nodes is removed? and check it' using getElementsByClassName one more time

michaelbdai: not as easy as it sounds, but done.

tests/ember_debug/profile-manager-test.js
michaelbdai: great idea, done.

tests/ember_debug/profile-manager-test.js
```diff
@@ -1,8 +1,252 @@
...
+    // enable highlight
+    run(() =>
+      EmberDebug.port.trigger('render:updateShouldHighlightRender', {
+        shouldHighlightRender: true,
+      })
+    );
+
+    const newHighlights = [];
+
+    const observer = new MutationObserver(function (records) {
```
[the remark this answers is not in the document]
michaelbdai: I don't think this is a good approach. We expect to have two highlights, so it is possible observer can be triggered twice. This code will stop watching once the first highlight is rendered. Additionally, addedNodes has two elements. I wonder what does it mean when `resolve(node)` get called twice within the promise.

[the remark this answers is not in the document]
michaelbdai: updated.

tests/ember_debug/profile-manager-test.js
```diff
@@ -1,8 +1,246 @@
...
+        classNames: ['simple-component'],
+        toString() {
+          return 'App.TestFooComponent';
+        },
+      })
+    );
+
+    this.owner.register(
+      'component:test-bar',
+      EmberComponent.extend({
```
lifeart, line 101: do we have ability to register glimmer components here? likely we should check how it will work with multiple roots (1+ higlight nodes should exist)

tests/ember_debug/profile-manager-test.js
```diff
@@ -1,8 +1,246 @@
...
+    this.owner.register(
+      'template:components/test-foo',
+      hbs('test-foo', {
+        moduleName: 'my-app/templates/components/test-foo.hbs',
+      })
+    );
+    this.owner.register(
+      'template:components/test-bar',
+      hbs(
+        '<!-- before --><div class="another-component"><span>test</span> <span class="bar-inner">bar</span></div><!-- after -->',
```
lifeart, line 140: could we have  tagless ember component with 2 roots?
```js
{
 tagName: ''
}
template: `<div>root one</div> <div>root two</div>`
```

tests/ember_debug/profile-manager-test.js
```diff
@@ -1,8 +1,246 @@
...
+
+    const simpleComponent = find('.simple-component');
+    const anotherComponent = find('.another-component');
+
+    const [simpleHighlight, anotherHighlight] = newHighlights;
+
+    const simpleStyle = simpleHighlight.style;
+    const simpleExpected = simpleComponent.getBoundingClientRect();
+    assert.ok(simpleHighlight, 'simpleHighlight is visible');
+    assert.equal(
```
lifeart, line 191: let's have one unified asser functon for this two cases (likely we will add more cases to check)

```js

function assertNodeSizes(synthetic, real) {
    const style = synthetic.style;
    const box = real.getBoundingClientRect();
    const dimensions = [
        ['left', 'x'],
        ['top', 'y'],
        ['width', 'width'],
        ['height', 'height']
    ];
    for (const [styleKey, boxKey] of dimensions) {
        assert(
            getRounded(style[styleKey]),
            getRounded(box[boxKey]),
            `same ${boxKey} as component`
        )
    }
}
```

tests/ember_debug/profile-manager-test.js
```diff
@@ -1,8 +1,246 @@
...
+        records.forEach((record) => {
+          record.addedNodes.forEach((node) => {
+            if (node.className === 'ember-inspector-render-highlight') {
+              observedHighlights.push(node);
+            }
+          });
+        });
+      });
+      observer.observe(document.body, { childList: true });
+      await visit('/simple');
```
lifeart, line 175: likely this async action should be an function argument (for comatibility)

```js
async function highlightsPromise(task) {
/// ...
 await task();
/// ...
}

highlightsPromise(() => visit('/simple'));

File: tests/ember_debug/profile-manager-test.js [hunk 25]
```diff
@@ -1,8 +1,246 @@
...
+      await waitUntil(() => observedHighlights.length === 2, { timeout: 2000 });
```
lifeart, line 176: same hrere, amount of items should be an argument

tests/ember_debug/profile-manager-test.js
```diff
@@ -1,8 +1,246 @@
...
+    await visit('/posts');
+
+    // enable highlight
+    run(() =>
+      EmberDebug.port.trigger('render:updateShouldHighlightRender', {
+        shouldHighlightRender: true,
+      })
+    );
+
+    async function highlightsPromise() {
```
lifeart, line 163: I think it's safe to move function to the top (outside of test declaration)

tests/ember_debug/profile-manager-test.js
```diff
@@ -1,8 +1,246 @@
...
+    );
+
+    this.owner.register(
+      'component:test-bar',
+      EmberComponent.extend({
+        tagName: '',
+        toString() {
+          return 'App.TestBarComponent';
+        },
+      })
```
lifeart, line 106: Cases to check:

`rootNode` = `<div></div>`
`commentNode` = `<!-- comment -->`
`textNode` = `text`

## Ember:

- [ ] ember component with [textNode] and tagName = '';
- [ ] ember component with [commentNode] and tagName = '';
- [ ] ember component with [rootNode]  and tagName = '';
- [ ] ember component with [rootNode, rootNode] and tagName = '';
- [ ] ember component with [rootNode, commentNode, rootNode] and tagName = '';
- [ ] ember component with [commentNode, rootNode, commentNode] and tagName = '';
- [ ] ember component with [rootNode, rootNode] and tagName = '';
- [ ] ember component with tagName = 'div'

## Glimmer:

- [ ] glimmer component with [textNode];
- [ ] glimmer component with [commentNode];
- [ ] glimmer component with [rootNode];
- [ ] glimmer component with [rootNode, rootNode];
- [ ] glimmer component with [rootNode, commentNode, rootNode];
- [ ] glimmer component with [commentNode, rootNode, commentNode];
- [ ] glimmer component with [rootNode, rootNode];

MelSumner:
@rwwagner90 please review :)

rwwagner90:
@michaelbdai now that master has passing tests, can you please rebase?

michaelbdai:
@rwwagner90 rebased.

rwwagner90:
Thanks again so much for the PR! 🎉

PR #1706, opened 2021-09-18
Possible fix for ember 3.26+

### LINKED ISSUE #1634: Application that accesses global Ember fails to start after browser refresh
**Describe the bug**
If an Ember application accesses the global `Ember` and the ember-inspector plugin is active, then the application fails to start after a browser refresh.

**To Reproduce**
1. start the application [refresh-inspector](https://github.com/pieter-v/refresh-inspector). This is a simple application that renders the global `Ember.VERSION`
2. visit the application with chrome
3. open the ember-inspector 
3. refresh the browser
4. stacktrace in the console and nothing is rendered:
```
lazy_load.js:75 Uncaught RangeError: Maximum call stack size exceeded.
    at runLoadHooks (lazy_load.js:75)
    at Module.callback (index.js:520)
    at Module.exports (loader.js:106)
    at requireModule (loader.js:27)
    at r (loader.js:176)
    at get (index.js:10)
    at get (index.js:41)
    at <anonymous>:9649:7
    at runLoadHooks (lazy_load.js:75)
    at Module.callback (index.js:520)
```

**Expected behavior**
The application should render successfully

**Environment**
ember=3.27.5 (there is no problem with 3.26)
chrome=91.0.4472.106
os=macos 11.4

**Additional context**
We still need some addons that access the global Ember, so this is for us a nasty bug

tben:
## Description
Not sure if this is the best way to do this. However, since Ember doesn't appear to have global scope any more on ember 3.26+. This moves the variable to global scope, so it works within onApplicationStart again.

## Screenshots
image

closes #1634

### DIFF
--- a/ember_debug/vendor/startup-wrapper.js
+++ b/ember_debug/vendor/startup-wrapper.js
@@ -19,6 +19,8 @@ if (typeof env !== 'undefined') {
   currentEnv = env;
 }
 
+let Ember;
+
 // @formatter:off
 var EMBER_VERSIONS_SUPPORTED = {{EMBER_VERSIONS_SUPPORTED}};
 // @formatter:on
@@ -97,7 +99,6 @@ var EMBER_VERSIONS_SUPPORTED = {{EMBER_VERSIONS_SUPPORTED}};
         return;
       }
 
-      let Ember;
       try {
         Ember = requireModule('ember')['default'];
       } catch {


rwwagner90:
@tben thanks for taking a look at this! It looks like the tests are failing, so I am not sure if this is the correct fix.

PR #1714, opened 2021-09-20
Try to fix Ember

rwwagner90:


kdagnan:
Any updates on this? We're holding off on upgrading our apps because the inspector breaks with 3.27/3.28.

rwwagner90:
@kdagnan no updates yet, we need support from some Ember core team members to help us figure out all the things we need to update. I think 3.27 and 3.28 should be fine, just not the 4.0 betas.

PR #1721, opened 2021-10-02
Make Private/Underscore Properties visible

### LINKED ISSUE #1426: Make Private/Underscore Properties visible
I understand Ember Inspector is for the development and Debugging phase. And I am of the opinion that, having access to the private (by convention, underscore prefixed) fields/property is a very useful feature when hunting for problems like why a certain computed property based on other field and property have an unexpected value.

**Describe the solution you'd like**
Having the private properties visible in the Ember inspector would be a great feature when debugging, especially when other computed property depends on multiple 

**Describe alternatives you've considered**
I tend to make all my field, public (so to say) to be able to test and play around with certain values to profile the behavior of certain parts of the application. In some cases where you have similar property names for internal use, this can lead to confusing which property is meant to be used to pass value to a control. Prefixing everything after debugging is more error prone.

**Additional context**
I understand I could set breakpoints in the source code, but this can in some cases be very time consuming and cumbersome even more so when a component is used in iteration.

IgorKvasn:
Closes #1426

I wanted to hide all private Ember properties (e.g. `_currentState`, `__ARGS__ember1633159271771385348405711__`) so that they won't clutter the property list, but I am not sure if my approach to enumerate them is ok.


rwwagner90:
Hi @IgorKvasn can you please rebase your PR? We updated master to get all the tests passing again.

chancancode (approved):
Seems good to me. Going forward, with Ember 4.0, we should change upstream so it uses real symbols for these so we won't have to special case these anymore.

PR #1782, opened 2021-11-08
Update Chrome Publish Action to Use New Command

kdagnan:
The chrome-webstore-upload-cli package updated its command from "webstore" to "chrome-webstore-upload".

This should allow new versions to be published to the webstore.


chancancode (approved):
Thanks!

PR #1784, opened 2021-11-13
bump to ember 4.0

michaelbdai:
## Description
bump ember-cli to 4.0 beta2

## Screenshots
N/A, all test passed


package.json
```diff
@@ -88,7 +88,7 @@
     "ember-template-lint": "^3.6.0",
     "ember-test-selectors": "^6.0.0",
     "ember-truth-helpers": "^3.0.0",
-    "ember-try": "^2.0.0",
+    "ember-try": "^2.0.0-beta.2",
```
rwwagner90, line 91:
```suggestion
    "ember-try": "^2.0.0",
```

rwwagner90 (commented):
@michaelbdai shouldn't we run ember-cli-update here? I think we probably need to fix all the lint rules that are disabled before bumping to Ember 4 anyway.

PR #1793, opened 2021-11-17
Remove ember-did-resize-modifier dep (for now?)

chancancode:
Currently the addon is breaking the build. The code and purpose of the addon is small enough that we can just vendor it. If the addon is fixed eventually we can change it back.

See also https://github.com/gmurphey/ember-did-resize-modifier/pull/374


rwwagner90 (approved):

Turbo87:
looks like this might have caused #1797 🤔

PR #1803, opened 2021-11-18
Replace ember-did-resize-modifier with ember-on-resize-modifier

### LINKED ISSUE #1797: 4.5.2 seems to be broken
**Describe the bug**
Ember inspector addon remains white when opening in chrome. In addons console I see this error:

```
Uncaught Error: Could not find module `element-resize-detector` imported from `ember-inspector/modifiers/did-resize`
```

**To Reproduce**
1. Open ember inspector

**Expected behavior**
The ember inspector works as usual 

**Screenshots**
image
(Right console is that from the addon)

**Environment**
Ember version: 3.20.7
Browser version: Version 95.0.4638.69 (Official Build) (64-bit)
Ember-inspector: 4.5.2

**Additional context**
-

rwwagner90:
Fixes #1797


NullVoxPopuli:
This is a temporary fix and dodges the root of the actual issue.
until chunks are supported, any package from NPM will cause a runtime error due to publicAssetURL not being correct :thinking: 
(and the chunk assets 404ing).

It'd be great of the build had an option to force everything into one file though.
ember-inspector doesn't need bundle splitting.

rwwagner90:
@NullVoxPopuli setting `publicAssetURL` had no effect for me, and I don't think that is the fix. My guess is either we are not shipping all the built files or the url generation for assets in the app is wrong.

PR #1833, opened 2021-12-12
fix issue, test destroyed the component before EmberDebugger is destr…

michaelbdai:
## Description
fix issue, test destroyed the component before EmberDebugger is destroyed

## Screenshots
Screen Shot 2021-12-11 at 11 31 07 PM
Screen Shot 2021-12-11 at 11 32 03 PM


rwwagner90 (approved):
Thanks so much @michaelbdai!

PR #1906, opened 2022-02-09
octane migration - deprecation-item-source

michaelbdai:

### DIFF
--- a/app/components/deprecation-item-source.hbs
+++ b/app/components/deprecation-item-source.hbs
@@ -1,1 +1,28 @@
-{{yield this}}
+<div class="deprecation-source pb-2" data-test-deprecation-source>
+  <button
+    class="send-to-console send-to-console--chevron-only mr-3"
+    title="Trace deprecations in console"
+    data-test-trace-deprecations-btn
+    onclick={{this.handleClick}}
+    type="button"
+  >
+    {{svg-jar "send-with-chevron" width="6px" height="9px"}}
+    Trace in the Console
+  </button>
+  <span class="source">
+    {{#if this.isClickable}}
+      <a
+        class="font-mono"
+        href="#"
+        data-test-deprecation-source-link
+        {{on "click" this.handleRedirect}}
+      >
+        {{this.url}}
+      </a>
+    {{else}}
+      <span class="font-mono" data-test-deprecation-source-text>
+        {{this.url}}
+      </span>
+    {{/if}}
+  </span>
+</div>
--- a/app/components/deprecation-item-source.js
+++ b/app/components/deprecation-item-source.js
@@ -1,26 +1,37 @@
-import { tagName } from '@ember-decorators/component';
-import { computed, get } from '@ember/object';
+import { get, action } from '@ember/object';
 import { inject as service } from '@ember/service';
-import { and, readOnly, bool } from '@ember/object/computed';
-import Component from '@ember/component';
-
-@tagName('')
+import Component from '@glimmer/component';
 export default class DeprecationItemSource extends Component {
   @service port;
 
-  @bool('model.map.source') known;
-
-  @computed('model.map.{line,source}', 'known')
   get url() {
-    let source = get(this, 'model.map.source');
+    const source = get(this.args, 'itemModel.map.source');
     if (this.known) {
-      return `${source}:${get(this, 'model.map.line')}`;
+      return `${source}:${get(this.args, 'itemModel.map.line')}`;
     } else {
       return 'Unkown source';
     }
   }
 
-  @readOnly('port.adapter') adapter;
+  get adapter() {
+    return this.port.adapter;
+  }
 
-  @and('known', 'adapter.canOpenResource') isClickable;
+  get isClickable() {
+    return this.known && this.adapter.canOpenResource;
+  }
+
+  get known() {
+    return this.args.itemModel.map.source;
+  }
+
+  @action
+  handleClick() {
+    this.args.traceSource?.(this.args.modelGroup, this.args.itemModel);
+  }
+
+  @action
+  handleRedirect() {
+    this.args.openResource?.(this.args.itemModel.map);
+  }
 }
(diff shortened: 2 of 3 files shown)


app/components/deprecation-item-source.js
```diff
@@ -1,26 +1,37 @@
...
+import Component from '@glimmer/component';
 export default class DeprecationItemSource extends Component {
   @service port;
 
-  @bool('model.map.source') known;
-
-  @computed('model.map.{line,source}', 'known')
   get url() {
-    let source = get(this, 'model.map.source');
+    const source = get(this.args, 'itemModel.map.source');
```
locks, line 8:
```suggestion
    let source = this.itemModel.map.source;
```

app/components/deprecation-item-source.js
```diff
@@ -1,26 +1,37 @@
...
     if (this.known) {
-      return `${source}:${get(this, 'model.map.line')}`;
+      return `${source}:${get(this.args, 'itemModel.map.line')}`;
```
locks, line 10:
```suggestion
      return `${source}:${this.args.itemModel.map.line}`;
```

app/components/deprecation-item-source.js
```diff
@@ -1,26 +1,37 @@
-import { tagName } from '@ember-decorators/component';
-import { computed, get } from '@ember/object';
+import { get, action } from '@ember/object';
 import { inject as service } from '@ember/service';
-import { and, readOnly, bool } from '@ember/object/computed';
-import Component from '@ember/component';
-
-@tagName('')
+import Component from '@glimmer/component';
```
locks, line 3:
```suggestion

import Component from '@glimmer/component';
```

locks (dismissed):
Left some minor suggestions inline. Thanks for helping out :)

michaelbdai:
> Left some minor suggestions inline. Thanks for helping out :)

Updated, thanks for the suggestion.

app/components/deprecation-item-source.js
```diff
@@ -1,26 +1,38 @@
...
   }
 
-  @readOnly('port.adapter') adapter;
+  get adapter() {
+    return this.port.adapter;
+  }
 
-  @and('known', 'adapter.canOpenResource') isClickable;
+  get isClickable() {
+    return this.known && this.adapter.canOpenResource;
```
rwwagner90, line 22: Since we are removing computeds, do we need to make `known` etc `@tracked`?

locks: Hm, I don't think so because `known` and `adapter` are getters.

michaelbdai: We don't need to make it tracked. The tracked property should be defined in this component instead of derived from args. And it should be mutable by this component. For example, `@tracked known = true` then we have an action which can mutate this property `this.known = false`. 
However in our use case, known is computed by getter and it is automatically tracked.
```  
  get known() {
    return this.args.itemModel.map.source;
  }
```

app/components/deprecation-item-source.js
```diff
@@ -1,26 +1,38 @@
...
+import Component from '@glimmer/component';
 export default class DeprecationItemSource extends Component {
   @service port;
 
-  @bool('model.map.source') known;
-
-  @computed('model.map.{line,source}', 'known')
   get url() {
-    let source = get(this, 'model.map.source');
+    let source = this.itemModel.map.source;
```
rwwagner90, line 9:
```suggestion
    let source = this.args.itemModel?.map.source;
```

app/components/deprecation-item-source.js
```diff
@@ -1,26 +1,38 @@
...
     if (this.known) {
-      return `${source}:${get(this, 'model.map.line')}`;
+      return `${source}:${this.args.itemModel.map.line}`;
```
rwwagner90, line 11:
```suggestion
      return `${source}:${this.args.itemModel?.map.line}`;
```

app/components/deprecation-item-source.js
```diff
@@ -1,26 +1,38 @@
...
+    return this.port.adapter;
+  }
 
-  @and('known', 'adapter.canOpenResource') isClickable;
+  get isClickable() {
+    return this.known && this.adapter.canOpenResource;
+  }
+
+  get known() {
+    return this.args.itemModel.map.source;
```
rwwagner90, line 26:
```suggestion
    return this.args.itemModel?.map.source;
```

app/components/deprecation-item-source.js
```diff
@@ -1,26 +1,38 @@
...
+  }
+
+  @action
+  handleClick() {
+    this.args.traceSource?.(this.args.modelGroup, this.args.itemModel);
+  }
+
+  @action
+  handleRedirect() {
+    this.args.openResource?.(this.args.itemModel.map);
```
rwwagner90, line 36:
```suggestion
    this.args.openResource?.(this.args.itemModel?.map);
```

rwwagner90 (changes requested):
Noticed tests were failing with `map` being undefined. Left some suggestions. We may not need optional chaining everywhere, but I think we were missing `args` at least.

app/components/deprecation-item-source.js
michaelbdai: updated.

app/components/deprecation-item-source.js
michaelbdai: updated

app/components/deprecation-item-source.js
michaelbdai: updated

app/components/deprecation-item-source.js
michaelbdai: updated

michaelbdai:
> Noticed tests were failing with `map` being undefined. Left some suggestions. We may not need optional chaining everywhere, but I think we were missing `args` at least.

all fixed now.

app/components/deprecation-item-source.hbs
```diff
@@ -1 +1,28 @@
-{{yield this}}
+<div class="deprecation-source pb-2" data-test-deprecation-source>
+  <button
+    class="send-to-console send-to-console--chevron-only mr-3"
+    title="Trace deprecations in console"
+    data-test-trace-deprecations-btn
+    onclick={{this.handleClick}}
```
rwwagner90, line 6:
```suggestion
    {{on "click" this.handleClick}}
```

@michaelbdai what do you think of using more Ember style on click?

PR #1940, opened 2022-03-15
Remove support for the TargetActionSupport mixin

### LINKED ISSUE #1939: Deprecation triggered in 3.28 LTS app
**Describe the bug**
It seems that the inspector triggers the following deprecation warning when running in a new 3.28 application:

> DEPRECATION: Using Ember.TargetActionSupport is deprecated. [deprecation id: ember.built-in-components.import] See https://deprecations.emberjs.com/v3.x#toc_ember-built-in-components-import for more details.

Which I assume comes from here: https://github.com/emberjs/ember-inspector/blob/e6f5ecb0ab4a8cff27bc83912ccd0efe20cb6c4c/ember_debug/object-inspector.js#L38

The warning isn't shown when the inspector is disabled. What's strange is that this doesn't cause issues in a 4.0 app where this mixin shouldn't exist anymore. My best guess would be that the object destructure simply returns undefined instead of the mixin and since nothing is really using it things don't break?

**To Reproduce**
1. Generate a new Ember 3.28 app (`npx ember-cli@3.28 new lts`)
2. start the app `cd lts && npm run start`
3. Open the dev tools and activate the Ember Inspector
4. The warning should appear

**Expected behavior**
The Ember Inspector shouldn't trigger deprecations.

**Screenshots**
image

**Environment**
Ember 3.28, Chrome and FireFox, node 12 / 16, npm 6/8, yarn 1

Windvis:
## Description
This mixin triggers a deprecation warning in Ember 3.28 applications.

Closes #1939

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -35,7 +35,6 @@ const {
   MutableEnumerable,
   NativeArray,
   ObjectProxy,
-  TargetActionSupport,
 } = Ember;
 
 const GlimmerComponent = (() => {
@@ -92,12 +91,15 @@ const emberNames = new Map([
   [NativeArray, 'NativeArray Mixin'],
   [Observable, 'Observable Mixin'],
   [ControllerMixin, 'Controller Mixin'],
-  [TargetActionSupport, 'TargetActionSupport Mixin'],
   [ActionHandler, 'ActionHandler Mixin'],
   [CoreObject, 'CoreObject'],
   [EmberObject, 'EmberObject'],
   [Component, 'Component'],
 ]);
+
+if (compareVersion(VERSION, '3.27.0') === -1) {
+  emberNames.set(Ember.TargetActionSupport, 'TargetActionSupport Mixin');
+}
 
 try {
   const Views = Ember.__loader.require('@ember/-internals/views');


Windvis:
This only fixes the TargetActionSupport deprecation. It is also possible to trigger Ember global deprecations but I haven't figured out what code in the inspector triggers that.

More information: https://github.com/emberjs/ember-inspector/pull/1940

rwwagner90 (dismissed):

rwwagner90:
@Windvis could we wrap this in a version check and only remove it for 3.28+? Something like this https://github.com/emberjs/ember-inspector/blob/09a10b657202fe567d9cb3b72deb9c5d81c2483e/ember_debug/object-inspector.js#L424

ember_debug/object-inspector.js
```diff
@@ -92,13 +91,16 @@ const emberNames = new Map([
...
   [Observable, 'Observable Mixin'],
   [ControllerMixin, 'Controller Mixin'],
-  [TargetActionSupport, 'TargetActionSupport Mixin'],
   [ActionHandler, 'ActionHandler Mixin'],
   [CoreObject, 'CoreObject'],
   [EmberObject, 'EmberObject'],
   [Component, 'Component'],
 ]);
 
+if (compareVersion(VERSION, '3.27.0') === -1) {
```
Windvis, line 100: It seems the deprecation was added in v3.27 according to the deprecation guides: https://deprecations.emberjs.com/v3.x#toc_ember-built-in-components-import

rwwagner90 (approved):

rwwagner90:
Thanks for the PR!

PR #1959, opened 2022-04-08
component-tree: make left/right arrow nav in certain conditions (fixes #1537)

### LINKED ISSUE #1537: Better keyboard navigation & devtools consistency

The recently added keyboard navigation in the components inspector is great but is often hard to work with because it doesn't follow the behavior of devtools that many developers are used to. Not only is it inconsistent but it also lacks functionality that makes it easy to navigate to the parent with one keypress in chunks of trees with many children.

**Describe the solution you'd like**

To mimic devtools behavior: 
- in collapsed view **right arrow** should expand the element (this is implemented)
- in expanded view **left arrow** should collapse the element (this is implemented)
- in expanded view **right arrow** should jump to the first child
- in collapsed view **left arrow** should jump to the parent

Note that both Chrome and Firefox have the same behavior.

geneukum:
## Description

Right now, the left and right arrow keyboard shortcuts will collapse and
expand the currently selected node in the component tree respectively.

It would be nice (and more consistent with devtools) if:
- the left arrow keyboard shortcut would jump to the parent node if the
  currently selected node is already collapsed
- the right arrow keyboard shortcut would jump to the child node if the
  currently selected node is already expanded

### DIFF
--- a/app/controllers/component-tree.js
+++ b/app/controllers/component-tree.js
@@ -200,15 +200,31 @@ export default class ComponentTreeController extends Controller {
       case KEYS.up:
         this.pinned = this.previousItem.id;
         break;
-      case KEYS.right:
-        this.findItem(this.pinned).expand();
+      case KEYS.right: {
+        const pinnedItem = this.findItem(this.pinned);
+
+        if (pinnedItem.isExpanded) {
+          this.pinned = this.nextItem.id;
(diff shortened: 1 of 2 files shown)


rwwagner90 (approved):
LGTM! Thanks for the PR!

PR #1961, opened 2022-04-13
object-inspector: migrate to glimmer component

geneukum:
Let's upgrade the object-inspector component to extend '@glimmer/component'
rather than '@ember/component'. While we're here, let's also upgrade the
component to use native JS syntax.


rwwagner90 (approved):
This looks awesome, thanks @geneukum! We definitely need help updating these things.

geneukum:
@rwwagner90, sure thing! I'd be happy to try to work through some more of these in the coming days.

PR #1973, opened 2022-04-22
deprecations: resolve instances of this-property-fallback

geneukum:
## Description

It is currently possible to reference properties on a component without
a preceding `this`. This style of looking up properties is known as
"property fallback", and has the potential to collide with other types
of values. The alternative way to lookup properties is with a preceding
`this`. This style does not have any ambiguity, since it's clear that
we're looking up the property on this instance of the component, and not
the global helper/component. Property fallback has been deprecated in
favor of this style in general.

As such, let's work through all of our templates in order to eliminate any
instances of this issue. Finally, let's stop silencing violations of this
deprecation in our deprecation-workflow config file. This will ensure that
future violations are not added to the codebase.

## Screenshots
n/a

### DIFF
--- a/.template-lintrc.js
+++ b/.template-lintrc.js
@@ -8,7 +8,6 @@ module.exports = {
 
     // TODO: enable these rules
     'no-action': false,
-    'no-implicit-this': false,
     'no-positive-tabindex': false,
     'table-groups': false,
   },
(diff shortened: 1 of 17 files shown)


geneukum:
> Awesome, thanks so much for the PR!

For sure! I'm trying to do 1-2 of these a day. I'll probably slow down a bit once I get into the ones that require a bit more effort.

PR #1975, opened 2022-04-23
list: fixup dot access for run.bind deprecation

geneukum:
## Description
Since Ember 3.27, using . to access run loop functions has been
deprecated, such as run.bind. Instead, we must import the value directly
from the module like so:

import { bind } from '@ember/runloop';

See the deprecated-run-loop-and-computed-dot-access rule in ember's
deprecation-app for more details.

## Screenshots
n/a

### DIFF
--- a/app/components/list.js
+++ b/app/components/list.js
@@ -1,6 +1,6 @@
 /* eslint-disable ember/require-tagless-components */
 import Component from '@ember/component';
-import { run, scheduleOnce } from '@ember/runloop';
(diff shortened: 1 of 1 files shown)


geneukum:
It looks like some package dependencies will also trigger this, so we can't 'un-silence' this one in the `config/deprecation-workflow.js` quite yet. I figured this small mixup was probably still worth pulling in though?

rwwagner90 (approved):

PR #2008, opened 2022-05-10
Deprecation toolbar clear btn does not clear deprecations

geneukum:
## Description
Let's refactor how we pass deprecations from the route to the controller
to avoid using `this.controller`. `this.controller` is discouraged per
the `no-controller-access-in-routes` eslint rule.

To workaround this, let's pass the deprecations on the model and update
them via this.currentModel instead.

While we're here, let's also add a test case which verifies that the clear 
button on the toolbar will clear all of the deprecations in the deprecation 
pane when clicked. This should help us avoid future regressions to this 
feature.

## Screenshots
n/a


geneukum:
@rwwagner90, also noticed that the pill in the side-nav doesn't appear to update after a "clear." Should it? 

It looks like this is a separate problem in that the side-nav generates its list of items in the constructor, but does not re-render them if the args coming in change. I threw the proposed change in with this stuff, but totally happy to back it out if you don't think that behavior is desirable.

rwwagner90 (approved):
Looks great, thanks!

PR #2009, opened 2022-05-12
route-debug-test: avoid 'view:renderTree' race condition

geneukum:
## Description
Right now, the 'Route tree' test from the 'Ember Debug - Route Tree' has
a race condition between receiving the 'view:renderTree' message and
triggering 'route:getTree'.

The 'view:renderTree' message is sent after a 250ms debounce once a tree
is finished being profiled by the ProfileManager, so settled() isn't
going to be able to cover that operation.

To work around this, let's waitUntil we receive the 'view:renderTree'
message before proceeding with our test. This is fairly similar to how
this message is waited for in the view-debug-test suite.

## Screenshots
n/a


rwwagner90 (approved):
Thanks!

PR #2040, opened 2022-06-08
do not close the object inspector

### LINKED ISSUE #2013: object inspector closes when i click anything on the website
The object inspector closes when i click on anything on the website which i am debugging.
It should stay open so I can inspect the values in the component while interacting with the website.

**Expected behavior**
The Object Inspector should always be open.

**Environment**
Ember version: 4.4.0

the issue is probably here: 
https://github.com/emberjs/ember-inspector/blob/master/ember_debug/libs/view-inspection.js#L323
where hide is called if a component is pinned, and notification to the inspector is sent, which also hides the object inspector.
can we disable the notification?

patricklx:
when selecting a component on the webpage, e.g via context menu. a tooltip will show in the website highlighting the selected component.
after clicking outside the tooltip or clicking escape it will hide the tooltip on the webpage and also the object inspector in the ember inspector.
Hiding tooltip is enough. Keep inspector open so i can debug my app, interact with my app and see changes directly in ember-inspector/ object inspector

fixes #2013

### DIFF
--- a/ember_debug/libs/view-inspection.js
+++ b/ember_debug/libs/view-inspection.js
@@ -311,7 +311,7 @@ export default class ViewInspection {
       } else if (this.isInspecting) {
         event.preventDefault();
         event.stopPropagation();
-        this.stop();
+        this.stop(false);
       }
     }
   }
@@ -345,7 +345,7 @@ export default class ViewInspection {
     if (match) {
       this.show(match.id, pin);
     } else {
-      this.hide();
+      this.hide(false);
     }
 
     if (isInspecting) {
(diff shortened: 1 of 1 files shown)


patricklx:
what about this one @rwwagner90 ? Unless you do not see this issue?

rwwagner90:
@patricklx I have been trying to decide what the correct behavior is here. It could be confusing to have the object inspector open when you have nothing selected.

patricklx:
@rwwagner90 right, but it will still be selected in the ember inspector. Just no tooltip.

patricklx:
Any new thought about this? Maybe also include @chancancode, according to git blame wrote most of this https://github.com/emberjs/ember-inspector/blame/20e8c91051870b881c5741110c731512fc3cb6ba/ember_debug/libs/view-inspection.js#L386

rwwagner90 (approved):

PR #2048, opened 2022-06-21
use vendored requireModule if it becomes available

patricklx:
all requireModule which try to load ember related modules would fail in some cases of some race conditions.
window.requireModule is guaranteed to have the vendored one.
old relevant issue: https://github.com/emberjs/ember-inspector/issues/1595.

this happens when ember inspector was already opened and the app is reloaded. (e.g. during development). in that case it can happen that ember_debug is loaded before ember

### DIFF
--- a/ember_debug/vendor/loader.js
+++ b/ember_debug/vendor/loader.js
@@ -19,6 +19,14 @@ if (typeof define !== 'function' || typeof requireModule !== 'function') {
       }
 
       let mod = registry[name];
+      
+      if (!mod) {
+        mod = window.requireModule(name);
+        if (mod) {
+          return mod;
+        }
+      }
+      
       if (!mod) {
         throw new Error(`Module: '${name}' not found.`);
       }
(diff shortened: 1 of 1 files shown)


rwwagner90:
@patricklx I think we are already using `requireModule` see an example here https://github.com/emberjs/ember-inspector/blob/master/ember_debug/utils/ember/array.js#L6

I believe the correct fix would be to try to get the timing correct, but perhaps I am misunderstanding the issue.

patricklx:
What i mean is that that requireModule can be this one https://github.com/emberjs/ember-inspector/blob/f74a814074ea0234b4eda97462c325f05a09754d/ember_debug/vendor/loader.js#L16 instead of the one provided by the loader.js package. Then loading of inspector will fail

patricklx:
Alright, i looked into it. How does this look like : https://github.com/emberjs/ember-inspector/pull/2050

PR #2050, opened 2022-06-21
load ember debug after ember

patricklx:


app/services/adapters/web-extension.js
```diff
@@ -131,25 +131,35 @@ export default class WebExtension extends BasicAdapter {
...
-            emberDebug = xhr.responseText;
-            resolve(emberDebug);
-          }
-        }
-      };
-
-      xhr.send();
-    } else {
-      resolve(emberDebug);
+  const waitEmber = new Promise((resolve) => {
```
rwwagner90, line 134: Could we call this `waitForEmberLoad` perhaps?

patricklx:
ok, looks good. promise and render speed also show the first render and promise

PR #2051, opened 2022-06-22
fix loading ember-inspector after reload

patricklx:
* sometimes initializers already run or a deferReadiness exists, but still no instance ready. with _bootPromise and didBecomeReady we can ensure that we continue loading
* all requireModule which try to load ember related modules would fail in some cases of some race conditions.
this happens when ember inspector was already opened and the app is reloaded. (e.g. during development). in that case it can happen that ember_debug is loaded before ember. It would use the requireModule provided by ember_debug instead of the ember provided one.
* Without the setTimeout, onEmberReady will do a requireModule('ember') while ember is still loading inside the define. So it's state is not finalized and then ember and onEmberReady will run multiple times. https://github.com/ember-cli/loader.js/blob/master/lib/loader/loader.js#L294
this will also cause e.g requireModule('@ember/array') https://github.com/emberjs/ember-inspector/blob/master/ember_debug/utils/ember/array.js#L6 to return an empty object

### DIFF
--- a/app/services/adapters/web-extension.js
+++ b/app/services/adapters/web-extension.js
@@ -131,6 +131,19 @@ export default class WebExtension extends BasicAdapter {
 function loadEmberDebug() {
   let minimumVersion = config.emberVersionsSupported[0].replace(/\./g, '-');
   let xhr;
+
+  function loadEmberDebugInWebpage() {
+    const waitForEmberLoad = new Promise((resolve) => {
+      if (window.Ember) return resolve();
+      if (window.requireModule && window.requireModule.has('ember')) {
+        return resolve();
+      }
+      window.addEventListener('Ember', resolve, { once: true });
+    });
+    waitForEmberLoad.then(() => {
+      'replace-with-ember-debug';
+    });
+  }
   return new Promise((resolve) => {
     if (!emberDebug) {
       xhr = new XMLHttpRequest();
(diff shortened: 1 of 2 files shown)


rwwagner90:
@patricklx we have 4 PRs now around deferring loading. Which of these do we actually want to go with? Do any of them fix the issue for sure?

rwwagner90:
> so, i found multiple issues with some race conditions during reloading. this just fixes one symptom. So the combination of the prs fixes the problems i have and also work on other sites i tested.

It feels like there should be one thing to wait for and fix it all though, but perhaps not?

patricklx:
that would be the best for sure. but im not able to find something like that. using didBecomeReady and _bootPromise is the best combination i found. We could remove the initalizers, but i'm not sure if didBecomeReady and _bootPromise also work on older ember versions

rwwagner90:
@patricklx gotcha, could we combine the PRs into one and close any that are not needed? I would like to merge the boot fixes all as one, so if anything breaks we can easily revert.

rwwagner90:
@chancancode @teddyzeenny @rwjblue  do any of you have thoughts on this PR? If there are no objections, I am inclined to merge and see how it works. I don't really know how all the boot stuff works here.

rwwagner90 (approved):
In the interest of fixing these things, going to approve. If anyone later has ideas to simplify things, please let us know!

PR #2109, opened 2022-08-19
fix class names in object inspector

patricklx:
also check for Function prototype.toString
its now different from Object.prototype.toString as its overwritten by core-js...
https://github.com/zloirock/core-js/blob/master/deno/corejs/index.js#L1195


rwwagner90 (approved):
Not sure what this is for, but seems okay.

PR #2111, opened 2022-08-19
Convert to manifest v3

RobbieTheWagner:


RobbieTheWagner:
This all seemed to work locally when I tested it, so going to go ahead and merge and if we hit issues we can fix them.

PR #2158, opened 2022-09-20
scroll into view while previewing or selecting from page

patricklx:
## Description
this makes the component tree view scroll the previewing or selected component into view.
This is when previewing components on the page or selecting them via the context menu on the page
I also made sure that it only scrolls one time to the current item

### DIFF
--- a/app/components/scroll-container.js
+++ b/app/components/scroll-container.js
@@ -78,7 +78,7 @@ export default class ScrollContainerComponent extends Component {
     if (this.lastCurrentItem?.id === this.currentItem?.id && !this.previewing) {
       return;
     }
-    
+
     if (!this.previewing) {
       this.lastCurrentItem = this.currentItem;
     }
--- a/app/components/scroll-container.js
+++ b/app/components/scroll-container.js
@@ -66,17 +66,18 @@ export default class ScrollContainerComponent extends Component {
 
   @action
   scrollPreviewIntoViewIfNeeded() {
-    let { previewing } = this;
-    if (!previewing) return;
-    this.scrollIntoViewIfNeeded();
+    if (this.previewing) {
+      this.scrollIntoViewIfNeeded();
+    }
   }
 
   @action
   scrollIntoViewIfNeeded() {
     let { element, scrollTarget } = this;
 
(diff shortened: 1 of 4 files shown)


app/components/scroll-container.js
```diff
@@ -10,7 +10,9 @@ export default class ScrollContainerComponent extends Component {
 
   @tracked collection;
   @tracked currentItem;
+  @tracked previewing;
   @tracked itemHeight;
+  lastCurrentItem;
```
rwwagner90, line 15: Shouldn't this be tracked?

app/components/scroll-container.js
```diff
@@ -62,10 +64,23 @@ export default class ScrollContainerComponent extends Component {
...
+    if (!previewing) return;
+    this.scrollIntoViewIfNeeded();
+  }
+
   @action
   scrollIntoViewIfNeeded() {
     let { element, scrollTarget } = this;
 
+    if (this.lastCurrentItem?.id === this.currentItem?.id && !this.previewing)
+      return;
```
rwwagner90, line 79:
```suggestion
    if (this.lastCurrentItem?.id === this.currentItem?.id && !this.previewing) {
      return;
    }  
```

app/components/scroll-container.js
```diff
@@ -62,10 +64,23 @@ export default class ScrollContainerComponent extends Component {
     }
   }
 
+  @action
+  scrollPreviewIntoViewIfNeeded() {
+    let { previewing } = this;
+    if (!previewing) return;
+    this.scrollIntoViewIfNeeded();
```
rwwagner90, line 71:
```suggestion
    if (this.previewing) {
      this.scrollIntoViewIfNeeded();
    }
```

rwwagner90 (changes requested):

app/components/scroll-container.js
patricklx: It did work correctly without it.... I try it out tomorrow with tracked.

patricklx: After reviewing, its not needed.

rwwagner90 (approved):

PR #843, opened 2018-07-27
Context menu item for inspecting components

Bestra:
context_menu
Features:
- Once the inspector is open and on screen, users can right click on an element and open it in the component tree via the context menu
- The component tree will scroll the relevant component into view.
This PR supersedes #788.  Closes #689

PR #2159, opened 2022-09-21
rename inspect to select

patricklx:
I suggest to rename this so to prevent confusion. the native inspect opens the inspect tools itself, but thats not possible to do with an extension.

### DIFF
--- a/skeletons/web-extension/background-script.js
+++ b/skeletons/web-extension/background-script.js
@@ -92,7 +92,7 @@
     if (isEmberApp && !contextMenuAdded) {
       chrome.contextMenus.create({
         id: 'inspect-ember-component',
-        title: 'Inspect Ember Component',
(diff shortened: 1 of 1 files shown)


patricklx:
i think it worked for firefox: https://github.com/emberjs/ember-inspector/blob/main/skeletons/firefox/lib/devtools-utils.js#L43.

patricklx:
i tried to find something in the github history, but no success. 
for chrome i found this
https://bugs.chromium.org/p/chromium/issues/detail?id=112277

the initial implementation also did not have it.
https://github.com/emberjs/ember-inspector/pull/843

the [implementation ](https://github.com/emberjs/ember-inspector/blob/main/skeletons/firefox/lib/devtools-utils.js#L43) for firefox is also very old and not supported anymore

PR #2162, opened 2022-09-23
fix args of template only components

patricklx:
just noticed that the objects in the args had only an id as it was serialized. no need to serialize args at this point


rwwagner90:
> just noticed that the objects in the args had only an id as it was serialized. no need to serialize args at this point

@patricklx would it be possible to add a test to prevent this regression in the future?

rwwagner90:
@patricklx thanks for adding a test! Looks like some prettier errors still remain.

patricklx:
now the test works correctly

rwwagner90 (approved):

PR #2168, opened 2022-09-29
cleanup iframes & firefox skeleton

patricklx:
iframe is now done via windows messages
firefox skeleton is not used


rwwagner90:
@patricklx are you sure firefox skeleton is not used? I know we still build for firefox.

patricklx:
Yeah, but the build output for Firefox shows its just using the web extension skeleton. Nothing from Firefox inside

patricklx:
@rwwagner90 anything special i need to do to load the extension in Firefox? I was unable to do it. With or without this change

patricklx:
confirmed that this works in firefox

rwwagner90:
@patricklx so Firefox just uses the Chrome extension build? I'm confused why we don't need any configs for Firefox.

patricklx:
yeah, it uses the same as chrome. those are the WebExtension APIs and there are only some incompatibilities https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Chrome_incompatibilities

rwwagner90:
@patricklx so if I approved and merge this PR, our builds will still create a firefox.zip and publish it to their store? Just making 100% sure here.

patricklx:
yeah, i tried it locally

rwwagner90 (approved):

PR #2189, opened 2022-10-13
fix post message for testem

### LINKED ISSUE #2188: v4.8.0 on Chrome raises errors from Testem during Integration Tests
**Describe the bug**
During an integration test run, the `content-script.js` injection for the v4.8.0 Chrome Extension causes Testem to raise `[Object object] is not valid JSON.`. Appears to log to console at the very start of `hooks.beforeEach()` for integration tests. The text attempting to be parsed is `ember-inspector-iframe-ready`.

**To Reproduce**
1. In most recent Chrome version, with most recent Ember Inspector extension installed (v4.8.0):
2. Start the ember test server
3. Run any test module containing a `hooks.beforeEach` call

**Expected behavior**
No errors are logged to console due to Ember Inspector's content script injection.

**Screenshots**
image

**Environment**
Testem v3.9.0 with latest Chrome and extension on MacOS.

**Additional context**
Unlike a separate Ember-Inspector-borne error (one in which having the extension's dev tools tab instantiated at test teardown will cause test failure due to attempts to "remove a function listener"), the dev tool tab does not need to be instantiated for this to throw.

Additionally, despite the nature of the beforeEach hook, the error will log to console only once when running a test module that contains multiple test cases. However, it will log to console regardless of if you run all tests in the module or any single one of them by itself.

patricklx:
do same as `debugger-client` post message

testem does a JSON.parse on all messages they receive without verification...

JSON.parse works with strings

fixes #2188


rwwagner90 (approved):

rwwagner90:
@patricklx seems like the tests are hanging?

patricklx:
It didn't even start the tests, i closed and reopen now it's running

rwwagner90 (approved):

PR #2201, opened 2022-10-19
bring back tracked dependencies

patricklx:
## Description
in earlier versions of ember, tags had an _propertyKey which was used to show some tracked dependencies of getters. this brings it back for current versions.
in addition I tried to improve the way its displayed 

## Screenshots
image


rwwagner90:
@patricklx Thanks for the PR! Could we add some tests around this? It seems rather complex and I want to make sure we don't have regressions.

rwwagner90 (approved):

PR #2204, opened 2022-10-24
remove terser

patricklx:
because firefox needs non-minified code

minifyJS is already set to false, but its not applied for terser
https://github.com/emberjs/ember-inspector/blob/main/ember-cli-build.js#L33


rwwagner90:
@patricklx We've been fine on Firefox for quite awhile. I believe we can keep terser, since this is open source and we point them to the source.

PR #2219, opened 2022-10-31
update vertical-collection & ember-table

### LINKED ISSUE #2200: ember inspector stops responding, error: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
**Describe the bug**
inspector stops responding to any interaction, error in inspector is 
Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
- While rendering:
  -top-level
    application
      app-detected
        component-tree
          scroll-container
            vertical-collection

**To Reproduce**
inconsistent, hard to reproduce consistently 

happens when children are being removed from the component tree.
This seems to happen when there are many renders.
I have a clock which updates a few times per second. When addition interaction is done it increases the likelihood

patricklx:
fixes #2200

### DIFF
--- a/app/templates/component-tree.hbs
+++ b/app/templates/component-tree.hbs
@@ -20,11 +20,10 @@
   {{did-insert this.arrowKeysSetup}}
   {{will-destroy this.arrowKeysTeardown}}
 >
-  {{#vertical-collection
-    this.visibleItems
-    estimateHeight=this.itemHeight
-    key="id" as |item|
-  }}
+  <VerticalCollection
+    @estimateHeight={{this.itemHeight}}
+    @items={{this.visibleItems}}
+    @key="id" as |item|>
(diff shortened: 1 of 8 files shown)


patricklx:
this always fails locally, but looks like its working in CI...

patricklx:
i think its really just a problem on vertical collection. vertical collection is used directly for the component tree. other routes are indeed using ember-table and I saw that it has the same problem

patricklx:
@rwwagner90 you might be right about updating ember-table. its still using the older version of vertical-collection and it looks like ember-inspector is picking that one...

patricklx:
right, the test for the issue passed, but a bunch of other issues. ill look into them to update ember-table

rwwagner90 (approved):

rwwagner90:
Thanks @patricklx!

PR #2244, opened 2022-11-18
add script tag to the head instead

### LINKED ISSUE #2243: in-page-script.js inserted into Wordpress editor
**Describe the bug**
`<script src="chrome-extension://bmdblncegkenkacieihfhpjfppoconhi/scripts/in-page-script.js" type="text/javascript"></script>` is inserted into the Wordpress editor and saved to the database. 

**To Reproduce**
1. Edit a page/post in Wordpress with the old/classic editor enabled and the 'Visual' mode as default. (classic-editor plugin is enabled on the sites I tested)
2. Click on the 'text' mode for the editor and see that the script is inserted

I don't know if the new gutenberg editor also has this problem, but I guess that every editor that uses iframes is effected.

**Expected behavior**
ember-inspector should not insert the in-page-script in the editor

patricklx:
fixes #2243 
with this it wont be intercepted by the wordpress editor

### DIFF
--- a/skeletons/web-extension/content-script.js
+++ b/skeletons/web-extension/content-script.js
@@ -76,10 +76,10 @@
   var script = document.createElement('script');
   script.type = "text/javascript";
   script.src = chrome.runtime.getURL("scripts/in-page-script.js");
-  if (document.body && document.contentType !== "application/pdf") {
-    document.body.appendChild(script);
+  if (document.head && document.contentType !== "application/pdf") {
+    document.head.appendChild(script);
     script.onload = function() {
-      document.body.removeChild(script);
+      document.head.removeChild(script);
     };
   }
 


patricklx:
its already beeing removed

rwwagner90:
@patricklx could this cause potential timing issues?

patricklx:
I don't think there should be any timing issue. Or what do you mean. At least the loaded script waits for the page load to complete and then executes the code

rwwagner90:
> I don't think there should be any timing issue. Or what do you mean. At least the loaded script waits for the page load to complete and then executes the code

I mean moving to executing in the head will make it run earlier. Could there be things that are not ready at that time? I would assume we were putting it in the body for a reason before?

patricklx:
I just checked the history of the script tag, and indeed in older versions it was depending on jQuery to subscribe to page load. Now its not using any library. So it can be safely loaded in the head tag

rwwagner90 (approved):

Issue #2237, opened 2022-11-10
Property Inspector attempts nested access of a property whose name includes a period


tehhowch:
**Describe the bug**
If an inspected object contains a property whose name contains period, the Property Inspector pane will report either `undefined` or, if it happens to be a valid path, the value when following the path.

**To Reproduce**
1. Define some route / controller / component property to be an object with at least one key that has period in it, e.g. `oopsie = { ok: 'shown', 'not.ok': 'please show me', not: { ok: "don't show this!" } }`
2. Open the application and load the route/controller/component
3. Open the extension and inspect the relevant entity
4. click on the object property

**Expected behavior**
The inspected object should have 3 properties with the following displayed content under "Own Properties"
```
not: { ok: don't show this! }
not.ok: "please show me"
ok: "shown"
```

**Screenshots**
![image](https://user-images.githubusercontent.com/20871346/200967720-d2c2a406-c4b5-4619-9209-67b8ef8618c1.png)

**Environment**
Ember 3.28, firefox, extension v4.9.1

**Additional context**
If one clicks the `>$E`, the object in the console tray has the expected layout. Thus, this is just a display issue due to the use of some lookup method that treats a period as a directive to follow a nested property path. Probably this one:
https://github.com/emberjs/ember-inspector/blob/6853fac6dcd63e75fc5a436ddd9ffdce8b6511fc/ember_debug/object-inspector.js#L1407

PR #2270, opened 2022-12-18
fix: Use correct value path in object inspector (#2237)

rossketron:
## Description
Fixes known issue #2237.

If a property name includes a `.` such as in `{ 'property.name': 123 }`, the value is now accessed using bracket notation instead of `get()` in order to treat the name as a string literal and avoid attempting to access a nested value.

If a property name does not include a `.`, there is no change in functionality.

A test was added to ensure the correct value is accessed given this scenario.

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -1404,7 +1404,9 @@ function calculateCP(object, item, errorsForObject) {
     if (object instanceof Ember.ArrayProxy && property == parseInt(property)) {
       return object.objectAt(property);
     }
-    return item.isGetter ? object[property] : get(object, property);
+    return item.isGetter || property.includes('.')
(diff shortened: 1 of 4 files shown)


rwwagner90:
@rossketron thanks for the PR! Forgive my ignorance, but will this still work correctly for nested properties as well?

patricklx:
I think it should be fine as it only iterates over the object keys. So there will be no nested access.

rossketron:
@rwwagner90 @patricklx - Thanks for the feedback! 

I'm happy to add more robust testing for this as well if it would be helpful.

patricklx:
What about digging into the property with the dot? Does it work? Look at function digIntoObject

rwwagner90 (approved):

PR #2402, opened 2023-06-02
support in-element & wormhole

### LINKED ISSUE #1255: Cannot select any element with an in-element block.
**Describe the bug**

When clicking the black box with yellow text, 
the component that contains that element is not selected in the component tree, `RenderInTarget`
instead, the template containing the in-element target is highlighted -- I don't think this is the desired behavior?

image

**To Reproduce**
1. Clone https://github.com/NullVoxPopuli/repro-repo-ember-inspector-1255
2. start app
3. Click the component picker button:
    image

4. see that the application route is highlighted and selected instead of `RenderInTarget`

**Expected behavior**
RenderInTarget should be highlighted and selected

**Environment**
Firefox
Ember Inspector 4.1.0
Ember 3.18.1

patricklx:
this injects a InElement component into the component tree, which has its bounds set to the target. it also adds it as a root tree element, so the click & hover work correctly.
This is also done for EmberWormhole, the code checks for the EmberWormhole instance and changes the bounds to its target element

resolves #1255 
closes #1375

### DIFF
--- a/ember_debug/libs/render-tree.js
+++ b/ember_debug/libs/render-tree.js
@@ -1,5 +1,225 @@
 import captureRenderTree from './capture-render-tree';
 import { guidFor } from 'ember-debug/utils/ember/object/internals';
+import { A } from 'ember-debug/utils/ember/array';
+import Ember from 'ember-debug/utils/ember';
+
+class InElementSupportProvider {
+  constructor(owner) {
+    this.nodeMap = new Map();
+    this.remoteRoots = A([]);
+    this.currentNode = null;
+    this.nodeStack = A([]);
+    this.remoteNodeStack = A([]);
+    this.runtime = this.require('@glimmer/runtime');
+    try {
+      this.Wormhole = requireModule('ember-wormhole/components/ember-wormhole');
+    } catch (e) {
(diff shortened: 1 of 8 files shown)


patricklx:
@RobbieTheWagner this is ready for review

RobbieTheWagner (approved):
This seems okay to me, I just wonder if there is anything built in to the internals of glimmer or modifiers that could make this easier? @wycats @chancancode @MelSumner @kategengler who is working on things related to this? Would love to get a thumbs up from them that this is the best way to do things.

NullVoxPopuli (approved):
lgtm, though, I'd like to see it tested out first -- got a demo / gif or something of before/after?

lifeart (approved):

RobbieTheWagner:
@NullVoxPopuli I am sure this is functionally fine, I just wanted to make sure there wasn't an easier glimmer API we could hook into or something because this all seems very manual.

RobbieTheWagner:
Since there has been no response from stakeholders, I am going to merge this. If there is a better way, we can always update things later. Thanks for the PR @patricklx!

PR #2410, opened 2023-06-14
use yarn 3

patricklx:


RobbieTheWagner:
@patricklx any advantage to using yarn 3 over pnpm?

patricklx:
Not sure, pnpm is probably still faster than yarn 3

PR #2411, opened 2023-06-15
use pnpm

patricklx:
## Description
now you can choose :)


.github/workflows/build.yml
```diff
@@ -193,14 +199,14 @@ jobs:
         env:
           PR: ${{ github.event.pull_request.number }}
       - name: Build
-        run: yarn ember build --environment production
+        run: pnpm ember build --environment production
       - name: Pack
         run: |
-          VERSION="$(jq -r '.version | split("+") | .[0]' package.json)"
```
RobbieTheWagner, line 199: Why was this changed?

.github/workflows/build.yml
```diff
@@ -193,14 +199,14 @@ jobs:
...
+          VERSION="$(jq -r '.version' package.json)"
 
-          yarn pack
+          pnpm pack
           mkdir -p dist/npm
-          tar xvzf "ember-inspector-v$VERSION.tgz" -C dist/npm --strip-components 1
```
RobbieTheWagner, line 203: We probably want to keep this name the same, right?

RobbieTheWagner (changes requested):

.github/workflows/build.yml
patricklx: it doesnt matter as its extracted anyway into dist/npm

patricklx: I misunderstood.
Pnpm pack generates a different filename

.github/workflows/build.yml
RobbieTheWagner: So the new one will work with all possible version types? Not sure if this is also needed for nightly versions etc

.github/workflows/build.yml
RobbieTheWagner: Okay, just makes me worried changing names

.github/workflows/build.yml
patricklx: Well, the version string during the t

patricklx:
I squashed all the changes into one commit

RobbieTheWagner (approved):

PR #2421, opened 2023-06-26
fetch latest version for ember try

patricklx:
with this we can drop the resolution-highest in the npmrc file and have even faster installs & tests (well, 1-2sec)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
 public-hoist-pattern[]=@babel/*
-resolution-mode=highest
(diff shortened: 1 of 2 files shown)


RobbieTheWagner:
How does this help us? Is this something new ember-try is going to be doing going forward?

NullVoxPopuli:
Ember try will now be using --no-lockfile, introduced in:

patricklx:
Lets leave it as it is then. As this is also not a library.

PR #2426, opened 2023-07-07
limit max width of tooltips

patricklx:
## Description
limit width of tooltips and clips too long content.
On hover it will show the full string as tooltip, via title attribute

### DIFF
--- a/ember_debug/libs/view-inspection.js
+++ b/ember_debug/libs/view-inspection.js
@@ -160,6 +160,8 @@ function makeStylesheet(id) {
       text-align: right;
       color: #000;
       background: transparent;
+      max-width: 400px;
+      word-break: break-word;
     }
 
     #${prefix}-tooltip-${id} .${prefix}-tooltip-arrow {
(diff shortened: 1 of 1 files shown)


RobbieTheWagner:
@patricklx can you please give me an example of when this is needed and possibly some before/after screenshots?

patricklx:
Screenshot 2023-07-10 141141
Screenshot 2023-07-10 141207

RobbieTheWagner:
I'm still not sure on whether we want to limit this or not.

patricklx:
That can also be done. Is that preferred @RobbieTheWagner ?

patricklx:
would like like this:

image

NullVoxPopuli:
oooo I like that

NullVoxPopuli (approved):

RobbieTheWagner:
Wrapping is definitely preferred here. Are the wrapping changes in?

RobbieTheWagner (approved):

PR #2443, opened 2023-08-02
Add ability to jump to source for functions and classes

patricklx:
## Description
this adds the action "goto source".
the behaviour is that it will call inspect on functions and classes which will show the source code of it.

It makes use of the existing function for dom node inspection. I only changed to name to be more generic (inspectJsValue)

this is ideal for going to a source of function or classes and set breakpoints for debugging

## Screenshots
image


patricklx:
@RobbieTheWagner can you have a look?

RobbieTheWagner:
@patricklx sorry for the delay! This looks great, but could we please add some test coverage? I want to make sure everything merged in has robust tests, so we do not have regressions.

RobbieTheWagner (approved):

RobbieTheWagner:
Thank you for the PR @patricklx! This is a great feature.

PR #2444, opened 2023-08-03
show icons in whats-new page

patricklx:
replace icon placeholders from lerna-changelog with actual icons


RobbieTheWagner (approved):
Great idea!

PR #2458, opened 2023-08-29
build ember_debug with vite

### LINKED ISSUE #2457: Breaks on ember 5.2 w/ staticEmberSource: true
**Describe the bug**

Visit: 
- https://limber.glimdown.com
- or locally at http://localhost:4201/
- open the inspector
- observe: `Uncaught TypeError: _computed.readOnly is not a function`

**Screenshots**

image

Refreshing with the inspector open and "pause on exceptions" checked gives the actual problem code:

image

image

**Environment**
ember-source 5.2, all browsers

patricklx:
and selective choose what we actually need from the clients ember
ember_debug.js  382.13 kB │ gzip: 118.16 kB when minimized

fixes #2457 

This is actually not required. But it could reduce the things we need to import from the clients ember


patricklx:
this will not work after all...
missing @glimmer/runtime in the client app and other required ones...

RobbieTheWagner:
@patricklx this seems to break Ember < 3.28. If this is something we are going to need, we will have to cut a new legacy branch and make the main branch only support Ember 3.28+

patricklx:
I could probably support older versions as well. But we need a compat adapter anyway. This just reduces it

patricklx:
🎉
it works, but I don't think we gain much.  
So, just a nice proof of concept. :)

It reduces the deps on ember to
* @ember/runloop
* @ember/instrumentation
* @ember/debug
* @ember/object/internals
* ember

patricklx:
@NullVoxPopuli @RobbieTheWagner 
Do you think it makes sense to continue work on this?
I can try to remove the require('ember') call as well and use direct imports of the needed deps

NullVoxPopuli:
maybe -- I'm not entirely sure what the purpose is -- in my 5.2 app with `staticEmberSource` enabled, I still get an error:
image



does this work make progress towards fixing that? I presume by somehow adding in modules that an app may not have?

patricklx:
Well, this will at least reduce those issues.
Whats missing there for example is the runloop, specifically, the backburner export.
But that is required to have from the client app. And having that export will not increase your apps size.
The biggest issue i think is the dependency on require ember, but ember_debug does not actually need all.

patricklx:
Actually, we could also fallback to own runloop, but then we need to have a setup a interval, so that backburner will be invoked. 
Will then be poll based instead.

patricklx:
this now works on [limber](https://limber.glimdown.com/)
but
* no tracked detection
*  polling based
* current inspect object is also iterated over all props without checking for changes (need @glimmer/validator)
* no promise inspection (need rsvp)
* no data tab (maybe you do not have?)

NullVoxPopuli:
Why close?

patricklx:
This fixes only half of the problem. 
It makes ember_debug use it's own ember.  But it still needs access to multiple parts of the clients ember to do the debugging.
And all the parts ember_debug depends on, are already in the clients ember, also with static ember. Just not exposed via the amd loader.
I think ember_debug could also be refactored to reduce more native things likr classes aglnd getters. Which  would also reduce the dependency on ember object, computed etc

NullVoxPopuli:
One idea someone had yesterday was to:
- make ember inspector wholly use its own ember-source
- have ember-source provide an "adapter" to  hook in to the inspector, the same way ember-data does -- that way ember-source is responsible for compatibility, rather than the inspector

thoughts?

patricklx:
Sounds good to me. could be backported to earlier ember versions? 
There is actually already ember/debug for that.
The problem is in earlier versions it did not provide enough debug tooling for specific things like tracked detection and tracked dependency detection.
Which is done now in a hacky way.

NullVoxPopuli:
> could be backported to earlier ember versions?

at the very least LTSes -- but it's not unheard of for folks on older embers to use older ember-inspectors

PR #2459, opened 2023-08-29
remove title

patricklx:
follow up of #2426 
Forgot to revert this...

### DIFF
--- a/ember_debug/libs/view-inspection.js
+++ b/ember_debug/libs/view-inspection.js
@@ -522,7 +522,6 @@ export default class ViewInspection {
       this._renderTokens(td, value);
     } else {
       td.innerText = value;
-      td.title = value;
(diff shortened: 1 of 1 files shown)


RobbieTheWagner (approved):
Changes look fine, but we have some test failures somehow.

patricklx:
strange... it's not the tooltip issue, but the highlighter offset is somehow changed now... 
works okay in browser test, but not in headless...

patricklx:
i guess showing the tooltip causes the scrollbars to appear, which then offsets the component

RobbieTheWagner (approved):

patricklx:
@RobbieTheWagner can you remove the label? Or change it? It's not necessary for this to appear in the changelog :)

PR #2460, opened 2023-08-31
use word-wrap and add line break controls

patricklx:
sorry for pr again...
found this issue when a small component is at the right side
image

### DIFF
--- a/ember_debug/libs/view-inspection.js
+++ b/ember_debug/libs/view-inspection.js
@@ -161,7 +161,7 @@ function makeStylesheet(id) {
       color: #000;
       background: transparent;
       max-width: 400px;
-      word-break: break-word;
+      word-wrap: break-word;
     }
 
     #${prefix}-tooltip-${id} .${prefix}-tooltip-arrow {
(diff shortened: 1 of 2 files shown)


patricklx:
found this https://www.fileformat.info/info/unicode/char/200b/index.htm
with this we can control where line breaks are in addition allowed. I added them around the `/` character

RobbieTheWagner:
@patricklx it feels like we're playing whack-a-mole with this issue. I don't think we should be specifying which characters we wrap on. We need to set a max-width on the container for this text and just have it auto wrap.

patricklx:
with word-wrap: break-word; it only breaks max-width is reached. before it was breaking even sooner.
specifying the character is just a nice plus point. otherwise it will break anywhere. But if it finds a better spot to break on it uses that instead. Since we always have some `/` chars in the path, we can use that. (by default it will not break on them, it only breaks on spaces and `-`, maybe others as well)

patricklx:
it works well with `word-wrap: break-word` without specifying min-width.
default behaviour just looks ugly, when it breaks the word anywhere, because it cannot find a good point to break the word.

patricklx:
with specifying breaking points:
image


and without (I did remove the '-' to demonstrate):

image

NullVoxPopuli (approved):

RobbieTheWagner (approved):

PR #2475, opened 2023-10-09
modernize ember debug

### LINKED ISSUE #2457: Breaks on ember 5.2 w/ staticEmberSource: true

patricklx:
## Description
this modernizes ember debug to use modern js patterns and reduce usage of emberjs.
the only parts accessing emberjs is to do the actual debugging

fixes #2457, although there will not be tracked detection for static sources


RobbieTheWagner:
Wow, this is a big one, but something we probably should have tackled a long time ago. Thank you for taking this on!

patricklx:
this might also fix #2457

RobbieTheWagner:
@patricklx mind fixing the conflicts? Hopefully I can review this soon.

pnpm-lock.yaml
```diff
@@ -7692,8 +7702,8 @@ packages:
       - webpack
     dev: true
 
-  /ember-svg-jar@2.4.6:
-    resolution: {integrity: sha512-W+t6sRAiPhRl4imGKEMWsQDiCV8Nl2GRwzUpqZ1a+CZoW+2aCkF+kcngeDlHn5XzT16bQXg/QYXUvwtKJlSLfg==}
+  /ember-svg-jar@2.4.4:
```
RobbieTheWagner, line 7705: This seems to have been downgraded?

patricklx: strange... great catch
i fix it

ember_debug/view-debug.js
```diff
@@ -4,54 +4,56 @@ import RenderTree from 'ember-debug/libs/render-tree';
...
+
+      contextMenu() {
+        let { lastRightClicked } = this;
+        this.lastRightClicked = null;
+        this.inspectNearest(lastRightClicked);
+      },
+    };
+  }
 
   init() {
```
RobbieTheWagner, line 55: Should this be `init` still or should we update it to `constructor`?

ember_debug/view-debug.js
```diff
@@ -4,54 +4,56 @@ import RenderTree from 'ember-debug/libs/render-tree';
...
-  },
+export default class extends DebugPort {
+  get adapter() {
+    return this.namespace?.adapter;
+  }
+  get objectInspector() {
+    return this.namespace?.objectInspector;
+  }
+
+  static {
```
RobbieTheWagner, line 15: Do we need static and prototype stuff here or could this just go on the class instance?

patricklx: this is needed on prototype.
as `setupOrRemovePortListeners` accesses the `.messages` and `.portNamespace` in the constructor.
Which will happen before the sub class can set them up in their own constructor.
We could change this by either passing the messages and portNamespace as constrcutor param.
or by calling `setupOrRemovePortListeners` in the sub classes instead

patricklx: actually, I can make it work the same way as Ember Core Object.
```
const obj = new this();
obj.init();
return obj;
````

ember_debug/view-debug.js
patricklx: i left init, because it gets overwritten in some parts in the tests. I didn't check which ones exactly

PR #2478, opened 2023-10-12
catch errors while getting args in capture tree

patricklx:
workaround for
https://github.com/glimmerjs/glimmer-vm/pull/1447

### DIFF
--- a/ember_debug/libs/render-tree.js
+++ b/ember_debug/libs/render-tree.js
@@ -67,10 +67,30 @@ class InElementSupportProvider {
 
     const captureNode = this.debugRenderTree.captureNode;
     this.debugRenderTree.captureNode = function (...args) {
-      const capture = captureNode.call(this, ...args);
+      const nodeFor = this.nodeFor;
       const [id, state] = args;
       const node = this.nodeFor(state);
-      self.setupNodeRemotes(node, id, capture);
+      let capture;
+      try {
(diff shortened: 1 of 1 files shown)


ember_debug/libs/render-tree.js
```diff
@@ -67,10 +67,30 @@ class InElementSupportProvider {
...
       const [id, state] = args;
       const node = this.nodeFor(state);
-      self.setupNodeRemotes(node, id, capture);
+      let capture;
+      try {
+        capture = captureNode.call(this, ...args);
+        self.setupNodeRemotes(node, id, capture);
+      } catch (e) {
+        this.nodeFor = () => {
+          this.nodeFor = nodeFor;
```
RobbieTheWagner, line 79: Can you please explain this? I'm confused why we would set this to a function that then sets itself to something else immediately.

patricklx: CaptureNode will call this function also for children nodes, so need to reset to original immediately

RobbieTheWagner:
@patricklx should this be fixed in glimmer, instead of here?

patricklx:
Maybe, if it can be backported. 
@NullVoxPopuli ? What do you think.

lifeart (approved):

patricklx:
I think this is rare enough that we can just wait until we have the fix from glimmer vm

PR #2484, opened 2023-10-25
fix .get is not defined

patricklx:
use index lookup

### DIFF
--- a/ember_debug/libs/promise-assembler.js
+++ b/ember_debug/libs/promise-assembler.js
@@ -93,7 +93,7 @@ class PromiseAssembler extends BaseObject {
     if (guid) {
       const index = this.promiseIndex[guid];
       if (index !== undefined) {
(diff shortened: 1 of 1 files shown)


ember_debug/libs/promise-assembler.js
```diff
@@ -93,7 +93,7 @@ class PromiseAssembler extends BaseObject {
     if (guid) {
       const index = this.promiseIndex[guid];
       if (index !== undefined) {
-        return this.all.get(index);
+        return this.all[index];
```
RobbieTheWagner, line 96: Are we sure this is never a proxy and won't need `get` ever? Could we add a small test to prevent regressions here?

patricklx: Yes. I missed this during the migration to native array wl

RobbieTheWagner (approved):

PR #2489, opened 2023-11-03
Add id for manifest v3 support in Firefox

RobbieTheWagner:
From what I can tell, we need to add a unique id for the extension when using manifest v3 with Firefox. This should accomplish that goal.


RobbieTheWagner:
@patricklx I know you did some work on supporting manifest v3. Does this seem correct to you?

patricklx:
Sound okay. There is also an error link in the build log

patricklx (approved):

PR #2495, opened 2023-11-06
switch to source-map-js

### LINKED ISSUE #2491: Ember inspector 4.10 not working with 3.28
**Describe the bug**

since the last version ember inspector no longer detects the project

**To Reproduce**

1. Go to localhost
2. open inspector

 "ember-cli": "~3.28.6",
    "ember-data": "~3.28.6",
    "ember-source": "^3.28.6",

**Screenshots**
image

patricklx:
I wanted to load the source map from node_modules. but that wasnt working properly. 
Currently its not possible to trace in console for deprecations.

Also fixes #2491
Probably, it's the only place that actually could call `require` in certain conditions.
All others use `requireModule`

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -104,9 +104,13 @@ module.exports = function (defaults) {
 
   let emberDebug = 'ember_debug';
 
-  let sourceMap = new Funnel('node_modules/source-map/dist', {
-    files: ['source-map.js'],
+  let sourceMap = new Funnel('node_modules/source-map-js', {
+    files: ['**/*.js'],
     destDir: 'ember-debug/deps',
+  });
+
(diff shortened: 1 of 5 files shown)


patricklx:
@RobbieTheWagner I think we need this. But I wasn't able to test if this fixes https://github.com/emberjs/ember-inspector/issues/2491

robertpbaxter:
@patricklx I pulled your branch and tested in the hopes that it might get past the service worker issue in #2491, but no changes on my end.

patricklx:
Ah, right. This only fixes the first half half

RobbieTheWagner (approved):
This seems fine to me 👍

PR #2497, opened 2023-11-06
make amd safe

patricklx:
Some could overwrite these with their own implementation.
we can always use requireModule only

### DIFF
--- a/ember_debug/vendor/loader.js
+++ b/ember_debug/vendor/loader.js
@@ -1,4 +1,7 @@
-let define = window.define,
+let define = window.define && (...args) => window.define(...args),
+  exports = undefined,
+  module = undefined,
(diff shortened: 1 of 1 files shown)


ember_debug/vendor/loader.js
```diff
@@ -1,5 +1,10 @@
-let define = window.define,
+/* eslint-disable */
+let define = window.define && ((...args) => window.define(...args)),
+  exports = undefined,
+  module = undefined,
```
RobbieTheWagner, line 4: Why set these to undefined?

patricklx: so that global exports, module etc will not make it into ember_debug

RobbieTheWagner (approved):

PR #2498, opened 2023-11-07
Remove outdated code and unnecessary captureRenderTree patching

patricklx:
* remove outdated code, patched capture render tree is not needed anymore, its already in the locked ember version
* move Ember.x usages to requires (only left is Ember.Libraries, there is no import for it...)


ember_debug/utils/type-check.js
```diff
@@ -13,17 +12,6 @@ export function isComputed(object, key) {
   if (Debug.isComputed && Debug.isComputed(object, key)) {
```
patricklx, line 12: Ember > 3.10

RobbieTheWagner (approved):

PR #2499, opened 2023-11-07
fix inspector not showing up in certain cases

### LINKED ISSUE #2496: Inspector not loading: issues with chrome api
there are issues with chrome api. I'm not sure why it errors out for some, and for others is just hidden
Separate issue out of #2491 
**extensions using event pages or service workers...**
onclick event is the issue 

Use generic on click event?
https://github.com/GoogleChrome/chrome-extensions-samples/blob/9aaa5ca35817e6dfa151566f4441a2fe3bf057ff/api-samples/contextMenus/basic/sample.js#L16

**ember tombster not hide/shown:'**
chrome.action.hide is not a function. need to use enable/disable

patricklx:
fix issues with chrome api.
for some the inspector does not show up, for others it does, not sure why.

fix onclick event & hide/show

action icon is always shown
there is no option to hide/show anymore
instead i switch between grey/orange icon to indicate detection.

enable/disable actually only disables click actions

fixes #2496


skeletons/web-extension/manifest.json
```diff
@@ -39,8 +43,8 @@
 
   "action": {
     "default_icon": {
-      "19": "{{PANE_ROOT}}/assets/images/icon19.png",
-      "38": "{{PANE_ROOT}}/assets/images/icon38.png"
+      "19": "{{PANE_ROOT}}/assets/images/icon19_grey.png",
```
RobbieTheWagner, line 46: Why are the default icons the grey ones?

tests/integration/injection-test.js
```diff
@@ -66,13 +66,23 @@ class ChromeApi {
...
         delete self.registeredContextMenus[name];
       },
 
       create(menu) {
         self.registeredContextMenus[menu.id] = menu;
+        self.registeredContextMenus[menu.id].onclick = () =>
+          contextMenuListeners.forEach((fn) => fn({ menuItemId: menu.id }));
+      },
+
+      onClicked: {
```
RobbieTheWagner, line 82: I assume you tested this to ensure it still found the component in the tree and selected it?

skeletons/web-extension/manifest.json
patricklx: by default it should show `not detected` icon. as the detection state only shows when an ember app is detected

tests/integration/injection-test.js
patricklx: i tested it now :)
works

RobbieTheWagner (approved):

PR #2501, opened 2023-11-08
correctly load scripts in chrome & firefox

### LINKED ISSUE #2500: Inspector is stomping app's EmberENV
This is a non-deterministic failure that depends on how fast the inspector and the app load.

But if the app loads fast and the inspector loads slow, you can suddenly get bizarre results because the inspector completely replaces your `window.EmberENV` with its own [here](https://github.com/emberjs/ember-inspector/blob/ba75238bc4f1e8cdea83d41ebc4761f2e189f66f/skeletons/web-extension/scripts/boot.js#L1). This throws away your own settings.

I think it usually works because the inspector sets first, but that is apparently not guaranteed.

Probably it would be sufficient to make the inspector add properties to an existing `EmberENV` if it finds that one already exists.

patricklx:
this does not work for firefox
```
Reading manifest: Warning processing content_scripts.0.world: An unexpected property was found in the WebExtension manifest.
```
Also has warnings about permissions

but firefox keeps the order right for script loading

it needs to run before everything else so that ember gets the debug config. If the property is set later, component tree will not work.
this issue only happens on page refresh, then browser uses cached data and immediately runs the scripts, without waiting for the new script that have been added by extensions. this does not happen in firefox, only chrome.
https://bugs.chromium.org/p/chromium/issues/detail?id=634381

fixes #2500

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -273,9 +273,21 @@ module.exports = function (defaults) {
 
   replacementPattern = replacementPattern.concat(emberInspectorVersionPattern);
 
+  const firefoxBackgroundServiceReplacement = [
+    {
+      match: /"service_worker": "background.js"/g,
+      replacement: '"scripts": ["background.js"]',
+    },
+  ];
+
   const skeletonWebExtension = replace('skeletons/web-extension', {
     files: ['*'],
     patterns: replacementPattern,
+  });
+
+  const skeletonFirefoxWebExtension = replace('skeletons/web-extension', {
+    files: ['*'],
+    patterns: replacementPattern.concat(firefoxBackgroundServiceReplacement),
   });
 
   const skeletonBookmarklet = replace('skeletons/bookmarklet', {
@@ -285,7 +297,7 @@ module.exports = function (defaults) {
 
   let firefox = mergeTrees([
     mv(mergeTrees([tree, emberDebugs.firefox]), webExtensionRoot),
-    skeletonWebExtension,
+    skeletonFirefoxWebExtension,
   ]);
 
   let chrome = mergeTrees([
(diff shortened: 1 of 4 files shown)


skeletons/web-extension/boot.js
```diff
@@ -3,6 +3,7 @@
 // TODO: make this conditional, only do it when requested.
 // only inject boot script when on an HTML page
 if (document.contentType === 'text/html') {
+  window.EmberENV = { _DEBUG_RENDER_TREE: true };
```
RobbieTheWagner, line 6: Wouldn't we want to check for existence of `window.EmberENV` here too instead of setting it to our own object?

ember-cli-build.js
```diff
@@ -273,19 +273,31 @@ module.exports = function (defaults) {
 
   replacementPattern = replacementPattern.concat(emberInspectorVersionPattern);
 
+  const firefoxBackgroundServiceReplacement = [
```
RobbieTheWagner, line 276: What is this for? Firefox _should_ support manifest v3, right?

patricklx: It doesn't... Not all of it

skeletons/web-extension/boot.js
patricklx: Not needed here. This really absolutely runs before anything else

ember-cli-build.js
patricklx: https://github.com/mozilla/web-ext/issues/2532

patricklx:
@RobbieTheWagner i found a better way for firefox
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts

RobbieTheWagner (approved):

PR #2505, opened 2023-11-10
fix lazy routes debug

patricklx:


patricklx:
This was also needed. I think this happens for engines and embroider split routes

RobbieTheWagner (approved):

PR #2512, opened 2023-11-16
Proposal: serialize remote nodes on in-element nodes to avoid dom exceptions 

mattmarcum:
## Description
At my work we have a complex component that does direct dom manipluation and uses an in-element helper to boot.  This caused the ember-inspector to crash with a dom exception error when trying to deepClone the render tree (dom nodes can not be deep cloned).  Some debugging showed that the dom nodes were showing up in the `remotes` array of the in-element node, specifically in the `bounds` of the remote node items.  Running those nodes through the same `_serializeRenderNodes` method kept ember-inspector from crashing.

### DIFF
--- a/ember_debug/libs/render-tree.js
+++ b/ember_debug/libs/render-tree.js
@@ -165,7 +165,9 @@ class InElementSupportProvider {
   registerRemote(block, node) {
     const obj = this.buildInElementNode(node);
     if (this.currentNode) {
-      this.currentNode.remotes = this.currentNode.remotes || [];
+      if (!this.currentNode.remotes) Object.defineProperty(this.currentNode, 'remotes', {
+        value: []
+      });
       this.currentNode.remotes.push(obj);
     }
     this.remoteRoots.push(obj);


mattmarcum:
I'm sure someone else has more knowledge of how to fix this, but I wanted to put up the pr just to get the conversation started.  All tests pass locally for me.

patricklx:
I wonder if we actually need the remotes in the serialized tree?
It's only used for finding in-element nodes. I don't think it should be there

mattmarcum:
> I wonder if we actually need the remotes in the serialized tree? I do not remember anymore... It's only used for finding in-element nodes. I don't think it should be there

I think I tried initially just deleting the `remotes` key but that caused ember-inspector to crash.  Its possible I just deleted it at the wrong place in the process flow.

ember_debug/libs/render-tree.js
```diff
@@ -508,6 +508,7 @@ export default class RenderTree {
         ),
         bounds: this._serializeBounds(node.bounds),
         children: this._serializeRenderNodes(node.children, node),
+        remotes: node.remotes ? this._serializeRenderNodes(node.remotes, node) : undefined,
```
patricklx, line 511: Probably just setting this to undefined is enough

patricklx: Or use Object.defineProperty to hide it with enumerable: false. Then in should not be included here

patricklx: https://github.com/emberjs/ember-inspector/blob/5f14672530940fa5b48682783116530f67d58bc7/ember_debug/libs/render-tree.js#L168

patricklx: at line 168
replace
```js
this.currentNode.remotes = this.currentNode.remotes || []; 
```
with
```js
if (!this.currentNode.remotes) Object.defineProperty(this.currentNode, 'remotes', {
  value: []
});
```
Object.defineProperty will make in non-enumerable by default, that way it will not be copied when doing `{...node}`

patricklx (approved):

PR #2531, opened 2023-12-19
fix cannot access array values

### LINKED ISSUE #2530: CalculateCPError on Nested Arrays with Objects & Unresponsive After Less than 3 Mintues
**Describe the bug**
My team (4 of us) have all been experiencing 2 separate and fairly large bugs since 4.10.X series released about 3-4 weeks ago. We expected these were known issues that would be resolved, but after a few weeks I'm amazed that there's no issues here, or other reports in Discord, so I am reporting them myself. 

**TWO ISSUES:**
1. Objects within arrays are un-inspectable (from a service or component .js file). Clicking the array shows the objects in it, but clicking on one of those objects just results in "CalculateCPError" instead of showing you all the properties like previously:
Screenshot 2023-12-19 at 9 05 59 AM
2. The inspector goes entirely unresponsive after less than 3 minutes of opening it. We are constantly having to close DevTools and re-open.

**To Reproduce Issue 1**
1. Run `ember init`
2. Create a .js file (component or service)
3. Add an array of objects (as a property, native getter, or computed() ) to .js file:
```
  people = [
    {
      id: '1234',
      name: 'justin',
    },
    {
      id: '4321',
      name: 'jim',
    },
  ];
```
5. Save file, inject service or add component to Application template, `ember serve`, and open app in browser
6. Open Ember Inspector & Navigate to Inspect the class you just created (either from Components tree or container -> service).
7. Click on the array you added to class (works as expected - shows object 0, 1, etc).
8. Click to drill into one of the objects in this array, and **notice the CalculateCPError** with no property details

**To Reproduce Issue 2**
Simply let the inspector sit open for a few minutes on an app. Return to notice that none of the UI elements do anything when clicked. You must close DevTools & re-open to re-activate.

**Expected behavior**
On 4.9.X, we were able to drill down into objects within arrays and see all properties on it in the inspector. We also were able to keep the inspector open indefinitely without it becoming unresponsive.

**Environment**
- MacOS 14.0 with Apple Silicon
- Tested on Chrome 120.0.6099.109 & Firefox 120.0.1 
- Originally Experienced on our Ember 3.28.X applications, but confirmed on a fresh `ember init` app running 5.4.X.

_Worth noting that this is my team's setup, but we also asked someone with a Windows machine to test and they're experiencing the same things._

patricklx:
property can be a number

fixes #2530

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -1317,7 +1317,7 @@ function calculateCP(object, item, errorsForObject) {
     if (object instanceof Ember.ArrayProxy && property == parseInt(property)) {
       return object.objectAt(property);
     }
-    return item.isGetter || property.includes('.')
+    return item.isGetter || property.includes?.('.')
       ? object[property]
       : object.get?.(property) || object[property]; // need to use `get` to be able to detect tracked props
   } catch (error) {


kdagnan:
Running locally, can confirm this fixes inspecting arrays. Thank you for such a quick fix.

Does not fix the unresponsiveness issue. Can open as separate issue if preferred.

RobbieTheWagner (approved):

PR #2533, opened 2023-12-20
ensure port is connected

### LINKED ISSUE #2532: Inspector Becomes Unresponsive When Left Open for a Few Minutes
**Describe the bug**
This has started occurring with the 4.10.X series, it was fine a few weeks ago on 4.9.X.

After opening the inspector, it works as expected for a few minutes. Within a few minutes of opening and leaving unattended, it becomes unresponsive. Clicking anywhere does nothing most of the time, occasionally it will switch from your previous page to all grey:
Screenshot 2023-12-19 at 3 36 35 PM

**To Reproduce**
1. Open the Inspector
2. Let it sit for 3-5 minutes
3. Return to find that none of the functionality works
4. Close DevTools & re-open to get it working again

**Environment**
Tested on:
- MacOS 14.0 (Apple Silicon) Chrome 120.0.6099.109 & Firefox 120.0.1
- Windows 11 (Intel) Chrome 120
- MacOS 14.0 (Apple Silicon) Chrome 120

Tested with:
- All 4 of our company's applications, running Ember-source 3.28
- A fresh `ember init` application running Ember-source 5.X

patricklx:
fixes #2532 

looks like service workers can be shutown after idle time... with manifest V3.
but it can be stopped any time, as it is not guaranteed.
also we reconnect immediately so that we can receive updates  when some start coming in from ember_debug site
https://developer.chrome.com/docs/extensions/develop/concepts/service-workers/lifecycle#idle-shutdown

tested manually by stopping the service worker in chrome://serviceworker-internals/?devtools

### DIFF
--- a/app/services/adapters/web-extension.js
+++ b/app/services/adapters/web-extension.js
@@ -60,12 +60,23 @@ export default class WebExtension extends BasicAdapter {
   }
 
   _connect() {
+    if (this._isConnected) {
+      return;
+    }
     let chromePort = this._chromePort;
     chromePort.postMessage({ appId: chrome.devtools.inspectedWindow.tabId });
 
     chromePort.onMessage.addListener((...args) => {
       this._messageReceived(...args);
     });
+
+    chromePort.onDisconnect.addListener(() => {
+      this._isConnected = false;
+      this.notifyPropertyChange('_chromePort');
+      this._connect();
+    });
+
+    this._isConnected = true;
   }
 
   _handleReload() {
--- a/app/services/adapters/web-extension.js
+++ b/app/services/adapters/web-extension.js
@@ -60,9 +60,6 @@ export default class WebExtension extends BasicAdapter {
   }
 
   _connect() {
-    if (this._isConnected) {
-      return;
-    }
     let chromePort = this._chromePort;
     chromePort.postMessage({ appId: chrome.devtools.inspectedWindow.tabId });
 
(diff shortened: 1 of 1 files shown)


kdagnan:
Thank you! We will run this version today and report back

kdagnan:
One freeze so far today. These error messages were all there yesterday but the disconnectedPort one is gone now: 
image

patricklx:
@kdagnan I added some logging, can you test with this?

patricklx:
i added the logging to https://github.com/patricklx/ember-inspector/tree/port-connect-logging

i think this already makes sense to merge, a similar issue was reported on discord

patricklx:
@RobbieTheWagner

kdagnan:
I think that the _ensureConnected change actually fixed the original issue, I just realized the only time I am having issues now are when stopped at a breakpoint in the browser. Inspector freezes up, but when breakpoints are all stepped through, it works again. 

Installed 4.9.1 and it had the same behavior although I could've sworn it worked previously.

Thank you again for your help & quick fixes.

app/services/adapters/web-extension.js
```diff
@@ -60,12 +60,23 @@ export default class WebExtension extends BasicAdapter {
...
+    }
     let chromePort = this._chromePort;
     chromePort.postMessage({ appId: chrome.devtools.inspectedWindow.tabId });
 
     chromePort.onMessage.addListener((...args) => {
       this._messageReceived(...args);
     });
+
+    chromePort.onDisconnect.addListener(() => {
+      this._isConnected = false;
```
RobbieTheWagner, line 74: Do we need this `this._isConnected` boolean? If we're going to call `this._connect` `onDisconnect` wouldn't that be sufficient?

patricklx: We don't. fixed

RobbieTheWagner (approved):

PR #2536, opened 2023-12-27
improve route debug

patricklx:
* use backburner instead of setInterval
* use routeHandler.get
* show error if getting routes failed


RobbieTheWagner (approved):
This looks okay to me 👍

PR #2546, opened 2024-01-26
add 🔸 to tracked dependencies that caused the last invalidation

patricklx:
we also do our own revision/value tracking so we can show the changed deps between the updates


RobbieTheWagner:
@patricklx can you please explain what this is doing and add some tests for this? We seem to be making some complex changes here.

patricklx:
@RobbieTheWagner when a tracked property is changed, it will be marked in the display of tracked deps

patricklx:
e.g. in power-select after search:
image

RobbieTheWagner:
> @RobbieTheWagner when a tracked property is changed, it will be marked in the display of tracked deps

@patricklx gotcha, what is the use case for needing to know that? Also, should we add some kind of tooltip explaining what it is?

patricklx:
If there is some unexpected rendering or recomputation of getters, this will be of help and this came up sometimes in the community

patricklx:
A tooltip might be useful. But will need more redesign work of the ux. I would prefer to do that in a separate pr

RobbieTheWagner (approved):

PR #2548, opened 2024-02-02
add modifiers to render tree

patricklx:
## Description
this adds modifiers and their html element to the render tree.

this also improves the UX for positional args, especially functions

this builds on #2549

### DIFF
--- a/app/components/component-tree-arg.js
+++ b/app/components/component-tree-arg.js
@@ -12,6 +12,15 @@ export default class ComponentTreeArg extends Component {
 
   get displayValue() {
     if (this.isObject) {
+      if (this.args.value.inspect) {
+        if (this.args.value.type === 'function') {
+          return this.args.value.inspect
+            .replace(/"/g, '\\"')
+            .replace('bound ', '')
+            .replace('{ ... }', '');
+        }
+        return truncate(this.args.value.inspect.replace(/"/g, '\\"'));
+      }
       return '...';
     } else if (typeof this.args.value === 'string') {
       // Escape any interior quotes – we will add the surrounding quotes in the template
--- a/app/components/component-tree-item.hbs
+++ b/app/components/component-tree-item.hbs
@@ -23,8 +23,10 @@
 
     <code
       class="component-tree-item__tag flex whitespace-no-wrap
+        {{if @item.isClosingTag 'component-tree-item__closing'}}
+        {{if @item.hasModifiers 'component-tree-item__has_modifier'}}
         {{if
-          @item.isComponent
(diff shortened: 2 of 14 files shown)


patricklx:
image

patricklx:
this is ready

RobbieTheWagner (dismissed):
LGTM!

RobbieTheWagner:
@patricklx the code and tests look great here, thank you for taking this on! I was looking at the screenshot though, and shouldn't the modifier be inside the element? They seem to be outside of the closing `>`.

patricklx:
It's by design, i didn't try to put it inside. Maybe it can be done,  i can have a look tomorrow

patricklx:
I think it makes more sense to have it under the element for ux. Otherwise it can be too much on one line.

patricklx:
@RobbieTheWagner i tried a few things, but was not able to do it in a good way:
maybe add some indicator that its inside an element? 

Screenshot 2024-02-09 at 14 22 33

patricklx:
Wrapping is not good in this case, since we are using vertical-collection which needs predictable row heights. 
So everything would need to be in one line...


Also
i often prefer to have it like this:
```
<li
    class="foo"
   {{did-insert this.myCoolFunction}}
>
```
:)

patricklx:
how about this:
Screenshot 2024-02-09 at 17 19 23

RobbieTheWagner:
> how about this: Screenshot 2024-02-09 at 17 19 23

This looks great! 👍

patricklx:
pushed the changes

patricklx:
@RobbieTheWagner what about this?

skeletons/web-extension/content-script.js
```diff
@@ -24,6 +24,7 @@
    *     https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API
    */
   window.addEventListener('message', function(event) {
+    console.log('message', event);
```
RobbieTheWagner, line 27: Do we want to keep all of these console logs or were they for testing?

patricklx: ready

patricklx:
I found that this might need more work. The modifiers need to be removed from debug render tree when they get destroyed

patricklx:
@RobbieTheWagner ready

RobbieTheWagner:
@patricklx looks like there are some conflicts. Would you mind resolving those please?

patricklx:
@RobbieTheWagner rebased

RobbieTheWagner (approved):

PR #2549, opened 2024-02-02
improve in element support

patricklx:


RobbieTheWagner:
Thanks for the PR! @patricklx could we please write some tests to cover some of this? It all feels very brute force and I want to make sure we're covering our bases.

patricklx:
@RobbieTheWagner we already have tests for it:
https://github.com/emberjs/ember-inspector/blob/6e5ee34fc1abfaf56993422b9f72a023ab2e14f1/tests/ember_debug/view-debug-test.js#L654

I'm just improving it. I also implementing in in glimmer-vm. So maybe we can remove this code in the feature. I try to also get it into the older ember lts versions...

RobbieTheWagner (approved):

PR #2553, opened 2024-02-07
improve dependent keys ux

patricklx:
## Description
this now also marks the changed dependent keys with a red text colour and adds a tooltip

## Screenshots
Screenshot 2024-02-10 at 00 22 30

### DIFF
--- a/app/components/object-inspector/dependent-keys.hbs
+++ b/app/components/object-inspector/dependent-keys.hbs
@@ -1,14 +1,17 @@
 <li class="mixin__property-dependency-list relative">
   {{svg-jar "dependent-key-connection" width="20px" height="10px"}}
   <ul class="m-0 p-0 list-none">
-    {{#each @keys as |depKey|}}
-      <li class="mixin__property-dependency-item relative text-base12 text-sm">
-      {{#if (match depKey "•")}}
+    {{#each @keys as |dep|}}
+      <li
+        class="mixin__property-dependency-item relative text-base12 text-sm {{if dep.changed 'text-spec01'}}"
+        title={{if dep.changed 'this property changed'}}
+      >
+      {{#if dep.child}}
         <span
           class="mixin__property-dependency-name subkey"
           data-label="object-property-name"
         >
-          {{depKey}}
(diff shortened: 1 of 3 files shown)


app/components/object-inspector/dependent-keys.hbs
```diff
@@ -1,14 +1,17 @@
 <li class="mixin__property-dependency-list relative">
   {{svg-jar "dependent-key-connection" width="20px" height="10px"}}
   <ul class="m-0 p-0 list-none">
-    {{#each @keys as |depKey|}}
-      <li class="mixin__property-dependency-item relative text-base12 text-sm">
-      {{#if (match depKey "•")}}
+    {{#each @keys as |dep|}}
+      <li
+              class="mixin__property-dependency-item relative text-base12 text-sm {{if dep.changed 'text-spec01'}}"
```
RobbieTheWagner, line 6: It looks like there is extra spacing here, can we remove that?

RobbieTheWagner:
Would it be possible to get a screenshot please?

app/components/object-inspector/dependent-keys.hbs
patricklx: fixed

patricklx:
added screemshot

RobbieTheWagner:
Can you please explain what the idea is here? I don't find it very intuitive from the screenshot what it is that we are gaining.

patricklx:
this is a follow up https://github.com/emberjs/ember-inspector/pull/2546

RobbieTheWagner:
Are we able to make the title show up closer to the property that changed? It seems unclear what it belongs to in the screenshot.

patricklx:
That's the native tooltip. I can try to make another screen shot with the cursor at the right position. Need to find the right keycombo...

RobbieTheWagner:
> That's the native tooltip. I can try to make another screen shot with the cursor at the right position. Need to find the right keycombo...

If it matches up to the correct item in reality, then that is fine with me.

RobbieTheWagner (approved):

PR #2563, opened 2024-03-12
add release plan

patricklx:
i think it would be nice to have release-plan here as well

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -252,7 +252,7 @@ jobs:
     needs:
       - test
       - ember-try
-      - build    
+      - build
     runs-on: ubuntu-latest
     steps:
       - name: Set up node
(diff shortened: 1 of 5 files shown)


patricklx:
@NullVoxPopuli can you have a look if this setup is correct?
especiially because we have scheduled release as well (nightly) to npm as well. Does `pnpm release-plan` also create a github tag & release. and is it disabled of tag=alpha ?

RobbieTheWagner:
We cannot do automated releases because releasing to the Chrome store doesn't work. We need to find a way to get a new token or whatever for that.

patricklx:
But we are already doing it. This just gives a nice preview PR that can be merged to trigger the publishing.
You probably need to get a new refresh token?

patricklx:
Apparently it should be possible to use the same refresh token and client ID for multiple extensions.
We can try using the refresh token for the nightly extension. That one is working

patricklx:
Another thing i found:
Looks like refresh token expire after 60days of no usage. One more reason to use the same as nightly.

Or just do a custom token fetch :
https://dev.to/cardinalby/dont-let-google-refresh-token-expire-pie

patricklx:
@RobbieTheWagner i updated the publishing to use the same token as nightly

patricklx:
> > @RobbieTheWagner i updated the publishing to use the same token as nightly
> 
> You are sure that is okay? I am not sure why we had separate ones before.

Not really:)
Worst case is that publishing will fail and you have to do it manually

NullVoxPopuli:
> can you have a look if this setup is correct?

I'm nervous about non-boilerplate versions of the workflows :sweat_smile: 
One of the things I like about release-plan is that I (for the most part) don't need to think about what's going in to the workflows -- there's been one exception so far, but the change was small.

Personally, I think we should keep extension publishing separate, but have it run on the condition that the publish.yml (to npm) has completed successfully -- that way we can still manage our custom stuff separately from the boilerplate (which should be more easily replaceable via `npx create-release-plan-setup@latest --update`)

patricklx:
I changed the flow to let release plan make the publishing & tag & release. With minimal changes 
In tag, build.yml will run.

patricklx:
this is ready now

RobbieTheWagner:
@NullVoxPopuli does this seem correct to you? I am not super familiar with release-plan, but would love to introduce it here.

RobbieTheWagner (approved):
If NVP thinks it looks good, good enough for me. Thanks for the PR!

PR #2567, opened 2024-04-09
add internal to dependabot prs

patricklx:


RobbieTheWagner:
@mansona @NullVoxPopuli where are we on release-plan supporting `dependencies` labels? I would prefer not to add this.

NullVoxPopuli:
I haven't added support yet been busy -- but the place to do it would probably be here: 
https://github.com/embroider-build/github-changelog/blob/a5474628ddea80c2c528ef578a25a0d3f5b0f5d8/src/configuration.ts#L104

RobbieTheWagner:
I guess we can just do this for now

RobbieTheWagner (approved):

PR #2569, opened 2024-04-29
add request permissions page

patricklx:
## Description
this adds a link to a request permission page for ember inspector.
The page will open in a separate window, this is one of the possibilities to request permissions.
It  cannot be done directly from the extension page in devtools.

## Screenshots
image


RobbieTheWagner:
@patricklx can you please explain in more detail what this PR is for? Is it just a shortcut to request permissions? Can we not add the request permissions stuff to the icon in the toolbar of Chrome?

patricklx:
@RobbieTheWagner in the toolbar you can already click on the ember icon and select which permissions to give to local website. So this is just a shortcut and a link to the page will be presented on the ember inspector error page.

PR #2571, opened 2024-04-29
fix beta & canary tests

patricklx:
## Description
ember beta & canary now have support for in-element & modifiers in debug render tree


RobbieTheWagner (approved):
LGTM!

PR #2580, opened 2024-06-27
allow continue when beta/canary fails 

patricklx:


RobbieTheWagner:
@patricklx Wouldn't we want it to fail if any failures happen?

patricklx:
> @patricklx Wouldn't we want it to fail if any failures happen?

It would still show up as error. Only for the release process it would not block the publishing if beta/Canary fails

PR #2581, opened 2024-06-28
provide name for hbs template only components

patricklx:
for template only components coming from .hbs files we can know the name by using the moduleName.

this is when gjs/gts files import hbs template only components. Or also when references via js. e.g `<this.MyTemplateOnlyComponent />`

### DIFF
--- a/ember_debug/libs/render-tree.js
+++ b/ember_debug/libs/render-tree.js
@@ -663,7 +663,7 @@ export default class RenderTree {
       if (node.type === 'component' && !node.instance) {
         if (
           node.name === '(unknown template-only component)' &&
-          node.template.endsWith('.hbs')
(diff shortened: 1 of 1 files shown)


RobbieTheWagner (approved):

RobbieTheWagner:
Seems logical. We need to get beta and canary passing though.

patricklx:
canary & beta are failing because we cannot patch glimmer vm anymore to add the `tracked` inspection... 
I suggest we just ignore those for now until we have https://github.com/glimmerjs/glimmer-vm/pull/1489

RobbieTheWagner:
@patricklx can you please rebase this so we can skip that test?

RobbieTheWagner (approved):

PR #2585, opened 2024-07-10
Use GH_PAT token to hopefully fix releases

patricklx:
this is needed so that the tag push triggers a workflow run, which will not happen if using gihub bot token

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,6 @@
 name: Build and Publish
 on:
+  workflow_dispatch:
   pull_request: {}
   push:
     branches:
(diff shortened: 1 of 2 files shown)


RobbieTheWagner:
@patricklx can you please explain how this fixes the issue? Does GITHUB_DEPLOY_TOKEN exist already?

patricklx:
It doesn't exist. Its an example of what needs to change

RobbieTheWagner:
@patricklx can you explain? I don't understand how changing a token would change how things work.

patricklx:
The default GitHub token here would be used to create a git tag and push it.
We have a workflow that would then run on git tag push. The problem is that with the default GitHub token no workflows will be triggered.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

RobbieTheWagner (approved):

PR #2592, opened 2024-07-25
Just in case

kategengler:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
       - ember-2.7.0-3.4.0
       - ember-0.0.0-2.7.0
     tags:
-      - v*
+      - '*'
   schedule:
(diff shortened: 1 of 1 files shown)


RobbieTheWagner (approved):

RobbieTheWagner:
All the tags seem to start with `v` so I am not sure if we need this. Probably wouldn't hurt, but I just wonder what would happen if for some reason we pushed an unrelated tag?

kategengler:
It would try to deploy on that tag if you had another. The absence of the `v` isn't the part I really want to try, it is the quoting of the thing. Per docs, it shouldn't affect it but I've seen it sometimes make a difference.

RobbieTheWagner:
@kategengler since we had a successful publish run, I would conclude this is unnecessary. Should we close it?

PR #2595, opened 2024-07-27
fix boolean edit

### LINKED ISSUE #2594: Unable to update boolean config bit
**Describe the bug**
Ember inspector updated to version 4.13. Since the update, unable to update any boolean value. 

**To Reproduce**
1. Go to 'Components'
2. Click on 'application route'
3. Select 'config'
4. Search for property that's a bool that you want to change
5. Click into value to edit 

**Expected behavior**
Clicking the bool for the config you want to change opens the field. You can switch the value to be true or false. 

**Actual behavior**
Clicking the bool for the config you want to change only highlights the bool value. User can go to Console and change the config value there using `Ember.set(config, 'valueOne', false)`

**Environment**
Ember version 4.13 
Chrome Version 127.0.6533.73

patricklx:
## Description
fixes #2594

### DIFF
--- a/app/components/object-inspector/property.ts
+++ b/app/components/object-inspector/property.ts
@@ -122,10 +122,7 @@ export default class ObjectInspectorProperty extends Component<ObjectInspectorPr
 
   get cannotEdit() {
     if (this.args.model.name === '...' || !this.isCalculated || this.readOnly) return true;
-    return
-      this.args.model?.value?.type !== 'type-string' &&
-        this.args.model?.value?.type !== 'type-number' &&
-        this.args.model?.value.type !== 'type-boolean';
+    return !['type-string', 'type-number', 'type-boolean'].includes(this.args.model?.value?.type);
   }
 
   @action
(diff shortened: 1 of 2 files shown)


app/components/object-inspector/property.ts
```diff
@@ -122,7 +122,10 @@ export default class ObjectInspectorProperty extends Component<ObjectInspectorPr
 
   get cannotEdit() {
     if (this.args.model.name === '...' || !this.isCalculated || this.readOnly) return true;
-    return this.args.model?.value?.type !== 'type-string' && this.args.model?.value?.type !== 'type-number';
+    return
+      this.args.model?.value?.type !== 'type-string' &&
+        this.args.model?.value?.type !== 'type-number' &&
+        this.args.model?.value.type !== 'type-boolean';
```
RobbieTheWagner, line 128:
```suggestion
    return !['type-string', 'type-number', 'type-boolean'].includes(this.args.model?.value?.type);
```

Perhaps we could simplify this a bit?

PR #2598, opened 2024-10-10
Update to Ember 5.12

RobbieTheWagner:


RobbieTheWagner:
@patricklx could you please let me know if you notice anything wrong with this PR? I know it's huge and I don't expect you to review it line by line, but if you could please let me know what you think, that would be great, thank you!

ember_debug/object-inspector.js
```diff
@@ -993,7 +993,7 @@ function isInternalProperty(property) {
     '__LEGACY_OWNER',
     '__ARGS__',
     '__HAS_BLOCK__',
-    '__PROPERTY_DID_CHANGE__',
+    '-property_DID_CHANGE__',
```
patricklx, line 996: this doesn't look correct?

tests/acceptance/object-inspector-test.js
```diff
@@ -924,8 +902,6 @@ module('Object Inspector', function (hooks) {
   });
 
   test('Date fields are editable', async function (assert) {
```
patricklx, line 904: not sure why this test is failing now.
I see that we have 2 `respondWith('objectInspector:saveProperty')`, but we are only doing 1 save.
So I think you can just remove the first `respondWith('objectInspector:saveProperty')`

ember_debug/object-inspector.js
RobbieTheWagner: Ah, yes good catch! I did a lot of find/replace and I guess I messed up some of that.

tests/acceptance/object-inspector-test.js
RobbieTheWagner: @patricklx hmm, yes that does seem odd, however, the boolean tests are failing too, and that one just has one `respondWith`. I will try removing what you said though and see if anything changes.

config/ember-try.js
```diff
@@ -1,95 +1,108 @@
...
-            },
+module.exports = async function () {
+  return {
+    usePnpm: true,
+    scenarios: [
+      {
+        name: 'ember-lts-3.16',
+        npm: {
+          devDependencies: {
+            '@ember/test-helpers': '^2.4.0',
```
patricklx, line 14: also need to add ember-resolver v11.0.1 until ember-source < 4.12

RobbieTheWagner: @patricklx Yeah, is that safe to do? I honestly don't know if we will use consuming app's resolvers when they boot inspector? If so, that sounds like a good fix.

patricklx: @RobbieTheWagner it's safe. It's only used for the UI. We do not consume it from app code

RobbieTheWagner: Thanks @patricklx, that fixed some of the tests, but looks like we still have some issues.

patricklx: some are failing because of ember-cli-app-version, I do not think we need it. can just be removed from deps

patricklx: Probably need more cleanup. Ember 6 beta removes support for extend prototypes.

RobbieTheWagner:
Since the beta and canary failures are not related to the changes here, I am going to go ahead and merge and we should fix deprecations in a separate PR.

PR #2601, opened 2024-10-24
More progress removing prototype extensions

RobbieTheWagner:

### DIFF
--- a/app/controllers/promise-tree.js
+++ b/app/controllers/promise-tree.js
@@ -1,31 +1,44 @@
-// eslint-disable-next-line ember/no-observers
-import { action, observer } from '@ember/object';
+import { action } from '@ember/object';
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import { isEmpty } from '@ember/utils';
-import { equal, bool, and, not, filter } from '@ember/object/computed';
+import { filter } from '@ember/object/computed';
 import { debounce, next, once } from '@ember/runloop';
+import { tracked } from '@glimmer/tracking';
 
-export default Controller.extend({
-  queryParams: ['filter'],
+// eslint-disable-next-line ember/no-observers
+import { observes } from '@ember-decorators/object';
 
-  adapter: service(),
-  port: service(),
+export default class PromiseTreeController extends Controller {
+  queryParams = ['filter'];
 
-  createdAfter: null,
+  @service adapter;
+  @service port;
+
+  @tracked createdAfter = null;
+  @tracked filter = 'all';
+  @tracked searchValue = null;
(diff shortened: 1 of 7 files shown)


patricklx:
it looks like computed.x are incompatible with tracked array. e.g. computed.and in `controllers/promise-tree.js`

RobbieTheWagner:
> it looks like computed.x are incompatible with tracked array. e.g. computed.and in `controllers/promise-tree.js`.

@patricklx I am not sure I understand. What is the fix here? I do not see tracked arrays in that controller.

patricklx:
> > it looks like computed.x are incompatible with tracked array. e.g. computed.and in `controllers/promise-tree.js`.
> 
> @patricklx I am not sure I understand. What is the fix here? I do not see tracked arrays in that controller.

The model is a tracked array. The fix is to use getters instead of computed

RobbieTheWagner:
@patricklx I updated, but it's got a new error `Global error: Uncaught Error: Assertion Failed: EmberObject.create no longer supports defining computed properties. Define computed properties using extend() or reopen() before calling create()`

patricklx:
computed.filter should also be replaced with getter.
I think the issue is in models/promise . There is a custom decorator which returns a computed

RobbieTheWagner:
@patricklx do you mean `dateComputed`? Do you know how we would fix it?

app/controllers/promise-tree.js
```diff
@@ -1,33 +1,44 @@
...
   // below used to show the "refresh" message
-  isEmpty: equal('model.length', 0),
-  wasCleared: bool('createdAfter'),
-  neverCleared: not('wasCleared'),
-  shouldRefresh: and('isEmpty', 'neverCleared'),
+  get isEmpty() {
+    return this.model.length === 0;
+  }
+  get wasCleared() {
+    return !this.createdAfter;
```
patricklx, line 27: Should be 2 `!!`, or Boolean()

RobbieTheWagner, line 27:
```suggestion
    return Boolean(this.createdAfter);
```

RobbieTheWagner: Oops, good catch!

RobbieTheWagner:
@patricklx do you think this is good to merge now or did you see anything else we needed to tweak?

patricklx:
> @patricklx do you think this is good to merge now or did you see anything else we needed to tweak?

there are still some errors like `assembler.find(...).get is not a function`, not sure if its releated to prototype extension

RobbieTheWagner:
@patricklx yeah, this PR is not meant to fix all errors, just a stepping stone.

PR #2602, opened 2024-10-25
cleanup filterBy & mapBy prototype extensions

patricklx:

### DIFF
--- a/app/libs/resizable-columns.js
+++ b/app/libs/resizable-columns.js
@@ -221,7 +221,7 @@ export default class {
       let diff = this.tableWidth - totalWidth;
       while (diff > 0) {
         columns
-          .filter((x) => !!x.visible)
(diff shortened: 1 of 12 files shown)


app/libs/resizable-columns.js
```diff
@@ -221,8 +221,8 @@ export default class {
       let diff = this.tableWidth - totalWidth;
       while (diff > 0) {
         columns
-          .filter((col) => col.visible)
-          .sortBy('width')
+          .filter((x) => !!x.visible)
```
RobbieTheWagner, line 224: Was there a reason this was changed from `col` to `x`? `x` does not indicate what the item is at all.

patricklx: Probably bad rebase

RobbieTheWagner (approved):

PR #2624, opened 2025-01-23
ember-cli-update to latest blueprints

RobbieTheWagner:


tests/helpers/index.ts
Windvis, line 1: The [index.js](https://github.com/emberjs/ember-inspector/blob/4ba73a059dec03cdbba287cbf5a5d57cd6487129/tests/helpers/index.js) file can be removed.

tests/test-helper.ts
Windvis, line 1: There still is a [test-helper.js](https://github.com/emberjs/ember-inspector/blob/4ba73a059dec03cdbba287cbf5a5d57cd6487129/tests/test-helper.js) file which seems to contain test setup logic and I'm guessing that is being overwritten by this one now. I think that's the cause for the test timeouts.

RobbieTheWagner: Ah yes, a couple extra files here. Thanks for calling them out!

PR #2625, opened 2025-04-10
Add support for Vite

mansona:
This PR finalises a lot of work done to get the Ember Inspector to support Vite 🎉  This will require a global API that Embroider can polyfill after this PR is merged: https://github.com/embroider-build/embroider/pull/2522

There will be a follow up RFC that will move this polyfill implementation into ember-source 👍

PR #2628, opened 2025-04-25
require all ember modules in one place

mansona:
## Description

This PR is starting to split out the work from #2625 

This just tries to make sure that there is only one module that is calling requireModule and we can provide alternatives for the different environments isolated to that module 👍

### DIFF
--- a/ember_debug/utils/ember/runloop.js
+++ b/ember_debug/utils/ember/runloop.js
@@ -1,18 +1,21 @@
 import Ember, { Runloop as EmberRunloop } from '../ember';
 import * as runloop from './own-runloop';
 
+// it could happen that runloop is available but _backburner is not exported (dead code)
+// then we need to use our own.
 let module = runloop;
 let _backburner = runloop._backburner;
 
 const keys = ['cancel', 'debounce', 'join', 'later', 'scheduleOnce'];
 
 if (EmberRunloop) {
+  module = EmberRunloop;
   _backburner = EmberRunloop._backburner;
 } else {
   // eslint-disable-next-line ember/new-module-imports
-  _backburner = Ember?.run?.backburner || module._backburner;
-  // eslint-disable-next-line ember/new-module-imports
   module = Ember?.run || module;
+  // eslint-disable-next-line ember/new-module-imports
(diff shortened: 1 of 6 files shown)


BlueCutOfficial:
## Description

This PR is part of the basic implementation for Vite support. It's a refactoring step that centralizes the call to `requireModule` in one module when it's used to access the modules coming from ember-source. In other words, `ember_debug/utils/ember.js` should be responsible for how ember-source things are imported, and any other module that needs these things should import them from `ember_debug/utils/ember.js`.

----

@mansona can you explain with more details what you have in mind:

> we can provide alternatives for the different environments isolated to that module.

ember_debug/utils/rsvp.js
```diff
@@ -1,21 +1,14 @@
...
 
-try {
-  module = requireModule('rsvp');
-  RSVP = module.default;
-
-  // The RSVP module should have named exports for `Promise`, etc,
-  // but some old versions do not and provide `RSVP.Promise`, etc.
-  if (!('Promise' in module)) {
-    module = RSVP;
-  }
```
BlueCutOfficial, line 13: I don't know what version this code was supporting, nor how the assignment in the if block fixes what the comment says, but with this PR: the RSVP module we import is `emberSafeRequire('rsvp')?.default;` so the content of this block should probably be cleaned up.

ember_debug/utils/ember/runloop.js
```diff
@@ -1,21 +1,21 @@
...
-  // eslint-disable-next-line ember/new-module-imports
-  _backburner = Ember?.run?.backburner || module._backburner;
+if (EmberRunloop) {
+  module = EmberRunloop;
+  _backburner = EmberRunloop._backburner;
+} else {
   // eslint-disable-next-line ember/new-module-imports
   module = Ember?.run || module;
+  // eslint-disable-next-line ember/new-module-imports
+  _backburner = Ember?.run?.backburner || _backburner;
```
BlueCutOfficial, line 18: Just reverted those blocks so `module` is assigned before `_backburner`, because it's the order chosen at the top of the file so it eases the reading.

BlueCutOfficial (approved):

PR #2629, opened 2025-04-25
define packageManager and use-node-version

mansona:
## Description

This makes it easier for people not using volta to contribute. Personally since pnpm can manage node version and pnpm version itself now I think we should remove the volta configuration, if you agree then let me know and I'll update this PR 👍


RobbieTheWagner:
@mansona I use asdf and the versions are powered by https://github.com/emberjs/ember-inspector/blob/main/.tool-versions

I fully support removing volta

PR #2642, opened 2025-05-21
move ember debug tests into own app & fix tests

patricklx:
ember debug is the thing that should/can be tested with older ember versions.
for inspector ui its not required. we can extend it for it as well though.

test fixed by mostly using setComponentTemplate as dynamic template resolving has been deprecated 

This also 
* adds back support for ember alpha+,  ember/index has been removed and we have ember/barrel now
* fixes some InElement support issues with ember 6.4

PR #2643, opened 2025-05-29
test ember 5.8 and 5.12

mansona:
## Description

This PR adds explicit tests for Ember 5.8 and 5.12 and also temporarily turns off tests for canary, beta, and release as they are currently broken. I intend to open a PR to track support for 6.4 in a bit 👍


patricklx:
We could unblock them in #2642

BlueCutOfficial (approved):
I think this PR is compatible with #2642 in terms of "unit of work". It makes sense to me to merge this first, then fix ember-try for older versions.

PR #2644, opened 2025-05-29
add packageManager for people not using volta

mansona:
## Description

This is a tiny little DX improvement for people not using volta. Let me know if you would prefer that I remove the volta config all together 👍


RobbieTheWagner:
I swear I commented on this, but not sure where my comment went.  I use asdf to set versions and we should absolutely remove all volta stuff, IMO.

PR #2654, opened 2025-06-12
Remove loader

BlueCutOfficial:


ember_debug/utils/ember/object/internals.js
```diff
@@ -1,14 +1,9 @@
...
 
-let module;
-
-if (ObjectInternals) {
-  module = ObjectInternals;
-} else {
-  module = Ember;
-}
-
-let { cacheFor, guidFor: emberGuidFor } = module;
```
BlueCutOfficial, line 11: I encountered a situation with 3.24 where `ObjectInternals` was defined but didn't contain a definition for `guidFor` and `cacheFor`, which makes this logic break. That's why I refactored.

ember_debug/utils/ember/runloop.js
```diff
@@ -10,7 +10,7 @@ const keys = ['cancel', 'debounce', 'join', 'later', 'scheduleOnce'];
 
 if (EmberRunloop) {
   module = EmberRunloop;
-  _backburner = EmberRunloop._backburner;
+  _backburner = EmberRunloop._backburner || EmberRunloop.backburner;
```
BlueCutOfficial, line 13: Until at least 3.24, we're working with `backburner`, but in more recent versions it's `_backburner`.

PR #2661, opened 2025-06-20
Remove old code that appears to be unused

BlueCutOfficial:
## Description

// TODO

### DIFF
--- a/ember_debug/object-inspector.js
+++ b/ember_debug/object-inspector.js
@@ -899,18 +899,7 @@ function addProperties(properties, hash) {
     let options = { isMandatorySetter: isMandatorySetter(desc) };
 
     if (typeof hash[prop] === 'object' && hash[prop] !== null) {
-      options.isService =
-        !('type' in hash[prop]) && hash[prop].type === 'service';
-
-      if (!options.isService) {
-        if (hash[prop].constructor) {
-          options.isService = hash[prop].constructor.isServiceFactory;
-        }
-      }
-
-      if (!options.isService) {
-        options.isService = desc.value instanceof Service;
-      }
+      options.isService = desc.value instanceof Service;
     }
     if (options.isService) {
       replaceProperty(properties, prop, inspectValue(hash, prop), options);


ember_debug/object-inspector.js
```diff
@@ -899,18 +899,7 @@ function addProperties(properties, hash) {
...
-      if (!options.isService) {
-        if (hash[prop].constructor) {
-          options.isService = hash[prop].constructor.isServiceFactory;
-        }
-      }
-
-      if (!options.isService) {
-        options.isService = desc.value instanceof Service;
-      }
+      options.isService = desc.value instanceof Service;
```
BlueCutOfficial, line 902: Hey @patricklx, do you have an opinion on this one? I can't figure out if there's an actual use case for what I removed here - starting 3.16, or if that's just artefacts from older support. I couldn't find any situation where `isService` becomes true except for the `instanceof Service`. And if that situation exists, tests don't cover it 🤷‍♀️

patricklx: mmm, maybe we should increase test coverage before then.
it looks like there is only a test in object inspector tests.
`Service should be successfully tagged as service on serialization`
I think we should improve it to test with `service` decorator:
```js
this.owner.register('service:test-inspect-service', inspectedService);

    let inspected = EmberObject.extend({
      service: inspectedService,
      service2: service('test-inspect-service'),
    }).create();
 ```

and do the same with a GlimmerComponent

BlueCutOfficial: I was also looking at that test, and I am not sure if we can trust it. I mean, I get the intent: you check you have `isService` on the object that is communicated to the UI so the UI can adapt the display to the service. But in practice, I struggle to trigger that execution path when I run my Inspector locally, with an actual app.  Many things are going on during the execution, and when I just run the published Inspector with a test app, I get pretty weird results 🤔 

<details open>
<summary>Screenshots</summary>
<br>

Counter service injected on a controller:

counter-service

Store injected on a route:

store-service

That one looks better:

router-service

</details>

patricklx: And it exists since < 3.16

PR #2664, opened 2025-06-24
fix service detection

patricklx:
## Description

The current code in object-inspector.js doesn't fit well how Services can be detected in the versions we support. Part of the code is now unused, and services appear in the inspector as Computed Properties.

this does the service detection by checking the value after obtaining the value

based on #2661 

## Screenshots
before:
image

after
image

### DIFF
--- a/tests/ember_debug/object-inspector-test.js
+++ b/tests/ember_debug/object-inspector-test.js
@@ -11,7 +11,7 @@ import EmberObject, { computed } from '@ember/object';
 import MutableArray from '@ember/array/mutable';
 import ArrayProxy from '@ember/array/proxy';
 import ObjectProxy from '@ember/object/proxy';
-import Service, { service } from '@ember/service';
+import Service, { inject as service } from '@ember/service';
 import { VERSION } from '@ember/version';
 import { tracked } from '@glimmer/tracking';
 import { module, skip, test } from 'qunit';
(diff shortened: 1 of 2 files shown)


BlueCutOfficial:
Hey, I would have a request about the PR description 😇 

- Before explaining the change you've done, could you add a sentence to describe the problem and why that change was needed, I suggest something like: 

"The current code in `object-inspector.js` doesn't fit well how Services can be detected in the versions we support. Part of the code is now unused, and services appear in the inspector as Computed Properties."

- I think this PR would also deserve 2 tiny screenshots before / after, just the part where we can see clearly the C that turned into an S in the fixed version.

tests/ember_debug/object-inspector-test.js
```diff
@@ -902,17 +902,23 @@ module('Ember Debug - Object Inspector', function (hooks) {
...
 
     let inspected = EmberObject.extend({
-      service: inspectedService,
-    }).create();
+      service: inspectedService.create(),
+      service2: service('test-inspect-service'),
+    }).create(this);
 
     let message = await inspectObject(inspected);
+    console.log('mesg', message);
```
RobbieTheWagner, line 915:
```suggestion
```

RobbieTheWagner (approved):

Issue #2612, opened 2025-01-10
Inspector doesn't work with `ember-source` v6.1


mkszepp:
**Describe the bug**
After updating my app to v6.1 (and also with new ember app), the inspector isn't working anymore.
The inspector doesn't detect the app.

**To Reproduce**
```ember new new-app --embroider --pnpm --typescript```

**Expected behavior**
Inspector should find the app

**Screenshots**
![grafik](https://github.com/user-attachments/assets/b4f802fa-8cdf-4b6c-b3dd-8067afc71883)

**Environment**
Ember cli: 6.1
Ember version: 6.1
Embroider: latest
Browser: Chrome / Firefox

**Comment by patricklx** [comment]
since ember-source is now a v2 addon, no AMD modules are beeing created. So inspector cannot  detect them. There is only a AMD compat mode for classic builds without embroider.
you have to include this code to make it work:
```js
import Ember from 'ember';
import * as runtime from '@glimmer/runtime';
import * as tracking from '@glimmer/tracking';
import * as validator from '@glimmer/validator';
import { RSVP } from '@ember/-internals/runtime';

import config from './config/environment';


window.define('@glimmer/tracking', () => tracking);
window.define('@glimmer/runtime', () => runtime);
window.define('@glimmer/validator', () => validator);
window.define('rsvp', () => RSVP);
window.define('ember', () => ({ default: Ember }));
window.define('<my-app>/config/environment', () => ({
  default: config,
}));
```

RobbieTheWagner:
@patricklx this does not seem ideal, is there any way we can fix this?

esbanarango:
Same issue here.

`"ember-source": "6.1.0",`

RobbieTheWagner:
@emberjs/framework This seems like a pretty critical issue we need a fix for.

ef4:
To be clear, this is an issue under embroider builds on ember > 6.1. We were careful not to destabilize the default build and the inspector should work fine in a completely default ember 6.1 app.

But yes, the inspector makes assumptions about AMD that we can't possibly continue to support in an ES modules world.

RobbieTheWagner:
Thanks for the context @ef4! Any thoughts on how we can support both going forward? It would be great if we could patch this and get some new versions out.

ef4:
[...]

PR #2666, opened 2025-06-26
add vite test app

patricklx:
## Description
based on #2665 

this adds a vite test app, with which we can test if ember debug can work with it once with have correct integration.
Currently this uses the workaround mentioned in https://github.com/emberjs/ember-inspector/issues/2612
see https://github.com/emberjs/ember-inspector/pull/2666/files#diff-8e8f9e09dc6186fb011afbb145595aaf7f3e5ca289f84de2b0350343f7a321a2

when we have better integration we should be able to remove it.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -631,6 +631,189 @@ importers:
         specifier: ^5.97.1
         version: 5.97.1
 
+  test-apps/vite-app:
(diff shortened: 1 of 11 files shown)


RobbieTheWagner (approved):
LGTM! @mansona mind taking a look at this one?

Issue #2632, opened 2025-05-19
Each component should have a summary of its parents -- which is easier to read than trying to view the actual tree


NullVoxPopuli:
**Is your feature request related to a problem? Please describe.**

example: 

![Image](https://github.com/user-attachments/assets/5226738d-ceca-4d6e-ae06-4ea957ab95ef)

**Describe the solution you'd like**

could be listed in component details sidepanel or something)

**Describe alternatives you've considered**

n/a

**Additional context**

n/a

PR #2667, opened 2025-06-26
show parents in sidebar

patricklx:
## Description
resolves https://github.com/emberjs/ember-inspector/issues/2632

## Screenshots
Screenshot 2025-06-26 at 17 15 25

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -1,4 +1,4 @@
-import Controller from '@ember/controller';
+import Controller, { inject as controller } from '@ember/controller';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { debounce, schedule } from '@ember/runloop';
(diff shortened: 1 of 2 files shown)


patricklx:
This is an alternative to the focus mode. or extra way to quickly see the parents of current component and select one of them for inspection
 Might be easier/faster than through the component tree

RobbieTheWagner:
> > visually tree like
> 
> in big apps it would indent off the screen tho.
> 
> I think it _should_ be upside down like a stack trace

It doesn't have to indent much, it's just unclear what I am looking at if I am a casual user and I think we could find a way to represent this better.

NullVoxPopuli:
some stack traces do:

```
most recent
at: next
at: above that
at: great grand parent
```

so maybe;
```
top of stack
in: parent 
in: grand parent
in: great grand parent
in: etc
```
:shrug:

pichfl:
I don't think my suggestions will work, but the actual problem here is a bit more complex. I'll provide a separate PR for a `object-inspector` service which will hold the global state for the object inspector, making it easier and more consistent to achieve what this PR does.

mansona (approved):
I'm approving and merging this because the functionality is there 👍 If we have any questions about the right implementation and the order of rendering we can open them up in follow-up PRs 👍

PR #2668, opened 2025-07-08
Cleanup import ember

BlueCutOfficial:
## Description

This PR is part of preparing the ground for Vite support. When running the Inspector against a Vite app, the `Ember` global from `ember_debug/utils/ember` is undefined. The problem is twofold: first, the utils file itself is not prepared for this case; second, Ember is exported as default from the utils to be used in different places of the code. 

This PR:
- Complete the "centralization" work by removing all the `import Ember from 'ember_debug/utils/ember'` in favor of the named exports.
- Re-order things in the utils so we never try to access `Ember` properties if `Ember` is not defined.

### DIFF
--- a/ember_debug/adapters/web-extension.js
+++ b/ember_debug/adapters/web-extension.js
@@ -1,7 +1,7 @@
 import BasicAdapter from './basic';
 import { typeOf } from 'ember-debug/utils/type-check';
 
-import Ember from 'ember-debug/utils/ember';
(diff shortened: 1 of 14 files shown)


mansona (approved):
This is a good PR, I've added one comment/question but I don't consider it a blocker

mansona (approved):
Perfect 👨‍🍳 👌

PR #2670, opened 2025-07-18
remove overkill use of env to simplify new build process

mansona:
## Description

This is to simplify the removal of the `wrapWithLoader` function in ember-cli-build since the build env is only being used to determine a different structure of error that is being passed to the console.

This is totally overkill for the cost of the complexity it adds to the build 🙈


BlueCutOfficial (approved):
One tiny comment, else looks good to me 👍 Rebasing the POC Vite support branch on this will make it cleaner already.

mansona:
This has been marked as internal because it only changes the behaviour in development and will not change any behaviour on production 👍

PR #2671, opened 2025-07-21
use relative imports inside ember-debug

mansona:
## Description

This converts local imports **inside** ember-debug to use relative. I'm not exactly sure why it doesn't work very well but when we move to ESM (for top level await support) this breaks without relative imports 🤷 

This PR essentially splits out the work to make the following change simpler 👍


ember_debug/route-debug.js
```diff
@@ -1,11 +1,11 @@
 /* eslint-disable ember/no-private-routing-service */
-import DebugPort from './debug-port';
-import { compareVersion } from 'ember-debug/utils/version';
-import { VERSION } from 'ember-debug/utils/ember';
-import classify from 'ember-debug/utils/classify';
-import dasherize from 'ember-debug/utils/dasherize';
-import { _backburner, later } from 'ember-debug/utils/ember/runloop';
-import bound from 'ember-debug/utils/bound-method';
+import DebugPort from './debug-port.js';
+import { compareVersion } from './utils/version';
```
BlueCutOfficial, line 3: I can see that source `ember-debug` was changed to relative, and what was already relative now has the `.js` extension. But can you please remind me where the difference lies in the first place?

mansona: to be perfectly honest I don't really know why some of them are .js and some have no extension 🤔 I'm pretty sure it's doing extension searching with `@rollup/plugin-node-resolve` in a similar way that node would.

Do you think we should remove or add the extensions? personally I think it's fine to mix and match 🤷

mansona: it all gets rolled into one (or a few) file(s) anyway

BlueCutOfficial: That's fine, my comment was not request change, just an honest question.

PR #2672, opened 2025-07-24
Convert ember-debug to ESM and simplify wrapper

mansona:
## Description

This PR is the last of the refactors necessary before we can simply add Vite support. The two main changes in this PR: 

- removing responsibility of building the `ember-debug` script from ember-cli and making it 100% the job of Rollup
- injecting that `ember-debug` script using type=module now that it is ESM

I have tested this locally and it works 🎉 I have also been very carful not to refactor the `startup-wrapper` code, instead I've just split it out and moved the exact code around. This will give us the best chance that the code will work the same and any further improvements to the code can be made in separate PRs 👍

### DIFF
--- a/ember_debug/package.json
+++ b/ember_debug/package.json
@@ -1,5 +1,6 @@
 {
   "name": "ember-debug",
+  "version": "0.0.0",
   "private": true,
   "type": "module",
   "scripts": {


NullVoxPopuli (approved):

RobbieTheWagner (approved):
Seems fine to me! I assume this is still compatible with our setup to check if you are on older Ember versions and load a totally different inspector?

PR #2673, opened 2025-07-25
Fix value.hasOwnProperty is not a function error

gitKrystan:
## Description

I keep running into this error when `value` is a proxy. With this change, it works even w/ proxies.

TBH I didn't really look at what the point of this code is so IDK if this fix is correct. I just want it to stop borking my debug sessions. 😂 

## Screenshots

image

### DIFF
--- a/ember_debug/utils/type-check.js
+++ b/ember_debug/utils/type-check.js
@@ -101,7 +101,7 @@ export function inspect(value) {
     let broken = false;
 
     for (let key in value) {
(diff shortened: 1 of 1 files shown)


gitKrystan:
should maybe guard against null too, IDK. Don't think I've run into null here yet

patricklx:
This part of code is used to display object information in inspector object side panel. Basically a subset of key/value pairs gets shown for pure objects
It's shows a subset of object keys for performance.

patricklx:
Should also check when looking at component args

PR #2674, opened 2025-07-29
Remove the "ui" in-repo-addon

pichfl:
## Description

This separation made no sense and makes it more difficult to track down styles. The logical separation is still there, all components are nested into an actual `ui` folder inside the application.


mansona (approved):
thanks for this 🎉

PR #2675, opened 2025-07-29
Fix theme colors no rendering in tests and bookmarklet

pichfl:
## Description

`theme--light` and `theme--dark` do not exist and this only worked in the extension by accident because `_setThemeColors()` method added the `light-theme` class to the body.

## Screenshots

image

Dark mode

image

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -31,7 +31,7 @@
 
     {{content-for "head-footer"}}
   </head>
-  <body class="theme--light">
+  <body class="theme-light">
     {{content-for "body"}}
 
     <script src="{{rootURL}}assets/vendor.js"></script>
--- a/tests/index.html
+++ b/tests/index.html
@@ -31,7 +31,7 @@
     {{content-for "head-footer"}}
     {{content-for "test-head-footer"}}
   </head>
-  <body class="theme--light">
+  <body class="theme-light">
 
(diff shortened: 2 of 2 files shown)


pichfl:
They are, but that code is not executed in the bookmarklet nor in the tests. The class names were just off by one dash.

RobbieTheWagner:
> They are, but that code is not executed in the bookmarklet nor in the tests. The class names were just off by one dash.

I'm just concerned that adding them explicitly would maybe break dark mode? Will it remove the class and add the dark one still?

pichfl:
https://github.com/emberjs/ember-inspector/blame/5bc63e236f3d227a9d2b49e40a54e0139ac1bf6a/app/services/adapters/web-extension.ts#L98-L107

This assumed the names that I applied, and correctly removes the existing classes. 

Some time between 6 years ago and today, `.theme--*` was renamed to `.theme-*` (from two dashes to one), but not consistently across the codebase. The `index.html` files were never adjusted, which broke theming in the bookmarklet and tests.

mansona (commented):
I have confirmed this is still working in the extension in the artefact to this PR 🎉 
Screenshot 2025-07-29 at 18 13 41

mansona (approved):

PR #2677, opened 2025-07-30
Enable CORS for locally served bookmarklet

pichfl:
## Description

Trying to run the bookmarklet without CORS creates errors in the boot strip of the bookmarklet, keeping it from being able to connect to the Ember application that you want to inspect.

Adding `ember-cli-cors` fixes this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "test": "tests"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


mansona (approved):
Looks good 👍

PR #2678, opened 2025-07-31
Move Object Inspector toggle so it remains visible in edge cases

### LINKED ISSUE #2615: Improve placement of "Toggle Details" button in narrow contexts
This fella:
CleanShot 2025-01-10 at 11 58 53@2x

The button floats into the gutter of the main panel which causes issues in narrow window contexts, in some case putting it out of view:

collapse-button-hidden

It also causes some unwanted UI overlap :

https://github.com/user-attachments/assets/b6a142bc-36ea-4a51-8096-32b9685d456b

I would like a persistent button for this control, or the ability to toggle the pane by some other persistent control.

Here's a quick idea:
CleanShot 2025-01-10 at 12 13 12@2x

pichfl:
## Description

I've added a new header bar for this, so the icon doesn't use up more of the already very limited space. 

- Moved the state into the `layout` service
- Added a new icon to get rid of the CSS transforms
- Move the button into it's own component

Sidenote: We should refactor the previous header so the breadcrumbs can wrap and the two buttons can be sent to the new header bar to make some room

## Screenshots

image

Closes #2615

### DIFF
--- a/app/styles/ui/_split.scss
+++ b/app/styles/ui/_split.scss
@@ -90,10 +90,7 @@
   overflow-y: auto;
 }
 
-.split-panel-hd-row .end {
-  margin-inline-start: auto;
-}
-
-.split-panel-hd-row:has(:only-child.end) {
-  justify-content: end;
+/* There is better ways to do this, but they create more complicated CSS */
+.split-panel-hd-row .spacer {
+  flex: 1 1 100%;
 }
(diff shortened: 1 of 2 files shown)


patricklx:
I did look at this some time ago.
I think we should keep it in the current header and when the screen is small and the object inspector is open. We should hide the left sidebar and the other toolbar buttons and only keep the toggle button visible.
The left side of the panel should always stay visible with a minimal width to be able to show the toggle button.
Would that be possible?

PR #2730, opened 2026-03-12
Enable Opening of Components in Preferred Text Editor

trek:
## Description

The inspector can already show a path to file on disk related to a component being inspected. This PR allows a user to
* set a preferred editor
* have files open in that editor when paths are clicked.

### Preference Screen

https://github.com/user-attachments/assets/917e7c59-147e-4eba-bc0e-73c9d548b6a8

### The behavior in use
https://github.com/user-attachments/assets/b7db377c-a8be-456d-880d-5353c6826077

### DIFF
--- a/ember_debug/adapters/web-extension.js
+++ b/ember_debug/adapters/web-extension.js
@@ -85,6 +85,13 @@ export default class extends BasicAdapter {
       // think we most likely can. In the limited cases (if any) where the
       // runloop is needed, the callback code should just do the wrapping
       // themselves.
+      // Handle messages from both content-script and devtools
+      if (message.from === 'content-script' || message.from === 'devtools') {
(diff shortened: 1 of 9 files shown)


trek:
@tcjr the paths are already part of the Inspector, this PR is just making them URIs to launch editors. Do you get the wrong path even with the release version of the inspector?

RobbieTheWagner (approved):
Thanks for the PR @trek! I don't see any reason we should not go ahead and ship this. We can always iterate as needed.

tcjr:
> @tcjr the paths are already part of the Inspector, this PR is just making them URIs to launch editors. Do you get the wrong path even with the release version of the inspector?

Ah, I see now.  Yes, the paths are wrong with the release version of the inspector as well.

PR #2743, opened 2026-05-14
Refactor: Prefer top-level await over callbacks in entries

pichfl:
This focusses the behavior of the entry points

1. check for the presence of Ember in the page
2. asynchronously load adapter and startInspector which in turn trigger the asynchronous loading of the bridging modules
3. start the inspector boot

### DIFF
--- a/ember_debug/entrypoints/basic-debug.js
+++ b/ember_debug/entrypoints/basic-debug.js
@@ -1,10 +1,11 @@
-import loadEmberDebugInWebpage from '../lib/load-ember-debug-in-webpage.js';
+import hasEmber from './utils/has-ember.js';
 
-loadEmberDebugInWebpage(async () => {
-  const { onEmberReady, startInspector } = await import(
-    '../lib/start-inspector.js'
-  );
+await hasEmber();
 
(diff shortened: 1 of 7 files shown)


ember_debug/entrypoints/basic-debug.js
```diff
@@ -1,10 +1,8 @@
...
-loadEmberDebugInWebpage(async () => {
-  const { onEmberReady, startInspector } = await import(
-    '../lib/start-inspector.js'
-  );
+await waitForEmber();
 
-  const adapter = (await import('../adapters/basic.js')).default;
-  onEmberReady(startInspector(adapter));
-});
+const startInspector = await import('../lib/start-inspector.js');
```
mansona, line 8: I feel like we could do with a comment above this line (in each of the entrypoints) so that we prevent future devs from thinking they can move these to the top of the file 🤔

pichfl: Added a note on each entry point and renamed the util to make it clearer

mansona (approved):
nice work! this really cleans up some of the cruft I left in on my last refactor 🎉

PR #2744, opened 2026-05-14
Reduce complexity in startInspector

pichfl:
- flatter overall function structure
- replace dynamic imports with static import. The deferred dynamic imports are handled by the entry points already
- extracted helpers for `getApplicationInstance`, `getApplications`, etc. also reusing already existing helpers like `onReady`
- Don't eagerly instantiate EmberDebug in `main.js`

This should be functionally equal to before. The main goal was to get rid of the chaos and prepare the code for future changes. This should make it much easier to separate the functional patches to Ember classes like Application and ApplicationInstance from the actual boot process in a next step.

### DIFF
--- a/ember_debug/lib/applications.js
+++ b/ember_debug/lib/applications.js
@@ -1,26 +1,26 @@
 import { Application, Namespace, guidFor } from './ember.js';
 
+export function getApplications() {
+  return Namespace.NAMESPACES.filter(function (namespace) {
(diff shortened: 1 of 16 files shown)


mansona (approved):
I looked through this quite thoroughly, and I downloaded the built extension from the artefact list and double checked that it's still working 👍 

One more domino towards this repo finally being easy to maintain 😂

PR #2745, opened 2026-05-18
Remove TypeScript from the ember-inspector UI application

pichfl:
There is currently no benefit in the app being TypeScript based and the quirks within the setup make moving things around annoying.

We can bring it back once the Inspector is a proper Vite-powered and modern Ember app if we want.


app/models/promise.js
```diff
@@ -1,5 +1,4 @@
-/* eslint-disable @typescript-eslint/no-unsafe-call */
-// @ts-expect-error This does not seem to be typed
+// @ts-nocheck
```
mansona: do we need `@ts-nocheck` for a JS file? 🤔

jsconfig.json
mansona, line 1: is there any value in having this jsconfig? 🤔

mansona (commented):
two tiny comments/questions

mansona (approved):
Nice work 🎉

PR #2751, opened 2026-05-19
Move ember-debug into packages folder

pichfl:
ember-cli projects can't be nested, so to propertly integrate a new test harness for the Inspector that can use a separate app, we need to sort out the structure of this monorepo. 

This is the first step, establishing a separate `packages` folder.


ember-cli-build.js
```diff
@@ -98,7 +98,7 @@ module.exports = function (defaults) {
 
   ['basic', 'chrome', 'firefox', 'bookmarklet', 'websocket'].forEach(
     function (dist) {
-      let emberDebug = 'ember_debug/dist';
+      let emberDebug = 'packages/ember-debug/dist';
```
mansona: if we convert this to resolve the location of ember-debug then we won't need to change it once we move the ember-cli-build folder in following PRs 👍

mansona (commented):
looks mostly as I expected 🎉 one minor suggestion plus it looks like you need to reroll the package-lock

mansona (approved):
nice work 🎉

mansona (approved):
even better 😉

PR #2752, opened 2026-05-19
Move ember-inspector from root into packages/ember-inspector

pichfl:
This follows up on #2751 and removes the nested package configuration. 

I did update some of the script names inside the package and removed the `concurrently` from the workspace root in favor of using pnpm directly. The scripts for interacting with the codebase (like lint, lint:fix, build, start, watch) remain as before.

I did extend the CI configuration slightly to include separate linter steps for the packages (and added the missing CSS lint)


.github/workflows/build.yml
```diff
@@ -143,13 +149,13 @@ jobs:
         run: |
           DATE="$(date +"%Y.%-m.%-d")"
           SHA="${GITHUB_SHA:0:7}"
-          VERSION="$(jq -r '.version' package.json)"
+          VERSION="$(jq -r '.version' packages/ember-inspector/package.json)"
```
mansona: why not just add a working directory to this run? 🤔

mansona (commented):
one minor comment/question

.github/workflows/build.yml
pichfl: Good point, updated for consistency.

mansona (approved):
Nice work 🎉

mansona (approved):
I just downloaded the built artefact and confirmed it's working 👍

PR #2753, opened 2026-05-19
Separate prettier from eslint

pichfl:
Prettier no longer recommends using it through ESLint so I separated them. 

Prettier remains limited to JS files until we decide otherwise: There are quite a few files in the project that do not match prettiers formatting and fixing them would create a bunch of noise for this PR.

This also removes the `concurrently` package from the codebase completely, pnpm works just fine now.

### DIFF
--- a/packages/ember-debug/package.json
+++ b/packages/ember-debug/package.json
@@ -7,8 +7,12 @@
     "build": "rollup --config",
     "watch": "pnpm build --watch",
     "prepare": "pnpm build",
-    "lint": "eslint . && prettier --check .",
-    "lint:fix": "eslint . --fix && prettier --write ."
+    "lint": "pnpm \"/^lint:(?!fix$)[^:]+$/\"",
+    "lint:fix": "pnpm \"/^lint:[^:]+:fix$/\" && pnpm format",
+    "lint:js": "eslint . --cache",
+    "lint:js:fix": "eslint . --fix",
+    "lint:format": "prettier --cache --check \"**/*.{js,cjs,mjs}\"",
+    "format": "prettier --cache --write \"**/*.{js,cjs,mjs}\""
   },
   "devDependencies": {
     "@babel/core": "^7.26.0",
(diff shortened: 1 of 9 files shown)


packages/ember-debug/package.json
```diff
@@ -7,8 +7,12 @@
...
     "watch": "pnpm build --watch",
     "prepare": "pnpm build",
-    "lint": "eslint . && prettier --check .",
-    "lint:fix": "eslint . --fix && prettier --write ."
+    "lint": "pnpm \"/^lint:(?!fix$)[^:]+$/\"",
+    "lint:fix": "pnpm \"/^lint:[^:]+:fix$/\"",
+    "lint:js": "eslint . --cache",
+    "lint:js:fix": "eslint . --fix",
+    "lint:format": "prettier --check \"**/*.{js,cjs,mjs}\"",
+    "lint:format:fix": "prettier --write \"**/*.{js,cjs,mjs}\""
```
mansona: so the RFC that introduced this in ember-cli actually didn't put the format:fix in this way on purpose, it has a separate `format` script that ran **after the `lint:fix` ones. This is because any fixes might need to be formatted after the change and we need to make sure prettier runs after 👍

mansona: you can see a note about this here: https://rfcs.emberjs.com/id/1055-vanilla-prettier-setup-in-blueprints

packages/ember-inspector/package.json
```diff
@@ -18,12 +18,14 @@
...
-    "lint": "concurrently \"pnpm:lint:*(!fix)\" --names \"lint:\" --prefixColors auto",
+    "lint": "pnpm \"/^lint:(?!fix$)[^:]+$/\"",
     "lint:css": "stylelint \"**/*.scss\"",
     "lint:css:fix": "stylelint \"**/*.scss\" --fix",
-    "lint:fix": "concurrently \"pnpm:lint:*:fix\" --names \"fix:\" --prefixColors auto",
+    "lint:fix": "pnpm \"/^lint:[^:]+:fix$/\"",
     "lint:hbs": "ember-template-lint .",
     "lint:hbs:fix": "ember-template-lint . --fix",
+    "lint:format": "prettier --check \"**/*.{js,cjs,mjs}\"",
+    "lint:format:fix": "prettier --write \"**/*.{js,cjs,mjs}\"",
```
mansona: same comment as above 👍

mansona (changes requested):
2 minor things and then I think this is good 👍

mansona (approved):

PR #2756, opened 2026-05-26
Add vite-test-app stub

pichfl:
This will be used to provide a proper and consistent environment on which we test the Inspector. 

This is intentionally not wired up to anything yet beyond being compiled as part of the `prepare` step. This should ensure that the PRs for the next steps have less noise.


mansona (approved):
I decided to run this locally and noticed a NaN problem 😂 I pushed the fix and we're good now 🎉

PR #2760, opened 2026-05-27
Drop support for websocket adapter

pichfl:
This transport is not covered by tests. 11 years ago there was a separate addon, which relied on this https://github.com/joostdevries/ember-cli-remote-inspector/blob/b1d4d99105e1e052ebdf472c951a5f11b83ed55f/addon/debug-server.js#L29

Today, most browsers support remote debugging using built-in DevTools and we want to consolidate the communication works across Ember Inspector and Ember Debug.


mansona (approved):
I'm pretty sure this can't be used in the current state, and it's likely broken because it is not tested. I'm going to mark this as breaking because technically it is breaking but I suspect we're going to do a major release once we move things to vite 👍

patricklx:
ember-tui and ember-native both depend on web socket adapter

patricklx:
I wonder if ember electron also used it?

PR #2762, opened 2026-05-29
Revert "Drop support for websocket adapter"

pichfl:
Reverts emberjs/ember-inspector#2760 after some actual use cases for that adapter surfaced, see comments in #2760


mansona (approved):
I've marked the PR that this is reverting as `ignore` too so it will be as if this never happened 😂

PR #2764, opened 2026-06-01
Modernize SCSS setup

pichfl:
Remove reliance on SCSS/SASS, replacing SCSS variables with native custom properties.

This should make it easier to drop SCSS and move to plain modern CSS and simpler bundling once we move the build chain to Vite.

This also moves importing the `normalize.css` package from ember-cli-build.js straight into SCSS, slowly chipping away at the vendor imports.


packages/ember-inspector/app/styles/ui/_nav.scss
```diff
@@ -127,16 +125,16 @@ $nav-main-item-height: 38px;
...
 }
 
 // when nav is collapsed, layout vertically
 .nav-footer.nav-collapsed {
   display: grid;
   grid-template:
-    "left" $nav-main-item-height - 6px
-    "right" $nav-main-item-height - 6px;
+    "left" calc(var(--nav-main-item-height) - 6px)
+    "right" var(--nav-main-item-height) - 6px;
```
evoactivity, line 137: Missing a calc() here

evoactivity (changes requested):

PR #2768, opened 2026-06-22
Remove coverage reporting to Code Climate

pichfl:
Code Climate is an AI company now and permanently disabled their API in July last year when [they spun off their code quality tracking](https://codepulsehq.com/guides/code-climate-velocity-alternative) into [Qlty](https://qlty.sh).


mansona (approved):
Looks good to me 👍

